PR #11, opened 2012-03-18
implement a "Reload" button for script errors

maul-esel:
This is an attempt to fulfill the wish fragman posted [here](http://www.autohotkey.com/forum/viewtopic.php?t=83942) which can be **_VERY**_ useful when testing scripts.

### DIFF
--- a/source/script.cpp
+++ b/source/script.cpp
@@ -15700,11 +15700,11 @@ ResultType Line::LineError(LPCTSTR aErrorText, ResultType aErrorType, LPCTSTR aE
 			g_Debugger.OutputDebug(buf);
 		else
 #endif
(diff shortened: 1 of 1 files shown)


Lexikos:
I thought I commented on this long ago... oops.

I left it open to remind me to implement something better, but it was never really interesting enough to take priority. I didn't use this implementation because I felt that misappropriating the standard button names to perform ExitApp/Reload wasn't suitable for built-in behaviour.

Once I dropped XP support in v2 I considered using a Task Dialog, which allows for defining the button names using a properly supported API, but it turns out the Task Dialog doesn't expand tabs, which the error messages rely on for formatting the "vicinity lines".

So v2.0-beta.8 implements a new custom error dialog with Reload and ExitApp buttons, among other things. I don't plan to make any change to the v1 error dialog.

PR #18, opened 2012-08-31
Added a new ImageSearch method that has new features and speed improvements.

Rseding91:
Added a new ImageSearch method that by default supports transparency in
needle images (PNG, JPG, TIF, GIF, ICO) and in some cases can be much
faster than the old ImageSearch method. The new ImageSearch also
supports other features to boost performance.

See the commit comments for a full list of changes and features.

### DIFF
--- a/source/script.h
+++ b/source/script.h
@@ -698,6 +698,11 @@ private:
 	ResultType ImageSearch(int aLeft, int aTop, int aRight, int aBottom, LPTSTR aImageFile);
 	ResultType PixelGetColor(int aX, int aY, LPTSTR aOptions);
 
+	int ImageSearch2_EraseTransColor(unsigned char * Needle, int nw, int nh, unsigned char * Trans);
+	int ImageSearch2_PixelAverageScan(unsigned char * Needle, int Stride, int w, int h, int * suX, int * suY);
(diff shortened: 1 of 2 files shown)


jhertel:
That improvement sounds very good! :-) Thanks!

I just have a comment to this:

"If GDIPlus is not available searches will be unreliable and may fail completely or match anything."

It is very good and helpful that this is stated. 

But could there, in the case of no GDI+ present on the system, be an automatic fallback instead to a more stable method, so the programmer would not have to "manually" check for the availability of GDI+ before deciding to use *M2? 

If not, could there at least be stated a simple way of checking whether GDI+ is available or not, so the programmer can use that check before using the *M2 option? As a programmer I do not like to call a function that states that in some cases the function will not work at all without myself checking for that case before calling the function. 

You do state "This method requires GDIPlus which is included in XP and later operating systems. GDIPlus can also be added to older operating systems by copying the DLL." But I cannot see a method  mentioned on how to figure out whether GDI+ is actually present or not, and thus whether the function will be stable or not with the *M2 option turned on. Testing for XP or later would not be enough, as GDI+ could be present as a DLL on earlier systems.

Could the "manual" test be something about trying to load the "gdiplus.dll" library and close it again? If so, could that perhaps be mentioned, perhaps even with the code needed for that check?

Of course this issue is becoming of less and less importance as time goes by and old systems become more rare, but still I just like to be as "safe" as possible when programming.

PR #34, opened 2015-02-18
Fix size for Gui with scroll bars

HotKeyIt:
See http://ahkscript.org/boards/viewtopic.php?f=5&t=6440


Lexikos:
SM_CXHSCROLL and SM_CYVSCROLL are also wrong. See a00339d.

PR #47, opened 2015-11-13
ControlFocus resets hotstring buffer

orbfi:
### Summary

Fixed hotstring issue when using hotkeys to set keyboard focus to another control.
### Description

Consider a hotkey that focuses a certain control, and some hostring **::aa::Foo**

If user presses "a", then uses the hotkey to move focus to an input field and types "A ", the hotstring is triggered because the first "a" was unintendedly left in the buffer.

The default case should be (IMO) that hotstring buffer is cleared on every focus change, to match the behavior when pressing tab or clicking on another control.

### DIFF
--- a/source/script2.cpp
+++ b/source/script2.cpp
@@ -2648,6 +2648,11 @@ ResultType Line::ControlFocus(LPTSTR aControl, LPTSTR aTitle, LPTSTR aText
 
 	if (SetFocus(control_window))
 	{
+		// Reset hotstring buffer, so that a hotkey bound to ControlFocus
+		// works the same as pressing TAB or Shift-TAB
+		*g_HSBuf = '\0';
+		g_HSBufLength = 0;
+		
 		g_ErrorLevel->Assign(ERRORLEVEL_NONE); // Indicate success.
 		DoControlDelay;
 	}


mmikeww:
Here is script demonstrating his problem:

```
Gui, add, edit
Gui, add, edit
Gui, show, w300, hotstringbug
Msgbox, Click OK. Then press 'a', then press 'F9', then press 'a<space>'
return

::aa::Foo

F9::
   ControlFocus, Edit2, hotstringbug ahk_class AutoHotkeyGUI
return
```

I have no comment on whether this should be intended behavior or not

orbfi:
Good catch the ``GuiControl, Focus``.

Actually more of a problem than accidental misfires is that typing a hotstring right after changing focus almost always fails because the buffer isn't empty.

Theoretically, focus could be changing due to any external reason, so the problem exists as long as hotstrings are triggered "blindly" regardless of where the text is being entered.

Intended behavior or not, real life usability matters more IMO. We've been using AHK extensively at my company. Both hotstrings and simple focus-setting hotkeys get used a lot, so this patch has been necessary.

mmikeww:
@orbfi 
Many people have had difficulty compiling this. Did you have to do anything special, or do you have any gotchas that you had to overcome?

TLMcode:
@orbfi @mmikeww I would also like to know this.

Crayder:
You just need to know the basics of compiling any major software.
(without looking at the source) I'm pretty sure this uses cmake, which does most of the work for you. You just have to open the output solution and compile.

joedf:
VisualStudio pretty much takes care of everything.

mmikeww:
maybe related to the original issue:  
https://autohotkey.com/boards/viewtopic.php?f=5&t=10911

Lexikos:
Commit 803cd16 supersedes this. It comes at no extra cost because commit a8b55c4 already retrieves the focused control (which turns out to be cheap anyway) as required to fix UWP apps, and there was already a check to detect when the foreground window changes.

However, it hasn't been merged into master yet.

mmikeww:
@Lexikos 

These Excel bugs might be related:
https://autohotkey.com/boards/viewtopic.php?f=14&t=44527

PR #53, opened 2016-03-11
Fixing "Image Has Safe Exception Handlers: No (/SAFESEH:NO)"

AndersBillLinden:
Project -> properties -> Configuration Properties -> Linker -> Advanced "Image Has Safe Exception Handlers: No (/SAFESEH:NO) because opening the solution fresh and building it with vs 2013 update 5 gave me the output "Unable to generate SAFESEH image".

### DIFF
--- a/AutoHotkeyx.vcxproj
+++ b/AutoHotkeyx.vcxproj
@@ -282,6 +282,9 @@
     <ClCompile>
       <TreatWarningAsError Condition="!$(ConfigDebug)">true</TreatWarningAsError>
     </ClCompile>
+    <Link>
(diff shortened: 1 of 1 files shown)


Lexikos:
Firstly, your patch only applies to the Debug build configuration.

Secondly, as the error message tells you, the problem is with win2kcompat.obj.  The linker will also warn you "invalid subsystem version number 5; default subsystem version assumed".  In other words, it is producing an executable that will never run on Windows 2000.  Guess what win2kcompat.obj is for?

I will change it to automatically exclude win2kcompat.obj when using the VS >= 2012 toolset.

AndersBillLinden:
Ok, thanks, then I close this.

PR #55, opened 2016-05-16
add names for Japanese special keys

ironsand:
Hi, I added some key names for Japanese special keys.

### DIFF
--- a/source/globaldata.cpp
+++ b/source/globaldata.cpp
@@ -726,6 +726,13 @@ modifier is specified along with it:
 , {_T("Launch_App1"), VK_LAUNCH_APP1}
 , {_T("Launch_App2"), VK_LAUNCH_APP2}
 
+// Special keys in Japanese keyboard
+, {_T("Muhenkan"), VK_NONCONVERT}
+, {_T("Henkan"), VK_CONVERT }
+, {_T("KanaHira"), VK_OEM_COPY }
+, {_T("HanZen1"), VK_OEM_AUTO } 
+, {_T("HanZen2"), VK_OEM_ENLW }
+
 // Probably safest to terminate it this way, with a flag value.  (plus this makes it a little easier
 // to code some loops, maybe).  Can also calculate how many elements are in the array using sizeof(array)
 // divided by sizeof(element).  UPDATE: Decided not to do this in case ever decide to sort this array; don't


Lexikos:
Hi,

Are there other programs or libraries that use these names?  Would they be immediately obvious to other Japanese users?

MDN has a convenient list of key codes/names on various platforms: [KeyboardEvent.key](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key#Key_values)

I've never used a Japanese keyboard, so perhaps you can explain the naming to me, with regard to how the keys are used.

VK_NONCONVERT and VK_CONVERT: It seems Linux, Android and Firefox OS use the terms "Muhenkan" and "Henkan", so that's fine.

VK_OEM_COPY: Why "KanaHira"? Does it swap between katakana and hiragana? It seems Gecko labels this key code as just "Hiragana". Linux and Qt seem to have three distinct keys: Hiragana, Katakana and a combined Hiragana/Katakana, but Android only has the latter.

VK_OEM_AUTO and VK_OEM_ENLW: Why "HanZen" for both keys? Gecko labels them "Hankaku" and "Zenkaku" respectively. Again, there seems to be three distinct keys on some platforms.

ironsand:
Hi,

The behavior of "VK_OEM_COPY" differs a bit by IME, but it basically swaps two input method "Kana-mode" and "Romaji-mode".

Kana-mode is a input method that directly type keys that printing are Japanese characters.
Romaji-mode (Romaji means English alphabet in Japanese) is a method that is using English keys as a Japanese phonetics.

And the surface printing of "VK_OEM_COPY" is "Katakana Hiragana Romaji". I abbreviated that.

VK_OEM_AUTO and VK_OEM_ENLW are emitted by a same key interchangeably in windows. It swaps Japanese input mode and English input mode. But I may misunderstood the behavior.

Furthermore after this pull request I noticed that `vkF2sc070` is a key code for "KanaHira", but when input language is changed to Thai or Korean, it becomes `vkFFsc070`. Honestly saying I don't know how should I treat this problem.

I thought it's just for matters of alias, but it seems not.
You can reject my pull request, but I'm very grateful if you enable these special keys.
If you need a help about Japanese information I'll gladly help you.

Thanks.

Lexikos:
> VK_OEM_AUTO and VK_OEM_ENLW are emitted by a same key interchangeably in windows. It swaps Japanese input mode and English input mode.

Do you mean *alternately* (i.e. after each press the keycode swaps)?  Does one keycode switch to "Hankaku" mode and the other switch to "Zenkaku" mode?  If that's the case, wouldn't it be better to use the more meaningful names to distinguish between the two rather than "1" and "2"?  Similarly, Numpad1 and NumpadEnd are the same key but named according to the effect of the keycode.

> Furthermore after this pull request I noticed that `vkF2sc070` is a key code for "KanaHira", but when input language is changed to Thai or Korean, it becomes `vkFFsc070`.

Perhaps that's why Gecko labels it: "Hiragana" for Japanese keyboard layout, "Unidentified" for the others.

0xFF is officially "reserved" according to the Windows SDK. The name could be mapped by SC instead of VK, but that might produce unexpected results on any keyboard layout (such as a custom one) which maps the "KanaHira" function to another physical key.

Lexikos:
I'd be happy to merge this if my questions could be answered.

joeyama:
Hi, I am using Win10Pro JP version and have JP layout keyboard.
All of his proposal names make sense for JP users like "Muhenkan", "Henkan", "KanaHira" and "HanZen1(and 2)", can immediately identify which keys.
And if there are any ways to confirm VK names with patched AHK then I can and I will check and confirm(like 'VK_NONCONVERT' is correct for "Mhenkan" key.)
This could be  JP users only but really, really helpful instead of search and put scan codes.
(I am not so sure about any other localized keyboards and there could be demands for other requests/patches. But looks like not huge number of VK exists so far, if this list is correct.)

http://kbdedit.com/manual/low_level_vk_list.html

ironsand:
I'm sorry, I couldn't take time for checking the problem.

The behavior of the key depends on status of the IME.
When input mode is Hiragana the key emits `VK_OEM_ENLW` for keydown and `VK_OEM_AUTO` for keyup.
When input mode is alphabet it emits `VK_OEM_AUTO` for keydown and `VK_OEM_ENLW` for keyup.

There are Google IME and Microsoft IME for Japanese, and with both IME the behavior are same.

And when other IME is active, the key emits `VK_OEM_3` for keydown/keyup.
I confirmed it with Chinese, Korean, Thai and English IME.

Is it enough info for the merge?

PR #78, opened 2017-06-24
Implement dynamic function

HotKeyIt:
Add inline keyword for function definition to allow inline function.

### DIFF
--- a/source/script.cpp
+++ b/source/script.cpp
@@ -1546,6 +1546,11 @@ bool IsFunction(LPTSTR aBuf, bool *aPendingFunctionHasBrace = NULL)
 // *aPendingFunctionHasBrace is set to true if a brace is present at the end, or false otherwise.
 // In addition, any open-brace is removed from aBuf in this mode.
 {
+	if (!_tcsnicmp(aBuf, _T("inline "), 7))
+	{
+		aBuf += 7;
+	}
+
 	LPTSTR action_end = StrChrAny(aBuf, EXPR_ALL_SYMBOLS EXPR_ILLEGAL_CHARS);
 	// Can't be a function definition or call without an open-parenthesis as first char found by the above.
 	// In addition, if action_end isn't NULL, that confirms that the string in aBuf prior to action_end contains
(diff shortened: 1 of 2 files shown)


Lexikos:
Why?

HotKeyIt:
Mainly to shorten repeating code.
Since this allows to use the variables in current scope, we can kind of "return"/set multiple variables and don't need a lot of parameters.

Lexikos:
Using dynamic variables? Ugh.

If this is something I wanted in the language (it isn't), I would think it should be more explicit (at the point the variable is assigned/exported).

"Inline function" already has a meaning, and it's not this.

HotKeyIt:
I see, yes the meaning is not the same but the effect is.
Unfortunately only when using dynamic variables, but I think it should be possible to allow to use the variables in current scope, right? Then it would do exactly the same as inline function does.

Lexikos:
What? No, the effect is totally different. `inline` in C++ gives the compiler a suggestion to inline the function; to insert its implementation directly at the call site, as an optimisation. It generally does not affect the behaviour of the function at all, and is only a suggestion, which may have no effect.

On the other hand, this affects the behaviour of dynamic variable references (and A_ThisFunc?) inside the function, in a very non-obvious way.

I don't understand any part of your post beyond the first sentence.

HotKeyIt:
Sorry, I completely confused myself with inline function and macro :(
What I had in mind is that the code would run as if it was not a function but 'gosub' in current scope similar to macro in c++ https://github.com/Lexikos/AutoHotkey_L/blob/master/source/script.h#L90.

HotKeyIt:
How about to call it dynamic?
I think it is fine that A_ThisFunc is not reported or relates to the function that calls this function.
It would be great if all variables would be resolved at current scope but as I understand this would require to make them all dynamic and a lot of code would needs to be changed.
I think it is still great to be able to resolve at least dynamic variables to current scope.

Lexikos:
That is no better. It doesn't make the function (more) dynamic. It doesn't make the variable references dynamic. They were already dynamic.

It does not resolve "dynamic variables to current scope". Inside the setvars/deref function, *current scope* is inside the setvars/deref function, as demonstrated by the fact that non-dynamic variables are local to that function. What it does is change dynamic variable references to use *the caller's* scope.

PR #81, opened 2017-09-02
Fix SetTimer to find A_ThisLabel in function.

HotKeyIt:


mmikeww:
reference:

https://autohotkey.com/boards/viewtopic.php?f=37&t=36576

PR #82, opened 2017-09-20
Fix bug for low memory when file has odd number of bytes

HotKeyIt:


Lexikos:
What "bug for low memory"?

I see one case where the extra byte is needed, but I would not have described it that way. Merely having an odd number of bytes is not enough for the function to fail. Before I merge this, please elaborate.

HotKeyIt:
Try running this code in debug mode:
``FileDelete("C:\Temp\temptestfile.txt")
,w:=FileOpen("C:\Temp\temptestfile.txt","w")
,VarSetCapacity(var,1)
,w.RawWrite(&var,1)
,w.Close()
,FileRead("C:\Temp\temptestfile.txt","RAW")``

PR #83, opened 2017-10-02
Adding check to avoid instr returning 1 for blank or object needle.

HelgeffegleH:
Currently `instr(str,"")` and (eg) `instr(str,[])` returns `1`.

### DIFF
--- a/source/script2.cpp
+++ b/source/script2.cpp
@@ -11566,7 +11566,8 @@ BIF_DECL(BIF_InStr)
 	LPTSTR haystack = ParamIndexToString(0, _f_number_buf, (size_t *)&haystack_length);
 	size_t needle_length;
 	LPTSTR needle = ParamIndexToString(1, needle_buf, &needle_length);
-	
+	if (*needle == NULL)  // avoid instr(str, "") and (eg) instr(str, []) returning 1.
+		_f_return_i(0);
 	// v1.0.43.03: Rather than adding a third value to the CaseSensitive parameter, it seems better to
 	// obey StringCaseSense because:
 	// 1) It matches the behavior of the equal operator (=) in expressions.
--- a/source/util.cpp
+++ b/source/util.cpp
@@ -566,10 +566,12 @@ LPTSTR tcsrstr(LPTSTR aStr, size_t aStr_length, LPCTSTR aPattern, StringCaseSens
 {
 	if (aOccurrence < 1)
 		return NULL;
-	if (!*aPattern)
-		// The empty string is found in every string, and since we're searching from the right, return
-		// the position of the zero terminator to indicate the situation:
-		return aStr + aStr_length;
+	// Currently the only calling function is already doing the below check, hence it is redundant.
+	// Future callers should consider how to handle the case where (!*aPattern) == true.
+	//if (!*aPattern)
+	//	// The empty string is found in every string, and since we're searching from the right, return
+	//	// the position of the zero terminator to indicate the situation:
+	//	return aStr + aStr_length;
 
 	size_t aPattern_length = _tcslen(aPattern);
 	TCHAR aPattern_last_char = aPattern[aPattern_length - 1];


Ragnar-F:
Technically, `instr(str,"")` returning 1 is correct. Every string contains a string with a length of 0.

HelgeffegleH:
Hello **Ragnar-F** :wave: 

`instr` answers a more specific question than _in?_, it answers, _in, then where?_. If you want to argue that `""` corresponds to `'\0'` then the answer should be `strlen(str)+1`, which would indeed be `1` for `instr("","")`. 

__Stretching the limits of my imagination:__ If you are drawing parallels to set theory where each set is said to contain the empty set and you consider `""`  to correspond to the empty set, then sure, `""` is a subset (or substr) of all strings, but elements in a set doesn't really have positions. You could use the argument for `infinstring var, % ""` being `true` in `v1` though. :wink: 

> This function returns the position of an occurrence of the string Needle in the string Haystack. Position 1 is the first character; this is because 0 is synonymous with "false", making it an intuitive "not found" indicator.

**source:** [`instr`](https://lexikos.github.io/v2/docs/commands/InStr.htm "ahk docs")

cheers.

Lexikos:
I agree with Ragnar-F. A substring of zero characters can be found at any position within any string. In particular, `InStr("", "")` should return 1, as  `a == b` implies `InStr(a, b) == 1` when a and b are strings.

As for objects; the convention in v2 for **invalid parameters** is to throw an exception. The parameters are clearly documented as being strings. (On the other hand, most functions ignore the possibility of the parameter type being incorrect, for code size/simplicity. There are many string parameters across many functions, so the cost of validating each and every one seems to outweigh the benefit of detecting such unlikely errors.)

Also, consider that `RegExMatch([], "")` returns 1 and `RegExReplace([], "", "x")` returns "x" (but StrReplace does not support an empty needle).

HelgeffegleH:
Hello __Lexikos__ :wave:

I'm very happy to see you responding. Your expertise is greatly missed at the forums. I hope you have had a long and happy vaccation and will soon return. :beach_umbrella: :bee: 

---

Unfortunately, my reply turned out lengthy, if you cannot find the time or desire to read, please just observe my main point,

> The proposed change yields that, `instr(a,b) > 0` implies there was a match of actual characters. 

which I find more useful and intuitive than matching _empty strings_ which could be considered to always exist anyways.

Full reply
---


> A substring of zero characters can be found at any position within any string.

In my opinion, the concept of finding nothing anywhere isn't useful in the context of this function. I guess you didn't write the function, since neither the `occurance` parameter, nor the option to search from the right comply (well) with the concept.

> In particular, `InStr("", "")` should return `1`, as `a == b` implies `InStr(a, b) == 1`

In particular not, since the haystack `""` doesn't have any characters, hence the returned position `1`, which means the first character according to the documentation, has no sensical meaning. I didn't add a check for a blank `haystack`, since the check for the blank `needle`  cause the same outcome. In effect the change implies that it is not meaningful to search for nothing, nor is it meaningful to search in something which doesn't contain anything except possibly nothing. If you think there is an empty string everywhere, there is never a need to search for it. Hence, instead of,  

> `a == b` implies `InStr(a, b) == 1`

we get  the more intuitive and useful, 

> `instr(a,b) > 0` implies there was a match of actual characters. 

Consequently, we can treat the common case, where a blank string is (at best) uninteresting, like this

```autohotkey
if instr(hay, needle) ; && needle != "" ; uncomment for current behaviour
	action(needle)
```

and the special case where we want to handle certain needles regardless if they match or not, we treat the blank needle like any other special case,

```autohotkey
if instr(hay, needle) || needle == "" || needle == "banana" ; currently we do not need to check if needle is blank
	action(needle)
```

__In summary__, the _minimal_ change gives inutitive functionallity, convenience, performance gain for the common case and consistency with the documentation, without any need to add anything to the documentation :+1: 

---

In trying to understand the result from searching for `""` from the right, I found that I had failed to remove this check in `tcsrstr`,

```cpp
if (!*aPattern)
	// The empty string is found in every string, and since we're searching from the right, return
	// the position of the zero terminator to indicate the situation:
	return aStr + aStr_length;
```

The function seems to only be called from `BIF_InStr`, so it will never be true.

---

Regarding exceptions for invalid parameter, I considered it, but decided it was a job better suited for someone more knowledgable :fearful:. Although it would be convenient to be notified on such mistakes, I agree that it might not be worth the trouble. 

---

Regarding `regexXXX`, other than the invaild parameter, `[]`, there is no relation to this. `""`  is a valid regular expression, although, perhaps it is more often used together with other expressions, eg, `regexreplace("a37bc","\d+|"," ")`.
The documentation on `regexXXX` says that the haystack can contain binary zero, but it could be emphasised that the returned _position_ might not contain an actual character.


Cheers :coffee:

Lexikos:
I am not convinced that the behaviour with empty strings should change.

I would accept a patch which purely adds type-checks for string parameters. However, if done for InStr, it should be done for all functions which accept only strings. I am generally inclined to omit such checks due to the rarity of need vs code size. Future revision to the function ABI might allow such type checks to be performed in a central place, based on metadata, which would provide better error-checking while minimising the cost.

HelgeffegleH:
Hello. 

Thanks for your considerations, I am glad decisions are being made :+1: .

PR #85, opened 2017-10-30
fixed from PVS-Studio

ip-gpu:
I'm a member of the Pinguem.ru competition on finding errors in open source projects. A bug, found using PVS-Studio. Warnings:

[V728](https://www.viva64.com/en/w/v728/) An excessive check can be simplified. The '(A && B) || (!A && !B)' expression is equivalent to the 'bool(A) == bool(B)' expression. wingroup.cpp 114
[V773](https://www.viva64.com/en/w/v773/) The function was exited without releasing the 'new_mem' pointer. A memory leak is possible. keyboard_mouse.cpp 2712
[V773](https://www.viva64.com/en/w/v773/) The function was exited without releasing the 'token' pointer. A memory leak is possible. script.cpp 12300
[V728](https://www.viva64.com/en/w/v728/) An excessive check can be simplified. The '(A && B) || (!A && !B)' expression is equivalent to the 'bool(A) == bool(B)' expression. script.h 1987

### DIFF
--- a/source/WinGroup.cpp
+++ b/source/WinGroup.cpp
@@ -111,7 +111,7 @@ ResultType WinGroup::CloseAndGoToNext(bool aStartWithMostRecent)
 	// Even if it's NULL, don't return since the legacy behavior is to continue on to the final part below.
 
 	WindowSpec *win_spec = IsMember(fore_win, *g);
-	if (   (mIsModeActivate && win_spec) || (!mIsModeActivate && !win_spec)   )
+	if (   mIsModeActivate && win_spec!=nullptr)
 	{
 		// If the user is using a GroupActivate hotkey, we don't want to close
 		// the foreground window if it's not a member of the group.  Conversely,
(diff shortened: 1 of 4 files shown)


Lexikos:
You have introduced an error into wingroup.cpp: the code block will no longer execute when mIsModeActivate is false. `A && B != nullptr` is clearly not the same as `bool(A) == bool(B)`.

 The "excessive check" is not a bug, and is arguably more intuitive (i.e. easier to comprehend) than the proposed alternative (bool(A) == bool(B)).

sAbortArraySend would be true only if either new_mem was already nullptr, or a previous call to ExpandEventArray failed. In the former case, free(new_mem) has no effect. In the latter case (which might be impossible), it would be better to avoid attempting another malloc in the first place, since (if somehow successful) it will only be discarded (i.e. check sAbortArraySend before malloc).

There is a potential bug in that `free(sEventSI)` might be called without changing the value of `sEventSI`, in which case sEventSI will be invalid when CleanupEventArray calls `free(sEventSI)`. It would be clearer and more correct to simply return immediately after setting `sAbortArraySend = true`. Commit f410f47 does this.

You are correct about `token`. Commit f410f47 deletes it.

You have introduced an error into script.h: `*aX` and `*aY` are not boolean, therefore `(*aX == *aY)` is not equivalent to `(!*aX && !*aY) || (*aX && *aY)`.

PR #87, opened 2017-11-27
Integer key (instance variable) assignment in class body causes invalid offsets for string and object keys.

HelgeffegleH:
Decrementing `mKeyOffsetObject` and `mKeyOffsetString` when removing integer key in `Object::EndClassDefinition()`.

Avoids invalid offsets for Object and String keys in `mFields` when assigning integer keys (instance variables) in class body.

Alternatively, one could unconditionally set,

```
mKeyOffsetObject = 0;
mKeyOffsetString = 0;
```

after the for-loop. Due to rarity of such assignments and clarity, the conditional decrementation might be preferable.

Issue example,

```autohotkey
msgbox(type(test)) ; Object, due to __class key not found

class test {
	0 := 1
}
```

### DIFF
--- a/source/hook.cpp
+++ b/source/hook.cpp
@@ -167,6 +167,8 @@ LRESULT CALLBACK LowLevelKeybdProc(int aCode, WPARAM wParam, LPARAM lParam)
 	// purpose of updating modifier and key states:
 	if (event.dwExtraInfo == KEY_PHYS_IGNORE)
 		event.flags &= ~LLKHF_INJECTED;
+	else if (event.dwExtraInfo == KEY_BLOCK_THIS)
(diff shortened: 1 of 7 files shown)


HelgeffegleH:
Sorry for the __noise__, that is the old commits and their reverts, that wasn't on purpose.  :-1:

Lexikos:
Thanks.

In future, you can strip out commits that shouldn't be there by using `git rebase` of some form and (to overwrite the remote branch if you've already pushed it) the `-f` switch with `git push`.  I would never merge it as it is.

HelgeffegleH:
Thank you for the information :+1:

PR #91, opened 2018-02-15
Removing many built-in (thread settings) variables in favour of functions.

HelgeffegleH:
Hello :wave: 

Removing the following built-in variables:
```autohotkey
A_TitleMatchMode
A_TitleMatchModeSpeed
A_DetectHiddenWindows
A_DetectHiddenText
A_StringCaseSense
A_FileEncoding
A_SendMode
A_SendLevel
A_StoreCapslockMode
A_KeyDelay
A_KeyDuration
A_KeyDelayPlay
A_KeyDurationPlay
A_WinDelay
A_ControlDelay
A_MouseDelay
A_MouseDelayPlay
A_DefaultMouseSpeed
A_CoordModeToolTip
A_CoordModePixel
A_CoordModeMouse
A_CoordModeCaret
A_CoordModeMenu
A_RegView
```

Introducing the corresponding functions, for setting and getting,

```autohotkey
detecthiddentext
detecthiddenwindows
StoreCapslock
TitleMatchFast
ControlDelay
MouseDelay
MouseDelayPlay
WinDelay
MouseSpeed
TitleMatchMode
SendMode
StringCaseSense
CoordModePixel
CoordModeMouse
CoordModeToolTip
CoordModeCaret
CoordModeMenu
FileEncoding
RegView
```

The general convention is, approximately,

`A_Setting` replaced by `previousSetting := Setting([newSetting])`

If no new settings is specified, the current setting is returned.

`KeyDelay()` returns an array:  `[prevDelay, prevDuration]`.

String options removed except for `FileEncoding()`.

Example usage:

```autohotkey
pDHW := detectHiddenWindows(true)	; set
; ...
detectHiddenWindows(pDHW)	; restore
```

Notes:
`Critical()` is impelemented but commented out, since it is not clear how detection of `Critical()` as the first line of a thread's subroutine should be handled.
`ListLines` + `A_ListLines` is fine as is in __v1.1.28.00__, it doesn't fit this convension since `ListLines` on its own is practical for opening the window.
Exceptions are thrown when values are out of bounds.

## String options - assorted comments

I see no reason to have `"On"` or `"Off"` string options, `true` or `false` is clear enough and performs better.  I am neutral on adding string options for the `coordmodeXXX` functions. `CoordMode` options are 
```
1 - client 
2 - window
3 - screen 
```
For `sendMode`, there is in practice only two options to remember, 
```
1 - event
2 - sendInput
```
`3` - is `play` and `4` is `InputThenPlay`

There are no good string option candidates for `titleMatchMode 1-3`, having `4` as `regex` is fine imo.  `StringCaseSense` is `0` for not `casesense`, `1` for `casesense`, and `2` for `locale`

## About the implementation and convention

I make no claims on the quality of the implementation, but I __[wish](https://autohotkey.com/boards/viewforum.php?f=13)__ the convention will be used. These built-in variables are a relic from the time of commands where you couldn't return the previous setting without using an output variable. Having two ways to set these settings isn't beneficial.

If there is an interest in using this, I have the following todo list:

* Update the documentation.
* Search for obsolete comments refering to removed `A_XXX` variables.
* Add back string options where/if required.

Cheers.


mmikeww:
> These built-in variables are a relic from the time of commands where you couldn't return the previous setting without using an output variable. Having two ways to set these settings isn't beneficial.

This is a good point, but v2 still allows "command-like" syntax, so I guess you're suggesting that if you need the previous value, you are forcing the user to now use the function syntax in order to retrieve it. Not a huge deal imo.

HelgeffegleH:
Hello __mmikeww__ :wave: 

That is indeed a consequence, but not the purpose :wink: 

Cheers.

Lexikos:
I am surprised that you would suggest this.

These settings are, in essence, just variables. Why should we have a function for each variable, instead of a variable? There are already long-standing conventions for reading from and writing to variables, and no great need for a new convention.

"KeyDelay(x)" isn't as obvious as "SetKeyDelay x" or "A_KeyDelay := x", and unless it's clearly indicated by the variable name, the convention of returning the previous value isn't obvious either.

If the goal is to increase convenience by avoiding the need to repeat the variable name, it still falls short when the script needs to change multiple settings (and you still repeat it when restoring the setting). It is fairly trivial for a user to roll their own solution which uses whatever convention suits the user.

Note that when you coerce a parameter to int, non-numeric values end up as zero (or perhaps the value of any numeric prefix), so invalid non-numeric values (which include "On" in your branch) are treated as 0 and not detected as errors if 0 is a valid value.

> Having two ways to set these settings isn't beneficial.

Given that the ability to write to built-in variables was only added in v2-alpha, the natural conclusion would be to remove the Set functions. That is my intention, but since it doesn't seem to actually matter either way, I haven't yet. Some of the functions would stay, such as SetKeyDelay, because it sets multiple parameters.

> [...] since it is not clear how detection of Critical() as the first line of a thread's subroutine should be handled.

It's not handled now; `Critical()` and (in recent builds) `Critical` both resolve to an expression, not ACT_CRITICAL.  I have prepared some changes which supersede these checks.

> I see no reason to have "On" or "Off" string options [...]

For these variables, I agree (and I had written a paragraph about it in v2-thoughts). However, there are some cases where boolean is not implied, so true/false would be less clear. For instance, `SetTimer(, 0)` would appear to override `SetTimer(, 100)`, but `SetTimer(, "off")` can turn off the timer without changing its interval. `Hotkey(k,, false)` is somewhat less clear than `Hotkey(k,, "off")`.

>  I am neutral on adding string options for the coordmodeXXX functions.

I am surprised. I think that "magic numbers" are bad, in general, hence why MsgBox now takes a string of options.

> [...] you are forcing the user to now use the function syntax in order to retrieve it. 

It's only the difference of A_ vs (), but still, I think it's not particularly rational/intuitive.

Lexikos:
>The function can perform two useful actions

Not good enough.

> the function is obviously more convenient, and shorter.

Yes, you are stating the obvious. You didn't address my point, which is that there are solutions which are shorter and more convenient, and can be implemented by the script. For instance,

    old := Opt({TitleMatchMode: "Exact", WinDelay: 50})
    WinMove x, y, w, h, "Some Window"
    old.Restore()  ; ... or this could be implied by __delete.

There are also other possibilities which could supersede the need to change these settings in the first place, such as:

    WinMove x, y, w, h, "Some Window ahk_mode 3"

Though much less likely (and more contrived), future language features (or meta-programming) could hypothetically provide a better solution.  For instance,

    localize A_KeyDelay := 10, A_KeyDuration := 20 {
        Send "some keys"
    }

> For example, ShowMsgBox would make it more clear to the uninitiated

This brings me to another point. A function or command is the basic means by which a script *does something*, or so I've written in the documentation. When one sees a function call named as a verb, it is reasonable to assume that's what the function does. But DetectHiddenWindows() does not detect hidden windows, KeyDelay() does not (perform a) key delay, StoreCapslock() does not store Capslock, and so on. MsgBox has no misleading apparent meaning, because it is just a noun.

Variables imply a different meaning; it is obvious that `x := A_KeyDelay` is merely taking a value from somewhere and putting it somewhere else. A new user seeing `A_KeyDelay := 50, Send("some keys")` probably has some idea of what it does without consulting the manual. The "A_" convention helps here, if the user has come across any built-in variables before.

> The convention to return what is changed or removed is quite widespread

"Quite widespread" doesn't quite compare to "fundamental to imperative languages everywhere", which is how I would describe assignments and variable references. Users must understand assignments and variables before they can understand code which uses your functions. A function can do and return virtually anything, but there are only so many ways a variable can be used.

> You have hinted earlier that it might not be worth the overhead to detect such input errors

Passing `"not a number"` to a math function (for example) is somewhat different to passing `"On"` to a function intended to be a direct replacement for a well-known command which would have accepted that value.

> I am surprised, I consider it on par with all the magic letters 

Those would just be letters, not magic. What makes a magic number magic is that it is totally incomprehensible to someone who hasn't already memorized its meaning. Letters have mnemonic value, such as R (recurse), D (directory), etc. Even after the user reads the documentation, there is nothing to help a user remember that 1 means "client" and 3 means "screen".

> Most often, what the settings are, is a black box to the user, hence it has no reason to question why it must set them via functions rather than just assign variables, hence there is no inutition to consider.

I disagree.

> hence it has no reason to question why it must set them via functions rather than just assign variables

That's totally beside the point, which was (I think) about an uneducated user understanding someone else's code more readily, and about the language being more intuitive and consistent in general.

HelgeffegleH:
Hello, thank you for your answer.

> There are also other possibilities which could supersede the need to change these settings in the first place, such as `WinMove x, y, w, h, "Some Window ahk_mode 3"`

That would be an __alternative__ way to set the settings, and not a very practical or efficient way for a function/subroutine which uses multiple calls to (eg) `winMove`.  Hence, I do not consider that _superseding_ or relevant for the decision on wether to use BIVs or BIFs for these settings.

> A function or command is the basic means by which a script does something

The script would call the functions, which in turn does something. So it is ok :+1: .

>  When one sees a function call named as a verb, it is reasonable to assume that's what the function does

For example, seeing `KeyDelay(10)` for the first time, given your assumed assumptions, it is then reasonable to ask which key(s) it is delaying, and in which time unit (if _time_ is deduced from `10`), hence, you look in the __manual__. It is less reasonable to assume it delays your next `space` for `10` minutes. I didn't see many questions in the forum asking where the _hidden windows_, which supposedly should have been detected, went.

> A new user seeing `A_KeyDelay := 50, Send("some keys")` probably has some idea of what it does without consulting the manual.

Imo, you are exaggerating difference in readability/clarity between 

```autohotkey
A_KeyDelay := 50, Send("some keys")
```

and 

```autohotkey
KeyDelay(50), Send("some keys")
```

And, I am not at all concerned about new users being able to guess what the language does without consulting the manual. These settings are AHK specific, they are not something which needs live up to new user's expectations, in contrast to, for example, a `while` loop, `if else` statments or `ret := f(param, ...)`, and such.

Besides, _keyDelay_ being a verb isn't obvious, and _msgBox_ could be a verb followed by a noun.

>  The "A_" convention helps here, if the user has come across any built-in variables before.

There is not much help implied by the `A_` convention, other than the prefix heeding a warning of inconsitent behaviour compared to _regular_ variables. You need to learn how to use each one of  the `A_` keywords. Typical questions: __Writeable__? __Constant__, __context__ or __time__ dependent value? That is not how __I__ would describe the fundamentals of variable references in languages everywhere. They are more like (unconventionally called) functions, than variables.

>Users must understand assignments and variables before they can understand code which uses your functions

What ever your point is, users should understand assignments, variables and function calls before using any thread/script settings regardless of how they are set.

>Passing "not a number" to a math function (for example) is somewhat different to passing "On" to a function intended to be a direct replacement for a well-known command which would have accepted that value.

I disagree, built-in functions documented to accept only numbers (integers) should behave the same on invalid input, regardless of wether the function had a similarily named command in v1 or not. I guess I  should have used `ParamIndexToNumber` :thinking:. Also, if you are really considering ease of conversion, then I think using the functions to set the settings are quite reasonable :wink:.

> What makes a magic number magic is that it is totally incomprehensible to someone who hasn't already memorized its meaning. Letters have mnemonic value, 

Sure, they are just _mysterious_ then, compared to full words, typed either as strings or methods/functions.

> You didn't address my point

You addressed it yourself just fine when you stated the obvious. Also, it makes a good argument for numeric options, rather than strings :+1:. It seems you are disregarding that there is a whole [community](https://autohotkey.com/boards/index.php) which likes to share code, and in shared code, a recognisable and convenient, built-in way, of setting these settings is preferable, imho.

Finally, it seems most of the discussion is pointless, because it wouldn't have mattered what I had called the functions or if they had used the same string options as the BIVs, you prefer the BIVs. I respect your opinion, but I disagree with the BIFs being irrational or unintuitive or otherwise difficult to understand or to use.

To be honest, when I implemented this, I was quite neutral to the change, I thought it would add a touch of convenience. However, after this discussion, I have changed my mind, I think this is a natural first step in the phasing out of (most of, if not all of) the `A_` variables, which will make the language more __readily__ understandable, and more __intuitive__ and __consistent__ in general :stuck_out_tongue_winking_eye:  . 

Cheers :coffee: .

HelgeffegleH:
I changed the functions to use `ParamIndexToNumber` instead of `ParamIndexToInt`, for the sake of completeness if nothing else.

Regarding [this recent change](https://github.com/Lexikos/AutoHotkey_L/commit/c2fc860f8187faaf970c0f5ae9ffed0d8b79e474), should `critical true` cause `g->PeekFrequency = 17;` for similar reasons?

_I wrote_:
>  It seems you are disregarding that there is a whole community [...]

This was a clumpsy choise of words, given your longstanding commitment to the community it might be interpreted as an insult, it was not my intention, __I apoligise, sincerely__.

Cheers :wave:

HelgeffegleH:
Even though I prefer this over the BIVs, I don't think this solution sufficient, more drastic changes are needed, I close this for that reason.

PR #92, opened 2018-03-07
ComObjActive() looks in the ROT for instances identified by Monikers

Cleric-K:
Now it is possible to get objects from the Running Object Table (ROT)
by Moniker even for names which are not present in the Registry.

CLSIDFromString() fails if the human readable name is not present
in the Registry. Now if CLSIDFromString() fails, the ROT is queried
for objects registered with the requested name.

### DIFF
--- a/source/script_com.cpp
+++ b/source/script_com.cpp
@@ -172,29 +172,49 @@ BIF_DECL(BIF_ComObjActive)
 	{
 		HRESULT hr;
 		CLSID clsid;
-		IUnknown *punk;
+		IUnknown *punk = NULL;
 		hr = CLSIDFromString(CStringWCharFromTCharIfNeeded(TokenToString(*aParam[0])), &clsid);
 		if (SUCCEEDED(hr))
 		{
+			// valid clsid
 			hr = GetActiveObject(clsid, NULL, &punk);
-			if (SUCCEEDED(hr))
+		}
+		else
+		{
+			// not a valid CLSID or name not present in the registry
+			// try getting the object from the ROT by name
+			LPRUNNINGOBJECTTABLE pRot;
+			LPMONIKER pMon;
+
+			if (SUCCEEDED(GetRunningObjectTable(0, &pRot)))
 			{
-				IDispatch *pdisp;
-				hr = punk->QueryInterface(IID_IDispatch, (void **)&pdisp);
-				punk->Release();
-				if (SUCCEEDED(hr))
+				if (SUCCEEDED(CreateFileMoniker(CStringWCharFromTCharIfNeeded(TokenToString(*aParam[0])), &pMon)))
 				{
-					if (obj = new ComObject(pdisp))
-					{
-						aResultToken.symbol = SYM_OBJECT;
-						aResultToken.object = obj;
-						return;
-					}
-					hr = E_OUTOFMEMORY;
-					pdisp->Release();
+					pRot->GetObjectW(pMon, &punk);
+
+					pMon->Release();
 				}
+				pRot->Release();
 			}
 		}
+
+		if (punk != NULL) {
+			IDispatch *pdisp;
+			hr = punk->QueryInterface(IID_IDispatch, (void **)&pdisp);
+			punk->Release();
+			if (SUCCEEDED(hr))
+			{
+				if (obj = new ComObject(pdisp))
+				{
+					aResultToken.symbol = SYM_OBJECT;
+					aResultToken.object = obj;
+					return;
+				}
+				hr = E_OUTOFMEMORY;
+				pdisp->Release();
+			}
+		}
+
 		ComError(hr);
 	}
 }


Lexikos:
I intended to look into this before merging it, since I'm not convinced that it is useful or appropriate, but that being the case, I don't really have any motive to do so.  Can you give some examples? How do you know a particular name is valid?

Your description makes it sound like the moniker is equivalent to the string normally found in the registry, but looking at the code, I would assume your code will get an active object by *file path*. That might be useful, but only if the user knows they can pass a file path (not just a "name" as you've described), and only if the program actually registers the object that way.

Looking at the objects currently active on my system, there are several with different kinds of monikers that I would never be able to guess, that certainly aren't documented. In those cases it doesn't make sense to pass a name to ComObjActive without having first discovered the name using some other function or tool. [GetActiveObjects](https://autohotkey.com/boards/viewtopic.php?f=6&t=6494) can both retrieve the names and get the objects (and is not limited to file monikers).

Cleric-K:
Hi,
What you say is perfectly correct. Here was my motivation for the patch:

I made the modification for myself when I was writing this https://github.com/Cleric-K/vJoySerialFeeder

Since the program is small (just the .exe and few dlls) I wanted to keep it portable and not to depend on the registry. That's why I wanted to use the ROT.

The File Moniker does not need to be a real _file path_. It can be any string you want. In my case I create File Moniker `vJoySerialFeeder.1` and register it in the ROT (see [here](https://github.com/Cleric-K/vJoySerialFeeder/blob/d005f3d3a28c90204404ff11c43a7d41af23f0a1/vJoySerialFeeder/ProcessInteraction/ComAutomation.cs#L311)). This string looks like (and is) the ProgID, but it could be anything else.

To catch all possible scenarios I also register by class id (see [here](https://github.com/Cleric-K/vJoySerialFeeder/blob/d005f3d3a28c90204404ff11c43a7d41af23f0a1/vJoySerialFeeder/ProcessInteraction/ComAutomation.cs#L314)) and also as out-of-process object - which requires the registry in all cases (see [here](https://github.com/Cleric-K/vJoySerialFeeder/blob/d005f3d3a28c90204404ff11c43a7d41af23f0a1/vJoySerialFeeder/ProcessInteraction/ComAutomation.cs#L317)) - that's why I also provide a [.reg](https://github.com/Cleric-K/vJoySerialFeeder/blob/master/Docs/COM.reg) file if this is really needed.

So all in one I give the user three ways to get the object:
1. Get the object from the ROT by _name_ (which in my case I have chosen to be the same as the ProgID). In VBScript you can do this: `GetObject("vJoySerialFeeder.1")` and it works OK. VBScript creates File Moniker _directly_ from the string and looks for such object in the ROT. This was what made me initially believe that this is normal behavior and that's why I made the AHK patch.
2. Get the object from the ROT by clsid or ProgID which is registered in the registry - this is what AHK supports.
3. Use Create Object (for out-of-process object).

You can see the [AHK example](https://github.com/Cleric-K/vJoySerialFeeder/blob/master/Docs/COM.md#autohotkey) in the documentation. I've put there in comments the different ways to get the object.

To recap: my whole motivation was to allow getting the object from the ROT by ProgID _without_ need for the registry. The fact that VBScript can do this led me to the conclusion that this is normal behavior and that's why I decided to implement it in AHK.

In my program I have provided the user plenty of ways to get the object so this modification to AHK is not even that important.

It all boils down to the question if you think that getting active objects by File Monikers (in the way VBScript can do it) is correct and potentially useful behavior.
I just thought that this might have been overlooked in AHK and that's why wrote it. But if you think that it is unnecessary - that's perfectly fine.

Thank you!

Lexikos:
I was under the impression that GetObject was equivalent to ComObj**Get**. So does ComObjGet work with your application?

ComObjActive (when given a single string parameter) is equivalent to VBScript's Get**Active**Object.

Cleric-K:
Ahaa!
So this clarifies it. Yes - `ComObjGet` works just fine!

VBScript seems to have only [two functions for creating objects](https://technet.microsoft.com/en-us/library/ee156598.aspx) - `CreateObject` and `GetObject`. That's why I was under the impression that `GetObject` is just short for GetActiveObject. And that's why I did not even noticed AHK's `ComObjGet` - I thought that `ComObjActive` is all there is.

Ok, now there is really no use for this patch, I'll update the docs in my app to use `ComObjGet`.

Thank you! And sorry for wasting your time.

Lexikos:
> VBScript seems to have only two functions for creating objects

GetActiveObject does not create objects. For an object to be active, it must already exist...

GetObject may create an object of there are none active, or perhaps just depending on the moniker.

Cleric-K:
Yes.

As I said before, what confused me is that in VB they have only `GetObject` function (no `GetActiveObject`) which seems to call somehow the C functions `GetActiveObject` and `CoGetObject`, depending on the arguments.
For example:
`GetObject("C:\docs\sheet.xls")` will start if needed the program associated with `.xls` files (I suppose it uses internally `CoGetObject`), while `GetObject(, "Excel.Application")` will return object only if excel instance is running (so probably it calls `GetActiveObject`).

Some info [here](https://support.microsoft.com/en-us/help/288902/getobject-and-createobject-behavior-of-office-automation-servers).

Anyway,
Thank you for helping me figure this out.

PR #93, opened 2018-03-09
Adding option to BIF_Format, enabling placeholders to be used for more than one index.

HelgeffegleH:
Hello :wave:

Adding option to `BIF_Format`, enabling placeholders to be used for more than one index.

New (optional) syntax for placeholders: `{x+y:fmt}`, `{x-y:fmt}`, uses the placeholder for indices `x+y*k`, where `k = 0,1,...`. For example, `{1+1:fmt} ` formats each one of the input values, according to `fmt`.

# Examples
```autohotkey
; Simple example
msgbox format("{3-1} ", "a", "b", "c") ; c b a

; Table manipulation.
table := "
(
1	2	ab
4	5	cd
7	8	ef
10	11	gh
)"
arr := strsplit(table,["`t","`n"])

text := "View the array.`n`n"
msgbox text . format("{1+1}`n", arr*)

text := "Remove second column, swap first and third column.`n`n"
msgbox text . format("{3+3}`t{1+3}`n", arr*)

text := "Remove first column, swap second and third column. Turn up-side-down.`n`n"
l := arr.length()
msgbox text . format("{" l "-3}`t{" l-1 "-3}`n", arr*)		; Equivalent to: "{12-3}`t{11-3}`n"

; Random example

text := "Print the array backwards, swapping every other item.`n`n"
msgbox text . format("{11-2} {12-2}`t",arr*)
``` 

_Formatting_ is stopped on the first placeholder which causes `x+y*k` to be outside the range `1...aParamCount`, this avoids printing a literal placeholder containing a `+` or `-`, this is assumed to never be wanted. So you cannot print `{10+1}` even if less than `10` values were passed, in contrast to `{10}`. Limitation or feature, I consider it a feature, example, `format("{1} {10+1} {2}",1,2)` compare to `format("{1} {10} {2}",1,2)`.  Partial _work around_ `format("{1} {{}10+1} {2}",1,2)`.

Note: omitting _index_, eg, `{+y}`, is permitted but might not be generally useful or intuitive. However, detecting it as literal doesn't seem worth the bother, it does no harm as is.

Side effect: Specifying _index_ `0`, eg, `format("{0:U}a")` no longer formats the format string.

# Pros and cons.

Enables versatile and easy to use '`strJoin`' functionality, while still keeping the function simple and with small amount of added code. The repeated parsing is inefficient. `arr.format(opt)` would be more efficient than `format(...,arr*)`.

Thanks for reading.

### DIFF
--- a/source/script2.cpp
+++ b/source/script2.cpp
@@ -13126,8 +13126,6 @@ BIF_DECL(BIF_NumGet)
 	}
 }
 
-
-
 BIF_DECL(BIF_Format)
 {
 	LPCTSTR fmt = ParamIndexToString(0), lit, cp, cp_end, cp_spec;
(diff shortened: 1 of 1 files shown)


HelgeffegleH:
I don't remember needing this since I wrote it, and interest seems low, I close this for those reasons.

Note: 

> Side effect: Specifying index __0__, eg, `format("{0:U}a")` no longer formats the format string.

PR #96, opened 2018-04-29
Throw error when property is called

nnnik:
From what I have seen in the source obj.property() leads to the properties get code being called.
However I think that it should rather throw an error.
Also I want to ask if this coding style is the correct way to do this.

### DIFF
--- a/source/script.h
+++ b/source/script.h
@@ -221,6 +221,7 @@ enum CommandIDs {CONTROL_ID_FIRST = IDCANCEL + 1
 #define ERR_NO_OBJECT _T("No object to invoke.")
 #define ERR_UNKNOWN_PROPERTY _T("Unknown property.")
 #define ERR_UNKNOWN_METHOD _T("Unknown method.")
+#define ERR_PROPERTY_CALL _T("Cannot call a property")
 #define ERR_NO_GUI _T("No default GUI.")
 #define ERR_NO_STATUSBAR _T("No StatusBar.")
 #define ERR_NO_LISTVIEW _T("No ListView.")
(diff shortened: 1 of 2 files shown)


Lexikos:
This was already planned, as noted in [v2-thoughts](https://autohotkey.com/v2/v2-thoughts.htm). The reason `x.y(...)` calls property getters is that "`x.y(...) := z` is equivalent to `x.y[...] := z`".

Lexikos:
I was a bit hasty with the merge. The way you have implemented it is okay, but it is probably better to implement in `Property::Invoke`, where the `// Use mGet even if IS_INVOKE_CALL, for symmetry:` comment should also be changed.

nnnik:
Hmm yeah but that would trigger the wrong error if a property that has no get would get called.
So if do that we should probably move the responability for checking `canGet` and `canSet` inside `Property::Invoke`

Lexikos:
That's a good point, and since you brought up CanGet/CanSet, I thought of a more appropriate way of handling call. That is, not handling it (see commit b1f8e908). The purpose of CanGet/CanSet is to allow the invocation to recurse into the base object, in case it implements a property or meta-function which should handle the operation. If the property merely passes as if CanCall was false, a method in the base class can be called. For example:
```AutoHotkey
class Type1 {
    Member() {
        MsgBox "Member call"
    }
}
class Type2 extends Type1 {
    Member {
        get {
            MsgBox "Member get"
        }
    }
}
x := Type2.Member
x := Type2.Member()
```

If the method isn't implemented, the standard "Unknown method" error is thrown. This is also how the built-in objects should behave, but I think at the moment the only example of this is ClipboardAll (which happens to be the newest type).
```AutoHotkey
c := ClipboardAll()
x := c.Size
x := c.Size()  ; Unknown method
```

> So would it be okay then to implement the changes you mention in the AHK v2 thoughts?

Generally, yes. Specifically, changes to how properties vs. methods behave will likely be superseded, as I have been planning to make more fundamental changes that I haven't written about. One part of that is to define all built-in methods/properties via a base object or another form of metadata, to make the syntax and capabilities more consistent with user-created classes, and possibly also reduce code size.

The "What should probably be changed" in v2-thoughts are probably all relatively trivial, though. Feel free to implement them.

PR #97, opened 2018-05-01
I changed the behaviour of the file object completely

nnnik:
I changed the behaviour of the File Object with regards to methods and properties.
They now throw apropriate errors when being used in a wrong way.
For that I:

1. changed the structure of the if else ladder for finding the member ( henceforth member lookup )
It will now 1st check if it is a call event or a get/set event and then try to find a corresponding member method or propert depending on the type of event.
Calling a property or getting/setting a method will result in a "unkown method/property" error - as discussed on sunday.

2. added parameter and get/set checks to the member lookup.
Calling a method with the wrong kind of parameters will throw "too many" or "too few" errors depending on what was passed. Trying to set a read-only property will throw an "invalid use" error.
I don't think it's a beautiful solution. 
The only other option I see is defining it inside the switch case statement that defines the members behaviour. Tell me what you prefer I will keep future modifications consistent.

3. changed the behaviour of tell/seek and close
I made them full methods and also gave tell and seek their own members with their own results in the switch case latter.

### DIFF
--- a/source/TextIO.cpp
+++ b/source/TextIO.cpp
@@ -733,59 +733,74 @@ class FileObject : public ObjectBase // fincs: No longer allowing the script to 
 		--aParamCount; // Exclude name from param count.
 		LPTSTR name = TokenToString(*aParam[0]); // Name of method or property.
 		MemberID member = INVALID;
-
-		// Read' and Write' must be handled differently to support ReadUInt(), WriteShort(), etc.
-		if (!_tcsnicmp(name, _T("Read"), 4))
+		if (IS_INVOKE_CALL) //Check for methods
 		{
-			if (!name[4])
-				member = Read;
-			else if (!_tcsicmp(name + 4, _T("Line")))
-				member = ReadLine;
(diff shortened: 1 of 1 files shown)


nnnik:
Well we have the files but it's a pretty major change to the RegExMatch Object.

Lexikos:
I would prefer not to merge redundant commits. You can `reset` the branch back to the previous commit and `push --force` to change which commit you are proposing to merge. (In future it might be helpful to make the pull request from a separate branch, rather than messing with `alpha` in this way.)

For any changes I decide not to use for now but don't wish to delete immediately, I tend to "save" it is a branch.

RegExMatch demonstrates the more general problem of mixing properties with data, and the constraints imposed by `x.y` and `x["y"]` being equivalent. I intend to address this general problem, somehow.

nnnik:
I see thank you. I have only recently started using the git shell and I'm still quite new to this.
I will change the commit using the technique you've shown.

PR #98, opened 2018-05-02
Divided into methods and properties and added errors

nnnik:
I changed the behaviour of the File Object with regards to methods and properties.
They now throw apropriate errors when being used in a wrong way.
For that I:

1. changed the structure of the if else ladder for finding the member ( henceforth member lookup )
It will now 1st check if it is a call event or a get/set event and then try to find a corresponding member method or propert depending on the type of event.
Calling a property or getting/setting a method will result in a "unkown method/property" error - as discussed on sunday.

2. added parameter and get/set checks to the member lookup.
Calling a method with the wrong kind of parameters will throw "too many" or "too few" errors depending on what was passed. Trying to set a read-only property will throw an "invalid use" error.
I don't think it's a beautiful solution.
The only other option I see is defining it inside the switch case statement that defines the members behaviour. Tell me what you prefer I will keep future modifications consistent.

3. changed the behaviour of tell/seek and close
I made them full methods and also gave tell and seek their own members with their own results in the switch case latter.

Do you suggest to use git like this?

### DIFF
--- a/source/TextIO.cpp
+++ b/source/TextIO.cpp
@@ -713,14 +713,16 @@ class FileObject : public ObjectBase // fincs: No longer allowing the script to 
 		WriteLine,
 		NumReadWrite,
 		RawReadWrite,
+		Close,
+		PositionMethodGet,
+		PositionMethodSet,
 		LastMethodPlusOne,
 		// properties
 		Position,
 		Length,
 		AtEOF,
 		Handle,
-		Encoding,
-		Close
+		Encoding
 	};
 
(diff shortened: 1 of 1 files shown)


nnnik:
Forgot a pull request -.-

Lexikos:
That's all fine.

For ERR_TOO_MANY_PARAMS, the current policy is to show a *load-time* error for non-dynamic calls and to ignore surplus parameters for dynamic calls. However, this was mainly to allow a limited substitute for variadic calls, which can be done properly now. On the other hand, it is useful when defining callback functions; it allows one to define `MyMenuItem()` or `MyMenuItem(ItemName)` instead of `MyMenuItem(ItemName, ItemPos, Menu)`.

For the built-in objects, I can't see any use for passing surplus parameters. Prohibiting it improves error detection and may allow for future expansion.

PR #99, opened 2018-05-02
Suggestion regex object

nnnik:
This is how I would imagine the RegEx Object.
Make `Pos`, `Name`, `Value` etc. methods only and only allow get as a shorthand for `Value`.
That would put a clear border between what is a value and what isn't.
Bein able to differ between [] and . might also work out the same way.

### DIFF
--- a/source/TextIO.cpp
+++ b/source/TextIO.cpp
@@ -713,14 +713,16 @@ class FileObject : public ObjectBase // fincs: No longer allowing the script to 
 		WriteLine,
 		NumReadWrite,
 		RawReadWrite,
+		Close,
+		PositionMethodGet,
+		PositionMethodSet,
 		LastMethodPlusOne,
 		// properties
 		Position,
 		Length,
 		AtEOF,
 		Handle,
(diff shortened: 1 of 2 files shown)


Vismund-Cygnus:
Is there a reason that a regex object returns 'true' when passed into IsObject() and yet does not work in a for-loop? Shouldn't a for-loop work for all objects?

nnnik:
All objects that return an enumerator when their method _NewEnum is called support a for-loop.
For AHK built in objects thats actually only the associative array as far as Im aware.
And no it should not work for all objects.

HelgeffegleH:
Also gui object support for loop.

nnnik:
Well it may be good if regex objects support a for loop and were structured differently in general.
In my PR I focused on dividing between methods and properties and adding errors.
Maybe I should have made those 2 PRs too.

Lexikos:
I'm confused. If I hover over the word `Closes` in the commit above, it says "This commit closes pull request #99."

But the pull request wasn't closed automatically...

PR #100, opened 2018-05-03
Changing BIF_Integer to handle scientific notation (string).

HelgeffegleH:
Hello :wave: .

Issue example,
```autohotkey
Value := "1234e-2"
msgbox integer(Value)

Value := "12.34e2"
msgbox integer(Value)
```

Note, from the documentation,
>Any fractional part of Value is dropped, equivalent to `Value < 0 ? Ceil(Value) : Floor(Value)`

It is currently not true, example

```autohotkey
Value := "9223372036854775807.0"
msgbox Value < 0 ? Ceil(Value) : Floor(Value)
msgbox integer(Value)
```

It is true in this branch, however, if it is desired that
```autohotkey
integer("9223372036854775807.0") == 9223372036854775807
```
be true, another fix is needed, then also consider that eg integer(`"922337203685477580.7e1")` should probably work the same, which implies more work, I guess.

### DIFF
--- a/source/script2.cpp
+++ b/source/script2.cpp
@@ -14166,8 +14166,10 @@ BIF_DECL(BIF_FloorCeil)
 
 BIF_DECL(BIF_Integer)
 {
-	Throw_if_Param_NaN(0);
-	_f_return_i(ParamIndexToInt64(0));
+	ExprTokenType token;
+	if (ParamIndexToNumber(0, token))
+		_f_return_i(token.symbol == SYM_INTEGER ? token.value_int64 : token.value_double);
+	_f_throw(ERR_TYPE_MISMATCH);
 }
 
 


Lexikos:
I did not have the limitations of Floor/Ceil/Round in mind when I wrote that. I suppose "equivalent to" should be re-worded. The purpose was to show that Integer() does not round toward zero like Round().

It is probably better with this new behaviour, even with the apparent loss of precision. However, implementing it in Integer() leaves the problem that every other function calling ParamIndexToInt64 or TokenToInt64 still does not recognize scientific notation.

I suppose that a better solution would be for TokenToInt64 to use a custom number-parsing function which can deal with both types, including scientific notation with integers (currently the expression `1e3` produces a `Float`).

HelgeffegleH:
> However, implementing it in Integer() leaves the problem that every other function calling ParamIndexToInt64 or TokenToInt64 still does not recognize scientific notation.

I originally implemented the fix by detecting scientific notation in `tcstoi64_o` and calling `_tstof`, casting to `__int64`. The benefit was that it _fixed_ all functions, eg, `setkeydelay "1.0e1"` sets the keydelay to `10` rather than (as now) `1`. But `integer()` produced inconsistent results from, eg, `"922337203685477580.7e1"` and `"9223372036854775807.0"`. Besides that, it is inefficient since scientific notation is already detected in `IsNumeric`, and then of course `_tstof` would detect it once more. And I couldn't come up with any function where a sci.not. string needs to be supported, I concluded one should call `integer()` when needed :wink:. However, thinking about it now, I guess it makes sense for `format` to support it.

> currently the expression `1e3` produces a `Float`

That is a good point, I didn't consider that. Should `0.1e3` be an integer?

Cheers :coffee:

HelgeffegleH:
This is superseded by #101. I'm closing this for that reason.

Lexikos:
> it is inefficient since scientific notation is already detected in IsNumeric

*Inefficient* use of IsNumeric is probably the most common: the string is often parsed to determine if it is numeric, and then parsed again by the appropriate conversion function. That's why I mentioned "a custom number-parsing function which can deal with both types".

PR #101, opened 2018-05-04
Changing details of scientific notation handling.

HelgeffegleH:
Hello :wave: 

This supersedes #100 .

### New

Numbers written in scientific notation produces the same result as the AHK expression `x*10**y`, where `y` is an integer. Consequently, if

* `x` is an integer and `y` is greater or equal to zero, the result is an integer. (`-0` equals `0`)
* `x` is a float, the result is float.

Further, any _float string_ converted to an integer produces the same number as if the float was written as a _pure float_ in the script source code and converted, eg,

```autohotkey
integer("9223372036854775807.0") == integer(9223372036854775807.0)
```
is `true`.

__Known limitation:__ If a positive exponenent (`y`) is so large that it _wraps around_ (__int64) to become negative that is not detected and it is considered an integer, but the corresponding ahk expression `x*10**y` will produce a float. Similarily, if a negative exponent _wraps around_ to become positive, the number in scinentific notation produces a float and the AHK expression an integer. That is scientific notation has the same limitation as the expression `x*10**y` but they are _swapped_.

The problem can only be seen if `y > 9223372036854775807` (in `xEy` or `x*10**y`). Maybe integers written as `>9223372036854775807` should be detected as errors? For example,
```autohotkey
a:=9999999999999999999999999999999999999999999999999999999999999999
```
is valid but is unlikely to be written for the purpose of producing `-1`.

Cheers.

### DIFF
--- a/source/util.cpp
+++ b/source/util.cpp
@@ -963,7 +963,7 @@ __int64 tcstoi64_o(LPCTSTR buf, LPCTSTR *endptr, int base)
 		// Check for scientific notation and float.
 		// Doing it here because, presumably, most often *p == 0, and the calls to _tcschr are avoided.
 		LPCTSTR e_char = 0;
-		if (*p && *p == '.' || ((e_char = _tcschr(p, 'e')) || (e_char = _tcschr(p, 'E'))))
+		if ( *p && ( *p == '.' || (e_char = _tcschr(p, 'e')) || (e_char = _tcschr(p, 'E')) ) )
 		{
 			BOOL is_float = TRUE;
 			if (e_char) // implies no '.', so it is an integer in sci. notation, eg, 1e3 or a float like 1e-3.
--- a/source/util.cpp
+++ b/source/util.cpp
@@ -19,7 +19,7 @@ GNU General Public License for more details.
 #include <gdiplus.h> // Used by LoadPicture().
 #include "util.h"
 #include "globaldata.h"
-
+#include "qmath.h" // for qmathpow in tcstoi64_o
 
 int GetYDay(int aMon, int aDay, bool aIsLeapYear)
 // Returns a number between 1 and 366.
@@ -344,12 +344,15 @@ SymbolType IsNumeric(LPCTSTR aBuf, BOOL aAllowNegative, BOOL aAllowAllWhitespace
 	if (is_hex)
 		aBuf += 2;  // Skip over the 0x prefix.
 
-	// Set defaults:
-	BOOL has_decimal_point = false;
-	BOOL has_exponent = false;
+					// Set defaults:
 	BOOL has_at_least_one_digit = false; // i.e. a string consisting of only "+", "-" or "." is not considered numeric.
-	int c; // int vs. char might squeeze a little more performance out of it (it does reduce code size by 5 bytes). Probably must stay signed vs. unsigned for some of the uses below.
+	BOOL has_decimal_point = false;
 
+	BOOL has_exponent = false;			// For scientific notation.
+	BOOL exponent_is_negative = FALSE;	// Negative exponents will be considered float.
+	BOOL exponent_is_zero = FALSE;		// Needs to be detected so that xe-0 can be detected as an integer, consitent with x*10**-0 (AHK expr.)
+
+	int c; // int vs. char might squeeze a little more performance out of it (it does reduce code size by 5 bytes). Probably must stay signed vs. unsigned for some of the uses below.
 	for (;; ++aBuf)
 	{
 		c = *aBuf;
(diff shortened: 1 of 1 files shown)


Lexikos:
> `if (*p && *p == '.' || ((e_char = _tcschr(p, 'e')) || (e_char = _tcschr(p, 'E'))))`

I think you mean `if (*p && (*p == '.' || (e_char = _tcschr(p, 'e')) || (e_char = _tcschr(p, 'E'))))`. `*p && *p == '.'` is the same as `*p == '.'`, since `'.' != 0` and `&&` has higher precedence than `||`.

nnnik:
Yeah but 0xFFFF FFFF FFFF FFFF might be written to get -1 or rather it's binary representation.

HelgeffegleH:
> I think you mean __...__

Yes, thanks. I will change it.

> I mentioned "a custom number-parsing function which can deal with both types".

[(src)](https://github.com/Lexikos/AutoHotkey_L/pull/100#issuecomment-386776860). I did not mean to ignore this. I thought this was a simple way to do it, with little added code, and I think the performance impact on most scripts will be minimal. But for converting a float string to integer it is a bit __inefficient__ :heavy_check_mark: .

> Yeah but 0xFFFF FFFF FFFF FFFF might be written to get -1

Yes, that is fine, my comments was in regards to base 10 integers only. For a number written in hex form the corresponing error would be for something like `0xffffffffffffffffffffffffffffffffffffffffffffffff`. I'm certainly not passionate about detecting such _errors_.

Cheers :coffee:

Lexikos:
Some issues:
  - _tcschr is unnecessary, since `p` already points to the first non-digit character, which must be `.`, `e` or the null-terminator. This has the effect of allowing strange things like `1 raise 2`.
  - Spaces and tabs are implicitly permitted after `e` by the recursive call, but this is not valid.
  - `endptr` is set to the `e` if the exponent is positive, so callers which use it to validate the number see scientific notation as invalid. I suppose that it should pass a local variable to the endptr parameter of the recursive call, and set `p` only when `exponent >= 0 && *the_local_endptr`.
  - `endptr` is not set at all if the function returns early due to `is_float`, causing undefined behaviour.
  - `ControlParseOptions` parses the option value as float if parsing as int fails. If `endptr` was set for floating-point values, it would never fall back to float parsing (i.e. only the truncated `(float)option_int` value would be available). Perhaps it would be suitable to leave `endptr` set to the `.` (i.e. indicate it's not a valid *integer*, but still return the result of string->float->int conversion). In that case we can keep using `_tstof` rather than `_tcstod` (which would be needed for proper detection of endptr).

```AutoHotkey
OnMessage(0x8000, (w) => w)

TestEndptr "1e2"  ; expected: 100, actual: Invalid
TestEndptr "2e 2"  ; expected/actual: Invalid
TestEndptr "3 raise2"  ; expected/actual: Invalid
; TestEndptr "4.e2"  ; expected: 400, actual: CRASH

TestCoerce "5e2"  ; expected/actual: 500
TestCoerce "6e 2"  ; expected: 6, actual: 600
TestCoerce "7 raise2"  ; expected: 7, actual: 700
TestCoerce "8.e2"  ; expected/actual: 800

TestCoerce(n) {
    A_EventInfo := n
    MsgBox A_EventInfo
    A_EventInfo := 0
}
TestEndptr(n) {
    DetectHiddenWindows 1
    try
        MsgBox SendMessage(0x8000, n,,, "ahk_id " A_ScriptHwnd)
    catch e
        MsgBox e.Message
}
ExitApp
```

HelgeffegleH:
Hello :wave:

Thanks for the corrections. Missing the `endptr` is impressive :woman_facepalming: .

`1.` and `1.e1` and such being valid is news to me, I see it is standard for the `XXXtof` functions.

I will try to fix this as soon as I can.

Cheers.

HelgeffegleH:
First, a correction regarding,

> Numbers written in scientific notation produces the same result as the AHK expression `x*10**y`, where `y` is an integer.

It also needs `x` to be an integer to be true.

### Update.

`endptr`  is now set to the `'.'` if such a decimal point is present. If the string is a float due to a negative exponent, the `endptr` is set to the `'e'` or `'E'`. This seems to work well with `ControlParseOptions`, specifically the `R` option.

To avoid `"1e1e1"` being interpreted as `1e10` due to the trailing `"1e1"` being interpreted as `10` in the recursive call, I pass `-1` for the `base` parameter to indicate that scientific notation and float should not be considered. It is a bit tacky, since it is an obvious patch of something which wasn't considered from the beginning, maybe an additional parameter is more appropriate :thinking: .

Regarding your demo script, I have assumed that _expected_ doesn't mean _desired_ in all cases, for example,
```autohotkey
TestEndptr "4.e2"  ; expected: 400, actual: CRASH (my assumption: desired: invalid (because it is a float))
; cf.
TestEndptr "400.0"  ; desired: invalid
```

Cheers.

Lexikos:
@HelgeffegleH in #129, you wrote:
> I think it is probably best to always treat scientific notation as float, and #101 should be closed

Why?

My answer would be: It probably doesn't make sense to differentiate based on the presence of a decimal point, since a string without `.` can still produce a fraction and a string with `.` can produce a whole number. For instance, `1.1e6` produces a whole number and `99e-2` produces a fraction.

> (But the problem with `Integer(sci_not_string)` needs to be remembered).

When you refer to it that way, it's easy to forget that the problem applies to most functions which accept an integer, not just Integer().

I think that the parts of the code that require an integer must either always reject floating-point values, or always interpret them properly. The latter seems better to me, so e5776c56 changes ATOI, etc. to support scientific notation.

Some observations can be made about the conversion:
  - If the exponent is zero or not present, the result is the same as if conversion stopped at `.`/`e` (whichever is first), as before.
  - For a negative exponent *x*, the result is divided by `10**Abs(x)`, effectively discarding the last *x* digits left of `.`/`e`.
  - For a positive exponent *x* following *f* fractional digits, the result is the same as if we ignored the decimal point, parsed `d := Min(x,f)` additional digits as part of the integer, and multiplied the result by `10**(x-d)`.

So there's no need for conversion to/from `double` (which could cause a loss of precision). Instead, I utilize the above observations. The only new limitation I am aware of is with abnormal cases like `100000000000000000000e-1`, where the base number would fit in `double` but not `unsigned __int64`. (It would be possible to work around this by detecting the negative exponent first, and reducing the number of digits parsed by that many. However, I do not think it would be worthwhile.)

I chose to use "wrapping overflow" (rather than "saturating overflow") since it is most consistent with integer expressions and `Integer(n)` where n has no exponent. Parsing as float and then converting to integer would produce `_I64_MIN` for out of range values, since this is how `(__int64)(double)n` behaves. However, I found that this is [actually undefined behaviour](http://blog.frama-c.com/index.php?post/2013/10/09/Overflow-float-integer).

I decided that tcstoi64_o() should not be tasked with both identifying *integer strings* and parsing *numeric strings* to `__int64`, so there are now two separate functions, `nstrtoi64` and `istrtoi64`. There is no net increase in code size since some of the parameters were removed (this reduces the size of all call sites, which are mostly inline calls via ATOI, etc.).

PR #102, opened 2018-06-14
StdIn Scripts and working directories.

gbln:
Allow scripts run from StdIn to use the initial working directory as the parent dir for the local library, to retrieve auto included functions.

Current implementation does not allow auto including. Since the behavior is to use BIV_ScriptDir, but the script's directory is empty when run from StdIn and thus fail (non-existent library folder).

This change also lets StdIn scripts preserve A_WorkingDir as A_InitialWorkingDir since the beginning without auto-changing it to A_ScriptDir which is empty. So another script/program/caller can specify a working directory and remain valid.

One of the reasons is allow custom ExecScript functions (from the documentation) to use the same local library as the calling script.
Another reason is to have the option (for the general public i guess) to implement in Flipeador's Ahk2Exe refactor for v2, the ability to preprocess scripts and syntax check them without having to use tmp files. Using stdout and stderr for this to forward iLib to stdout and be able to detect files that need to be included.

/iLib * currently fails to detect files in local library due to reason mentioned above (BIV_ScriptDir is empty) if script is *


Lexikos:
Thank you for the detailed explanation.

The base commit passes NULL to BIV_ScriptDir in order to determine the length prior to writing anything into the buffer. Although long paths aren't supported very well in this commit, it *is* possible for mFileDir to exceed MAX_PATH (with the `\\?\` prefix), in which case your call to BIV_ScriptDir will overflow the buffer.

Commit 88f1811f (not yet merged) removes the reliance on fixed buffer sizes, but it conflicts with this pull request. 

If the parent script/process may be expected to set the working directory to the base directory for the lib and other resources, perhaps it would be appropriate for A_ScriptDir to contain the initial working directory. Merely replacing `//mFileDir = _T(""); // Already done by the constructor.` with `mFileDir = g_WorkingDirOrig;` would work, and should be more future-proof. In that case, the local lib would work as you wish, and the documented path (`A_ScriptDir "\Lib\"`) remains accurate for both normal and stdin scripts.

> This change also lets StdIn scripts preserve A_WorkingDir as A_InitialWorkingDir since the beginning without auto-changing it to A_ScriptDir which is empty.

I think A_WorkingDir could only differ from A_InitialWorkingDir if #Include *Directory* was used to change the working directory. `SetCurrentDirectory(mFileDir)` has no effect when mFileDir is empty; it simply fails.

gbln:
Oh, didn't know you had reworked that whole section already on the test branch. Thank you. I completely missed that mFileDir line. I reverted the previous change, used what you suggested and it works fine. I'll go ahead and close this, since it seems to have brought on the gitignore changes I made too, instead of just the relevant file.

VISIONZJR:
Hello,
I am totally new to this community. I only ask a question once, to learn
proper eticate.  If I had a question, would I ask as reply to theese
messages or do I/should I post to another location.  Just wondering for
future reference.
Thanks
Visionz

derz00:
Please ask questions at the forums. There are many helpful people there. https://autohotkey.com/boards

Lexikos:
Commit 0cd12994 changes the working directory for stdin scripts.

PR #103, opened 2018-06-15
Adding CaseSensitive parameter to StrReplace.

HelgeffegleH:
Hello :wave: 

New:

```autohotkey
ReplacedStr := StrReplace(Haystack, SearchText [, ReplaceText, CaseSensitive := false, OutputVarCount, Limit := -1])
```

Implementation is copied from `BIF_InStr`.

Reason for change: convenience.

### DIFF
--- a/source/script.cpp
+++ b/source/script.cpp
@@ -227,7 +227,7 @@ FuncEntry g_BIF[] =
 	BIF1(StrLen, 1, 1),
 	BIFn(StrLower, 1, 2, BIF_StrCase),
 	BIFn(StrPut, 1, 4, BIF_StrGetPut),
-	BIF1(StrReplace, 2, 5, {4}),
+	BIF1(StrReplace, 2, 6, {5}),
 	BIF1(StrSplit, 1, 4),
 	BIFn(StrUpper, 1, 2, BIF_StrCase),
 	BIF1(SubStr, 2, 3),
--- a/source/script2.cpp
+++ b/source/script2.cpp
@@ -5699,16 +5699,22 @@ BIF_DECL(BIF_StrReplace)
 	_f_param_string(source, 0, &length);
 	_f_param_string(oldstr, 1);
 	_f_param_string_opt(newstr, 2);
-	Var *output_var_count = ParamIndexToOptionalVar(3); 
-	UINT replacement_limit = (UINT)ParamIndexToOptionalInt64(4, UINT_MAX); 
 	
+	// The code for setting string_case_sense is copied from BIF_InStr, except for the parameter number. See BIF_InStr for further comments on this part.
+	StringCaseSenseType string_case_sense = (StringCaseSenseType)(!ParamIndexIsOmitted(3) && ParamIndexToBOOL(3));
+	if (g->StringCaseSense != SCS_INSENSITIVE && string_case_sense == SCS_INSENSITIVE) // Ordered for short-circuit performance.
+		string_case_sense = SCS_INSENSITIVE_LOCALE;
+	
+	Var *output_var_count = ParamIndexToOptionalVar(4); 
+	UINT replacement_limit = (UINT)ParamIndexToOptionalInt64(5, UINT_MAX); 
+	
+
 	// Note: The current implementation of StrReplace() should be able to handle any conceivable inputs
 	// without an empty string causing an infinite loop and without going infinite due to finding the
 	// search string inside of newly-inserted replace strings (e.g. replacing all occurrences
 	// of b with bcd would not keep finding b in the newly inserted bcd, infinitely).
 	LPTSTR dest;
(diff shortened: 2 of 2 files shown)


Lexikos:
This seems like it could be useful, but I wonder: how often would it be used, and how often are the OutputVarCount and Limit parameters used?

It is already possible to control the case-sensitivity of StrReplace (though perhaps not in the most convenient way), whereas OutputVarCount and Limit provide functionality that would otherwise require either a custom replacement function or RegExReplace. The current parameter order is identical to RegExReplace, which aides memorization.

I couldn't find any instance of StrReplace with more than three parameters in any of my scripts, but I know I have used OutputVarCount (or at least ErrorLevel with StringReplace). I do not recall having ever used StringCaseSense with StrReplace/StringReplace, although that might be because I am comfortable with regex. (I'm not sure that I've used StringCaseSense at all.)

HelgeffegleH:
Hello.

>  I wonder: how often would it be used

In one sense it will always be used, that is, since not specifying the parameter is equivalent to passing `false`. I think it is very common that people do not set `stringcasesense 'off'`, even when this is the desired setting, just because this is default unless changed in auto-exec. section, that is a mistake for a general/shared function. Further, I would probably often specify `true` to avoid using the (much slower) locale obeying function when replacing something like `','`  with `'|'`, where neither locale nor case matters. That would probably be a rarely significant  _optimisation_ though.

> I do not recall having ever used StringCaseSense with StrReplace/StringReplace

I think `stringcasesense` is rarely used with `strreplace`, because, either the code author doesn't realise that the replacement will be case sensitive if used in a script where `stringcasesense` is `'on'`, that is, it is happy with the default `stringcasesense 'off'`, or the author is aware and then uses  `regexreplace` with or without the `i` option, since it is more succinct than setting and restoring `stringcasesense`. Personally, I use `regexreplace` sometimes just because of this, and sometimes I forget or just do not care about the implications of `stringcasesense` and use `strreplace`. Some of my code most certainly will not work well if put in a script where `stringcasesense 'on'` has been set in the auto-exec. section.

Regarding the parameter order, I'm quite neutral to it. I made a guess that the new parameter would be specified more often than the others. Whenever I use the `OutputVarCount ` or `Limit` parameters for either `regexreplace` or `strreplace`, I need  to check the docs for the order anyways. 


Cheers.

jeeswg:
I would add StartPos (consistent with RegExMatch) and CaseSensitive parameters:
[StrReplace() should have option to replace from rear of string](https://autohotkey.com/boards/viewtopic.php?f=13&t=52534&p=267962#p267962)
Or possibly turn Limit into Options (Limit/CaseSensitive, and maybe StartPos).

jeeswg:
Here's a possible solution for StrReplace.

# Proposals

modify StrReplace (before/after) (2 new params, 1 swap):
`ReplacedStr := StrReplace(Haystack, SearchText [, ReplaceText, OutputVarCount, Limit := -1])`
`ReplacedStr := StrReplace(Haystack, SearchText [, ReplaceText, CaseSensitive := false, Limit := -1, StartingPos := 1, OutputVarCount := ""])`

introduce StrCount (AHK v1/v2) (params match InStr, although StartingPos could be omitted until fully discussed/scrutinised):
`Count := StrCount(Haystack, SearchText [, CaseSensitive := false, StartingPos := 1])`

# Parameters

1 Haystack
2 SearchText
3 ReplaceText
4 (change) before: OutputVarCount, after: CaseSensitive
5 Limit
6 (new) StartingPos
7 (new) OutputVarCount

# Rationale

* only 1 param modified
* the first 6 params (except param 4) match RegExReplace (for easy transfer)
* case insensitivity matches InStr (case insensitive by default)

for reference (InStr/RegExReplace):
`FoundPos := InStr(Haystack, Needle [, CaseSensitive := false, StartingPos := 1, Occurrence := 1])`
`NewStr := RegExReplace(Haystack, NeedleRegEx [, Replacement := "", OutputVarCount := "", Limit := -1, StartingPosition := 1])`

I wrote the 2 functions as custom AHK functions and they neatly solved all my replace/count/StringCaseSense woes.
I was able to replace multiple separate custom functions into one StrReplace function.
Writing new code is easier and more intuitive.

# Conversion Cost Re. The Count Parameter

[I had 73 of these]
e.g. modify old code (before/after: count and replace) (add 3 commas):
`vText := StrReplace(vText, vBefore, vAfter, vCount)`
`vText := StrReplace(vText, vBefore, vAfter,,,, vCount)`

[I had 40 of these]
e.g. modify old code (before/after: count only):
`StrReplace(vHaystack, vNeedle, "", vCount)`
`vCount := StrCount(vHaystack, vNeedle)`

[I had 17 of these]
e.g. modify old code (before/after: count and increment):
`StrReplace(vHaystack, vNeedle, "", vCount), vCount++`
`vCount := StrCount(vHaystack, vNeedle) + 1`

PR #104, opened 2018-06-24
Adding binary operator: !== and changing !=.

HelgeffegleH:
Hello :wave: .

New:
`operand1 !== operand2`,
`x !== y` is equivalent to `!(x == y)`	(always case sensitive)
`x != y` is changed to be equivalent to `!(x = y)` (never case sensitive)

Example:

```autohotkey
msgbox 'a' !== 'A' ; 1
stringcasesense 'on'
msgbox 'a' != 'A' ; 0, previously 1
```

Reason:
Convenience, readability (subjective), predictability. Probably some performance gain in eg, `(x !== y)` vs `!(x == y)`

Cheers.

### DIFF
--- a/source/defines.h
+++ b/source/defines.h
@@ -180,7 +180,7 @@ enum SymbolType // For use with ExpandExpression() and IsNumeric().
 #define IS_SHORT_CIRCUIT_OPERATOR(symbol) ((symbol) <= SYM_AND && (symbol) >= SYM_IFF_THEN) // Excludes SYM_IFF_ELSE, which acts as a simple jump after the THEN branch is evaluated.
 #define SYM_USES_CIRCUIT_TOKEN(symbol) ((symbol) <= SYM_AND && (symbol) >= SYM_IFF_ELSE)
 	, SYM_IS, SYM_IN, SYM_CONTAINS
-	, SYM_EQUAL, SYM_EQUALCASE, SYM_NOTEQUAL // =, ==, <> ... Keep this in sync with IS_RELATIONAL_OPERATOR() below.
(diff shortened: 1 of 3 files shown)


Lexikos:
I like that `=` and `!=` would be more symmetrical, although `<=` and `>=` are similarly combined with `=` but are still dependent on StringCaseSense.

HelgeffegleH:
Hello.

> `<=` and `>=` are similarly combined with `=` but are still dependent on `StringCaseSense`.

It might seem natural to implement `<==` and `>==` to be case sensitive and change `<=`  and `>=`  to  never be case sensitive. However, I had something else in mind, I had planned, but not finished planning, to implement a `strCompare` function to replace `<`,  `>`,  `<=` and  `>=` for string compairsons, they would then only do numeric compairsons. I do not think `<, >, ...` are useful for string compairsons in general. Typically, there is a loss of information which require an extra compairsion, eg,

```autohotkey
if stra < strb
	; A
else if stra > strb
	; B
else
	; C
```
vs.
```autohotkey
result := strCompare(stra, strb)
if result < 0
	; A
else if result > 0
	; B
else
	; C 
```
Also, it has already become clear that the ability to have both string and number types on either side of the operator causes confusion. That is, eg, `vara > varb` may be

* __1)__ string comparison if one variable is a pure number but the other is a _non-numeric_ string
* __2)__ string comparison if both variables are strings, numeric or not.
* __3)__ numeric comparison if one of the variables is a pure number and the other is a numeric string, or both are pure numbers.

`strCompare` would have option to select case sensitivity or locale, I think it would be nice to be able to choose which locale to use too. Maybe something like `StartingPos` and/or `Length` parameters would be useful too, I leave it to the readers imagination to figure what I mean.

`<, >, ...` would throw for non numeric operands..

In contrast, I think `= == != !==` are very useful for string compairson. Also, `==` and `!==` does `memcmp` which I guess `<==` and `>==` would do too, although I'm not sure that is very useful. 

Cheers.

PR #105, opened 2018-06-24
Removing the <> operator.

HelgeffegleH:
Hello :wave:.
Example,
```autohotkey
(0<>0) ; Load-time error
```

Reason:
There is no obvious reason why both `!=` and `<>` are needed, keeping `!=` because it seems prefered by most. 

If it wasn't available, we wouldn't add it for the same reasons we wouldn't add, eg, `!<` meaning `>=`

Cheers.

### DIFF
--- a/source/defines.h
+++ b/source/defines.h
@@ -180,7 +180,7 @@ enum SymbolType // For use with ExpandExpression() and IsNumeric().
 #define IS_SHORT_CIRCUIT_OPERATOR(symbol) ((symbol) <= SYM_AND && (symbol) >= SYM_IFF_THEN) // Excludes SYM_IFF_ELSE, which acts as a simple jump after the THEN branch is evaluated.
 #define SYM_USES_CIRCUIT_TOKEN(symbol) ((symbol) <= SYM_AND && (symbol) >= SYM_IFF_ELSE)
 	, SYM_IS, SYM_IN, SYM_CONTAINS
-	, SYM_EQUAL, SYM_EQUALCASE, SYM_NOTEQUAL // =, ==, <> ... Keep this in sync with IS_RELATIONAL_OPERATOR() below.
+	, SYM_EQUAL, SYM_EQUALCASE, SYM_NOTEQUAL // =, ==, ... Keep this in sync with IS_RELATIONAL_OPERATOR() below.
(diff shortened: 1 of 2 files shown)


Lexikos:
> If it wasn't available, we wouldn't add it for the same reasons we wouldn't add, eg, !< meaning >=

Your reasons are your own. I don't know what they are, so what you've written tells me very little.

`<>` and `>!` would be redundant if we have `!=` and `>=`, but redundancy itself is not a reason. Larger documentation and binary code is bad, though trivial. Having more symbols to learn for the same effect is bad. Having different ways of expressing a thought can be good for the author, but bad for others.

Unlike `!<`, `<>` has been in AutoHotkey forever (even before expressions). Basic and derived languages use it (and not `!=`). Users sometimes translate VBScript to AutoHotkey. Some users are already familiar with VBScript before they encounter AutoHotkey. Because of this, there is a little value in keeping `<>`.

There are different ways of thinking; `!<` as "not less than" might better represent the author's thoughts than "greater than or equal". Sometimes it is easier or more fitting to add logical negation rather than reversing each operator. Perhaps that is why some SQL variants have both `!<` and `>=`.

The majority of my experience has been with languages which use `!=`, so my bias is toward removing `<>`. Further rationalizing, I could say that the reasons to have only one set of symbols for each operator (e.g. less to learn) probably outweigh the reasons for having redundant operators. Also, "not`!` equal`=`" fits my way of thinking better than "less than`<` or greater than`>`".

On the US layout, I think `<>` (Shift and two adjacent keys) is easier to type than `!=` (two keys on opposite hands, with only one shifted).

HelgeffegleH:
Hello.

You make good points.

> Your reasons are your own. I don't know what they are, so what you've written tells me very little.

The sentence you quoted wasn't intended to convey my reason for removing `<>`, it was meant to spark the readers mind, encouraging it to think about why we should have two ways of doing `!=`, with the hope that it would conclude that there was no good reason  ✨ 💭. My main reason was given in the preceding sentence.

>  redundancy itself is not a reason.

In this case, and probably most, redundancy implies some cost, such as those you mentioned. Also, removing `<>` now, reserves it for future use.

> easier to type

On my non-US keyboard, typing `<>` is not as you describe, but still probably easier than `!=`, hadn't I been so used to typing `!=`. It is a good point, I will enjoy typing it, if it is used for something else 😉.


Cheers.

PR #107, opened 2018-08-11
Adding function StrCompare and disabling string compairsion of operators: < <= > >=

HelgeffegleH:
Hello 👋 

This is a branch from #104 , it is branched because if there is no interest in this, either #104 can be kept as is or it can be extended with `<==` and `>==`.

### StrCompare:

```autohotkey
Result := StrCompare(str1, str2 [, CaseSensitive := false])
```
* `str1`, `str2`, the strings to compare.

* `CaseSensitive`, the case sensitive setting to use.

`Result`, the result of the comparison:

* `< 0`,	if `str1` is less than `str2`.
* `0`,		if `str1` is identical to `str2`.
* `> 0`,	if `str1` is greater than `str2`.

The `CaseSensitive` parameter behaves as `BIF_InStr`.

__Reason__: No loss of information as when using the compare operators to compare strings, i.e., `< <= > >=`.

### Disabling string compairson for operators: `< <= > >=`

Non-numeric operands causes a `Type mismatch.` error.

__Reason__: Superseded by `StrCompare` which doesn't cause a loss of information. Simplifies the behaviour of the operators.

There was some comments [here](https://github.com/Lexikos/AutoHotkey_L/pull/104#issuecomment-400404357).

If there is an interest, I'll look in to making the proper adjustments and additions in the documentation (unless someone else wants to do it 😉 ).

Cheers.

### DIFF
--- a/source/defines.h
+++ b/source/defines.h
@@ -180,7 +180,7 @@ enum SymbolType // For use with ExpandExpression() and IsNumeric().
 #define IS_SHORT_CIRCUIT_OPERATOR(symbol) ((symbol) <= SYM_AND && (symbol) >= SYM_IFF_THEN) // Excludes SYM_IFF_ELSE, which acts as a simple jump after the THEN branch is evaluated.
 #define SYM_USES_CIRCUIT_TOKEN(symbol) ((symbol) <= SYM_AND && (symbol) >= SYM_IFF_ELSE)
 	, SYM_IS, SYM_IN, SYM_CONTAINS
-	, SYM_EQUAL, SYM_EQUALCASE, SYM_NOTEQUAL // =, ==, <> ... Keep this in sync with IS_RELATIONAL_OPERATOR() below.
+	, SYM_EQUAL, SYM_EQUALCASE, SYM_NOTEQUAL, SYM_NOTEQUALCASE // =, ==, <> and !=, !==... Keep this in sync with IS_RELATIONAL_OPERATOR() below.
 	, SYM_GT, SYM_LT, SYM_GTOE, SYM_LTOE  // >, <, >=, <= ... Keep this in sync with IS_RELATIONAL_OPERATOR() below.
 #define IS_RELATIONAL_OPERATOR(symbol) (symbol >= SYM_EQUAL && symbol <= SYM_LTOE)
 	, SYM_REGEXMATCH // ~=, equivalent to a RegExMatch call in two-parameter mode.
--- a/source/defines.h
+++ b/source/defines.h
@@ -141,7 +141,6 @@ enum ToggleValueType {TOGGLE_INVALID = 0, TOGGLED_ON, TOGGLED_OFF, ALWAYS_ON, AL
 	, TOGGLE_MOUSEMOVE, TOGGLE_MOUSEMOVEOFF};
 
 // Some things (such as ListView sorting) rely on SCS_INSENSITIVE being zero.
-// In addition, BIF_InStr relies on SCS_SENSITIVE being 1:
 enum StringCaseSenseType {SCS_INSENSITIVE, SCS_SENSITIVE, SCS_INSENSITIVE_LOCALE, SCS_INSENSITIVE_LOGICAL, SCS_INVALID};
 
 enum SymbolType // For use with ExpandExpression() and IsNumeric().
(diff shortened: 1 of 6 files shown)


nnnik:
I think it would be better to split of the addition of the function from the removal of the operators feature.

HelgeffegleH:
Perhaps,  but they are split by two commits and it doesn't make sense to _remove_ them unless there is a replacement. There is a small overlap in the implementation of `!==` and the last operator commit, otherwise I would have kept this separate.

Lexikos:
I agree with nnnik. If you do not split them, we must accept both changes or reject both (or pick out parts and not merge the full PR). As it is, I am unlikely to merge this pull request.

The operators do not cause a loss of information. They represent a specific relation. They are not ideal for sorting algorithms, but that is not to say that StrCompare is more ideal for every possible string comparison. StrCompare is more verbose, which is both inconvenient and reduces readability if the script really does just want to know if `A >= B`. For example, compare `A_AhkVersion >= "2."` with `StrCompare(A_AhkVersion, "2.") > 0`.

Even being familiar with this type of comparison function, I have to stop and think (and refer to the documentation) every time I use it with `< 0` or `> 0` (because that isn't often). It isn't nearly as intuitive as the standard operators.

StrCompare somewhat simplifies the creation of Sort callbacks which sort by a substring of the input. For fully user-defined sorting algorithms, it removes the need to perform multiple *string* comparisons for each pair of values, but you still need multiple comparisons to branch three ways; only now you have to store the result of a (more verbose) function call and compare it to zero, instead of comparing A to B. CPU efficiency is improved at the expense of human efficiency.

The problem with adding `<==` and `>==` is that it throws into question how `>` and `<` should behave, since they have no obvious counterpart. Perhaps an alternative is to instead add two [spaceships](https://en.wikipedia.org/wiki/Three-way_comparison), `<=>` and `<==>` (or just use StrCompare when case-sensitivity must be specified).

For some reason I was expecting to be able to override or enable the locale mode. Perhaps it should be considered (separately) for this and InStr.

Originally I think `If (a < b)` performed numeric comparison if both a and b were numeric **strings**, so to force a string comparison one had to do something like `If ("Str" a < "Str" b)`. In one of the alphas I changed it to perform string comparison if a and b are both strings (numeric or not), but in hindsight this was probably a bad idea. It can give contradictory results with mixed types:

    a := "10"
    b := 2
    c := "3"
    MsgBox a > b
    MsgBox b < c
    MsgBox a < c

It appears that JavaScript has this same issue. There are ways to make mixed-type comparisons consistent (which would probably entail a change in how string-to-string comparisons are performed), but the safest option is probably to prohibit them.

On the other hand, there are undoubtedly users who will expect `<` to perform arithmetic comparison regardless of the type of input.

HelgeffegleH:
Hello.

I will split them, in my reasoning I failed to consider that one might actually want both ways of comparing strings, I was biased by my own preference of course. But I'd rather have both than just the operators.

>The operators do not cause a loss of information. They represent a specific relation. They are not ideal for sorting algorithms, but that is not to say that StrCompare is more ideal for every possible string comparison. StrCompare is more verbose, which is both inconvenient and reduces readability if the script really does just want to know if A >= B. For example, compare A_AhkVersion >= "2." with StrCompare(A_AhkVersion, "2.") > 0.

Sure, the loss of information I'm referring to is limited to _behind the scenes_, there is no loss for the logic of the script or the documented behaviour of the operators. Readability, I think `StrCompare` is superior. When using variables, it is still clear that we do string comparison. With the operators, it is only clear when using one quoted non-numeric string or two quoted strings. I think this sort of string compairsons, `A_AhkVersion >= "2."`, is a good argument for disabling string compairson for the operators, they are dangerous. It will work in your example (for now :wink:), but eg, `A_OsVersion < '6.1'` fails to identify windows 10 as newer than windows 7 (I think). What we really want to achieve here is a numeric compairson.

>StrCompare somewhat simplifies the creation of Sort callbacks which sort by a substring of the input. For fully user-defined sorting algorithms, it removes the need to perform multiple string comparisons for each pair of values, but you still need multiple comparisons to branch three ways; only now you have to store the result of a (more verbose) function call and compare it to zero, instead of comparing A to B. CPU efficiency is improved at the expense of human efficiency.

Yes, I had such things in mind certainly. In the many cases though, the `sort` callback will probably just end with

```autohotkey
return StrCompare(substr_1, substr_2)
```

Of course there can be more elaborate versions.

>Even being familiar with this type of comparison function, I have to stop and think (and refer to the documentation) every time I use it with < 0 or > 0 (because that isn't often).

Same for me, with __emphasis__ on _because that isn't often_.

> Spaceships

Cool 😉 . For me, it looks like [iff](https://en.wikipedia.org/wiki/If_and_only_if) though, but I guess I could get used to it.

>For some reason I was expecting to be able to override or enable the locale mode. Perhaps it should be considered (separately) for this and `InStr`.

I think `0` for case insensitive, `1` for case sensitive and `2` for locale would be suitable for both `StrCompare` and `InStr` (and `StrReplace` and perhaps other future (changes to) functions which will have a `CaseSensitive` parameter). I'm not convinced it is a good idea that these functions depend on `stringcasesense`.

Regarding _verbosity_, an alternative name would be `StrCmp`. It is less descriptive though.

I will close this and return with two separate pull requests.

Cheers.

PR #108, opened 2018-08-12
Adding StrCompare function for comparing strings.

HelgeffegleH:
Split from #107.

PR #109, opened 2018-08-12
Disabling string compairson for operators: < <= > >=

HelgeffegleH:
Split from #107.

### DIFF
--- a/source/defines.h
+++ b/source/defines.h
@@ -181,6 +181,7 @@ enum SymbolType // For use with ExpandExpression() and IsNumeric().
 #define SYM_USES_CIRCUIT_TOKEN(symbol) ((symbol) <= SYM_AND && (symbol) >= SYM_IFF_ELSE)
 	, SYM_IS, SYM_IN, SYM_CONTAINS
 	, SYM_EQUAL, SYM_EQUALCASE, SYM_NOTEQUAL, SYM_NOTEQUALCASE // =, ==, !=, !==... Keep this in sync with IS_RELATIONAL_OPERATOR() below.
+#define IS_EQUALITY_OPERATOR(symbol) (symbol >= SYM_EQUAL && symbol <= SYM_NOTEQUALCASE)
 	, SYM_GT, SYM_LT, SYM_GTOE, SYM_LTOE  // >, <, >=, <= ... Keep this in sync with IS_RELATIONAL_OPERATOR() below.
 #define IS_RELATIONAL_OPERATOR(symbol) (symbol >= SYM_EQUAL && symbol <= SYM_LTOE)
 	, SYM_REGEXMATCH // ~=, equivalent to a RegExMatch call in two-parameter mode.
--- a/source/script_expression.cpp
+++ b/source/script_expression.cpp
@@ -882,7 +882,7 @@ LPTSTR Line::ExpandExpression(int aArgIndex, ResultType &aResult, ResultToken *a
 				left_is_pure_number = TokenIsPureNumeric(left, left_is_number);
 			// Otherwise, leave left_is' uninitialized as below will short-circuit.
 			if (  !(right_is_number && left_is_number)  // i.e. they're not both numeric (or this is SYM_CONCAT).
-				|| IS_RELATIONAL_OPERATOR(this_token.symbol) && !right_is_pure_number && !left_is_pure_number  ) // i.e. if both are strings, compare them alphabetically.
+				|| IS_EQUALITY_OPERATOR(this_token.symbol) && !right_is_pure_number && !left_is_pure_number  ) // i.e. if both are strings, compare them alphabetically if the operator supports it.
 			{
 				// L31: Handle binary ops supported by objects (= == != !==).
 				switch (this_token.symbol)
@@ -932,12 +932,7 @@ LPTSTR Line::ExpandExpression(int aArgIndex, ResultType &aResult, ResultToken *a
 									: lstrcmpi(left_string, right_string)); break;
 				case SYM_NOTEQUALCASE:	this_token.value_int64 = !((left_length == right_length)		// Same as SYM_EQUALCASE but inverted result, code copied for performance (assumed).		
 										&& !tmemcmp(left_string, right_string, left_length) ); break;
-				// The rest all obey g->StringCaseSense since they have no case sensitive counterparts:
-				case SYM_GT:			this_token.value_int64 = g_tcscmp(left_string, right_string) > 0; break;
-				case SYM_LT:			this_token.value_int64 = g_tcscmp(left_string, right_string) < 0; break;
-				case SYM_GTOE:			this_token.value_int64 = g_tcscmp(left_string, right_string) > -1; break;
-				case SYM_LTOE:			this_token.value_int64 = g_tcscmp(left_string, right_string) < 1; break;
-
+				
 				case SYM_CONCAT:
 					// Even if the left or right is "", must copy the result to temporary memory, at least
 					// when integers and floats had to be converted to temporary strings above.


Lexikos:
From #107 
>> Even being familiar with this type of comparison function, I have to stop and think (and refer to the documentation) every time I use it with < 0 or > 0 (because that isn't often).
>
> Same for me, with **emphasis** on *because that isn't often*.

So do we agree that StrCompare </>/<=/>= 0 is not intuitive and must be learned through frequent repetition?

I think your implication is that you would be using it more often, and therefore would learn to recognize it more readily. I do not think it can be trivialized in that way. I do not believe I would be using StrCompare often for any kind of comparison, and by extension I do not believe the average user would become accustomed to this kind of comparison. Even if they did, there would be a period for each new user where it is hard to understand.

That aside, I do not believe that your arguments for removing the operators thus far are sufficient.

HelgeffegleH:
Hello 👋 .

I agree that `StrCompare </>/<=/>= 0 ` is not intuitive. But it is not difficult to use.

>I think your implication is that you would be using it more often, and therefore would learn to recognize it more readily.

No, I would use `StrCompare` as often as I use  `< >`, and _that isn't often_. I never learn it, I look in the documentation when I use this kind of string compairson functions. I also always look in the documentation on what the sort callback should return, and every time I need to stop and think on how to translate, eg, `stra < strb` to `< > 0`, it is the exact same problem, and people manage it just fine. 

***

That aside, the point isn't that `StrCompare` is better than the operators, although I think it is. The point is that the ability to do string compairson with the operators implies that the, by far most common, use of the operators gets more complicated and error-prone, i.e., numeric compairsons.  If only numeric compairson is possible, one never has to look in the documentation on how to use this very common operator, and there is no risk of hard-to-find bugs due to accidently using a non-numeric string in a compairson, a likely real case would be a blank string.

The only reasonable cases for these operators, imo, are to only allow,

* __a)__ two pure numbers. numeric compairson,

together with,

* __b)__ two strings, numeric or not, string compairsion,

or only,

* __c)__ two numeric values, numeric compairson.

Compare to the current state which is,

* __d)__ pure number and numeric string, numeric compairson,

together with,

* __e)__ pure number and non-numeric string, string compairson,

plus  __a)__ and __b)__, this is by far less intuitive and hard to keep track of than the use of `StrCompare`.

I'm fine with __a)__ and __b)__ instead of __c)__ (this PR), but as I have understood, you value the ability to use numeric strings as numbers, hence I picked __c)__, which is also _consistent_ with math operators and functions. The `StrCompare` function is in line with the other string functions, that is, pure numeric input is converted to string. Hence, the changes #108 and #109 covers all possibilities, __a-e__, with non of the potential problems and better error detection 👍. 

I'm not passionate about having it exactly this way, but just that a change is needed. I would be interested in seeing any other suggestions, from anyone.

***

A few more comments on #107 

>now you have to store the result of a (more verbose) function call and compare it to zero, instead of comparing A to B. CPU efficiency is improved at the expense of human efficiency.  

With the function, reversing the result is as easy as multiply by `-1`, with the operators, you would also have to store the result or multiply each return value by `-1` if you wanted to reverse the result.

> For example, compare A_AhkVersion >= "2." with StrCompare(A_AhkVersion, "2.") > 0.

Even if you continue to support string compairson with operators, the better option for the above is, `A_AhkVersion.Major >= 2.0`, and similar for `A_OsVersion`.

***

Cheers ☕️ .

Lexikos:
> But it is not difficult to use.

It is more verbose and less intuitive, and requires referring to the documentation more often. In other words, it is more difficult to write, read and use. "Difficult" is subjective, and there are degrees.

> ... a likely real case would be a blank string.

That reminds me I have utilized this. In a loop finding the smallest number, `(n > min)` is true for every numeric `n` if `min` is empty. Conversely, `"N/A"` is always greater than any number. It avoids the need to know and use the minimum or maximum numeric value (which is actually negative or positive infinity, which one can produce with NumPut).

> The point is that [...] use of the operators gets more complicated and error-prone

I think you exaggerate the issue, and I disagree with this proposed solution. (But perhaps if we take this to the forum and a majority of other vocal users offer opinions aligned with yours, it will wear down my resistance to the idea.)

I chose the current behaviour because I felt it was intuitive at the time. I would not expect to have to refer to the documentation. Aside from the change regarding two numeric *strings* (which I've already explained and admitted might be a bad idea), v1 has had this behaviour for a long time, and I do not see it as a significant source of problems.

> With the function, reversing the result is as easy as multiply by -1, with the operators, you would also have to store the result or multiply each return value by -1 if you wanted to reverse the result.

I don't understand. If you're not already storing the result, there is no need to store it to reverse it. I was talking about (in what you quoted) simple comparison of two values, giving a boolean result. Reversing the result is even easier than multiplying by -1.

If you are multiplying by -1, it is because you have a three-way result; for that case, it was already clear that using something which gives a three-way result is better than combining multiple operators to get it. But even so, reversing the result means exactly the same amount of additional code either way.

>  the better option for the above is, `A_AhkVersion.Major >= 2.0`

Hypothetically, this expression would always be **true** (and therefore pointless), since `A_AhkVersion.Major` is an empty string on v1. If there is any one feature that should be backward-compatible, it is `A_AhkVersion`.

This option implies A_AhkVersion is one of:

  - an object which has a property `Major`
  - a new type of value, `version`, which has a property `Major`
  - a string, where all strings have a `Major` property

If A_AhkVersion is not a string, it must be implicitly convertible to string. That isn't yet possible for objects. This isn't necessary in the first place, because string comparisons work for A_AhkVersion, and I take care to preserve this. For instance, I chose `2.0-a001` over `2.0.00.00-a001` because `"2.0-a001" < "2.0."` (and with any suffix on the latter) or even `"2.0-a001" < "2.00"`.

Also, `A_AhkVersion.Major >= 2 && A_AhkVersion.Minor >= 1` is somewhat more verbose than `A_AhkVersion >= "2.1."`. In theory, features like overloaded operators and custom literals can rectify this; but we don't have those luxuries yet. The alternative is a function for version comparison, which we *don't need* for A_AhkVersion so long as we can use string comparison.

We might even avoid the issue by using a floating-point version number (just major and minor), but that can't take into account alpha/test versions.

nnnik:
The fact that "" compares to numbers successfully does not seem like an advantage. How many times would you actually want to compare the number to a string and how many times did that happen just because you had a bug.

For me the latter is far greater and I can even accurately tell the amount of times I did the former: once

Is it really necessary to overload one of the most commonly used operators with a feature that is rarely used and heavily impairs the ability to find bugs related to that operator?

But I also disagree with Helgefs removal suggestion. The feature is already implemented and coded - its fairly easy to use and rather intuitive. I wouldn't want to completely remove it.

I think it might be for the best if we split the operator into 2 parts: one number comparison and one string comparison.

We already use the ~ prefix for regexmatch. It could also be used to hunt at this string comparison:
```
Result := var1 ~< var2
```

HelgeffegleH:
Hello, I'm not sure the following is containing anything new, it might just be a rephrasing (but maybe a clarification) of what I have already tried to say.

> It is more verbose and less intuitive, and requires referring to the documentation more often. In other words, it is more difficult to write, read and use. "Difficult" is subjective, and there are degrees.

`StrCompare` is easier to use because it is clearly (not verbosely) doing string compairson, the operators are harder to use because of all the issues that have been mentioned. Again, it is not clear what type of compairson they do.  For it to be clear you have to have at least one quoted non-numeric string as one of the operands, or use two literal numbers, in which case you know you do numeric compairson, hence their use is impaired with variables. So the operators are the opposite of verbose, which is much worse. Even if you clearly name your variables to indicate their type and/or value, mistakes are not causing exceptions. To be safe with the operators, much more verbose code is needed, eg, 

```autohotkey
my_less_than(var1, var2) {
	if (type(var1) == 'Integer' || type(var1) == 'Float') && (type(var2) == 'Integer' || type(var2) == 'Float')
		return var1 < var2
	throw
}
```

> That reminds me I have utilized this

You have exemplified one of the only good candidates for built-in variables, constant values which can be resolved at load time, such as `A_MinInt, ...`.

> I think you exaggerate the issue

That is clear, it should also be clear that I think you underestimate the issue and exaggerate the usefulness of being able of to do string compairsons with the operators. 

> Aside from the change regarding two numeric strings (which I've already explained and admitted might be a bad idea), v1 has had this behaviour for a long time, and I do not see it as a significant source of problems.

It was not a bad idea, it solved the need to do `'str' . numeric_strA < 'str' . numeric_strB` for string compairson, which I hope we do not revert to, it should be obvious that `String(var1) < String(var2)` is meant to do string compairson, but it might not do that with the v1 behaviour. Still, it didn't address the unnecessary complexity of the operators' behaviour depending on operand type and value. Only rarely one wants to do `var < 'non_numeric_str'` which is the only actual clear and reasonable way of using the operators with strings (with the current behaviour). 

The greatest reliefs of leaving v1 in favour of v2 has been less of these kinds of quirks and smart behaviour / assumptions to keep track of, the addition of more exceptions, and the distinction between string and number types. My proposed changes are in line with this, imo.

>But perhaps if we take this to the forum and a majority of other vocal users offer opinions aligned with yours, it will wear down my resistance to the idea.

The topic was touched on, in a recent thread, no-one agreed with me, one agreed with you. I am not trying to wear down your resistance, I recognise the fact that your view on the language is based on much more experience and knowledge of both implementation and usage, I respect that.

***

>_reverse_, `-1`

I probably misunderstood what I quoted.

> a string, where all strings have a Major property

So now you need verbosity 😉 ? I meant it to be an object with a `Major` property (returning a pure number). 

>For instance, I chose 2.0-a001 over 2.0.00.00-a001 because "2.0-a001" < "2.0." (and with any suffix on the latter) or even "2.0-a001" < "2.00".

It looks messy to me.

> `A_AhkVersion.Major >= 2 && A_AhkVersion.Minor >= 1`

I like it very much 😉 .

> string comparisons work for A_AhkVersion, and I take care to preserve this. 

That is fine then, but `A_OsVersion` is still a problem, and probably much more relevant to compair. Being able to do string compairson with `A_AhkVersion` might only make users expect they can do the same things with the OS variable.

> If there is any one feature that should be backward-compatible, it is `A_AhkVersion`

It needs only be _backwards-compatible_ within ahk `>=2.0`, imo.

***

@nnnik, it is not a bad suggestion, I thought of having a separate set of operators too, but couldn't think of any good ones. It could also handle dependency on `stringcasesense`, where `~<<` would be case sensitive compairson, the sacrifice is being less clear on first sight. 

It also makes it harder to have compairsons with `A_AhkVersion` _backwards-compatible_ with v1. I'm not concerned about that though 😄 👍 .

I am biased against the prefix `~`, since `~=` means `!=` in another language I like, so `~<` looks like _not less than_ to me, there is ofc no problem in practice, once one have learned. Maybe `@<`, _@lphabetically less_, it looks a bit like a blob though 😄 .

***

Cheers.

HelgeffegleH:
Hello :wave:.

No I do not suggest that. Disregarding my initial associations, the fact that `~` is part of the regex short hand doesn't necessarily make it more  suitable as a prefix for the string operators. Since `~` means bitwise-not, the negation association might be natural. But as I said, it is not a problem once learnt, but I wonder, is the `~` an important part of your suggestion? I doubt it is.

Btw, I do not value the meaning of `~=` in ahk at all. 

Cheers.

jeeswg:
My instinct is to allow the operators containing < or > [< <= > >= (and a restored <>)] to do string comparison based on A_StringCaseSense.
We could have a NumCompare function to force numeric comparison.
That function could have an Options parameter, e.g. 'V' for version/IP comparison.
Similar to: [How to check if an IPv6 address belongs to a network?](https://autohotkey.com/boards/viewtopic.php?f=5&t=63883&p=273845#p273845)
Or we could have a separate VerCompare/VersionCompare/IPCompare function.

PR #110, opened 2018-08-21
Adding help function: TokensToScriptObject and two related macros.

HelgeffegleH:
Hello :wave:

The help function returns an object with key-value pairs according to the first parameter.

Example usage:

```c++
ExprTokenType params[] = 
{
	_T("Key1"), _T("Value1"),
	_T("Key2"), _T("Value2")
};
Object* obj = TokensToScriptObject(params, 4);
```

The macros can be used in BIFs and methods and are easy to find in connection to the other useful and common `_f_xxx` and `_o_xxx` macros. Example, equivalent to the above (plus error check)

```c++
Object* obj;
_f_create_script_object(obj, _T("Key1"), _T("Value1"), _T("Key2"), _T("Value2"));
```

Objects are currently not very common as return values, however, future functions/methods might return objects, this addition will make this easier.

Reason: Facilitate the use of objects as return values. Reduce redundancy and improves code clarity. Making the source code easier for the uninitiated to work with. 

Perhaps it is odd to add something which isn't actually being used, however, I do have use cases in mind and I think many others can find it useful too.

Cheers.

### DIFF
--- a/source/script.h
+++ b/source/script.h
@@ -400,6 +400,8 @@ struct ArgStruct
 	ExprTokenType *postfix;  // An array of tokens in postfix order.
 };
 
+Object* TokensToScriptObject(ExprTokenType aTokens[], int aCount);
+
 #define BIF_DECL_PARAMS ResultToken &aResultToken, ExprTokenType *aParam[], int aParamCount
 
 // The following macro is used for definitions and declarations of built-in functions:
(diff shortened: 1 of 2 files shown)


nnnik:
Well most functions that return an object return a custom object.
It might be better to return a normal AHK Object for some.

PR #113, opened 2018-09-01
Added .vs folder to .gitignore

nnnik:


Lexikos:
This is redundant. The next merge with master (and #112) will cover it.

PR #114, opened 2018-09-01
Added A_AhkMajorVersion for v2

nnnik:
The built in variable A_AhkMajorVersion was added.
It tries to find the major part of T_AHK_VERSION.
It does that by finding the second . and returning the string before that.
If it cannot find a second dot it will look for a - instead of the second dot (the first dot needs to be found regardless).
If all fails it just returns the first 3 characters.

This code should also work for v1

### DIFF
--- a/source/script.cpp
+++ b/source/script.cpp
@@ -304,6 +304,7 @@ VarEntry g_BIV[] =
 // name to reduce code size and speed up the comparisons.
 VarEntry g_BIV_A[] =
 {
+	A_(AhkMajorVersion),
 	A_(AhkPath),
 	A_(AhkVersion),
 	A_w(AllowMainWindow),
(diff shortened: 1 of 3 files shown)


HelgeffegleH:
You could `#define AHK_MAJOR_VERSION`.

Cheers.

nnnik:
I had the same idea but decided against it since that would increase the maintenance needed when updating.

Lexikos:
I'm currently not in favour of adding this variable. I expect that writing scripts that work on both v1.x and v2.0.0 will be rather difficult, if not impossible.

If it is added, its value will not be computed at runtime. The added code is a waste of space. Splitting the version number constant into major and minor will not increase the maintenance cost; it would just require updating [the release script](https://github.com/Lexikos/AutoHotkey-Release/blob/master/prepare-ver.ahk) as well. I do not change `ahkversion.h` by hand.

PR #115, opened 2018-09-01
Added A_AhkMajorVersion

nnnik:
Added the built in variable A_AhkMajorVersion for v1.

it behaves exactly like the v2 counterpart described in  #114

### DIFF
--- a/source/script.cpp
+++ b/source/script.cpp
@@ -56,6 +56,7 @@ VarEntry g_BIV[] =
 // name to reduce code size and speed up the comparisons.
 VarEntry g_BIV_A[] =
 {
+	A_(AhkMajorVersion),
 	A_(AhkPath),
 	A_(AhkVersion),
 	A_x(AppData, BIV_SpecialFolderPath),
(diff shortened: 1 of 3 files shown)


jeeswg:
The variable returns 1.1 or 2.0 (major.minor) not 1 or 2 (major).

Instead of 'Major', I might suggest one of: Main, Core, Float, Value, Num.

I would propose: A_AhkVersionMain or A_AhkMainVersion e.g. 1.1 or 2.0.
And similarly: A_OsVersionMain or A_OsMainVersion e.g. 6.1 (Windows 7).

[Discussion](https://autohotkey.com/boards/viewtopic.php?f=13&t=53924).

nnnik:
I chose major because it implies the presence of another variable named minor (which I planned to add in another PR when this one got accepted) and is related to versioning (although it doesnt fit perfectly).
I don't feel particulary strong about the name at all - however these 2 points stand.
I don't dislike the idea with main, however I feel that we are just splitting hairs.

Lexikos:
Closing; see #114.

My preferred workflow for features that will exist in both branches is to implement them in v1, then merge into v2 and update. Having a pull request duplicated for the second branch is a distraction that I'd prefer to not have, especially for a proposed feature that I haven't agreed to. It also causes split discussions.

PR #116, opened 2018-09-21
Changing operators // and //=  to only do integer division.

HelgeffegleH:
Hello 👋 .

Float operands are truncated to integers before the division.

Reason, simplifies behaviour and documentation, floor divide is already possible with (more readable) `floor(x/y)`, disregarding it produces an integer. Integer division is not easily achieved in any other way, and is important to keep for accurate division of integers which cannot be represented as floats.

### About the implementation

As mentioned in code comment, some redundancy is added to avoid extra an extra check for the preceding branch. It might not be very significant, but _every little helps_. (I hope _that_ makes sense, I translated)

Edit, I will fix the redundancy issue.

### DIFF
--- a/source/defines.h
+++ b/source/defines.h
@@ -170,7 +170,7 @@ enum SymbolType // For use with ExpandExpression() and IsNumeric().
 #define SYM_CPAREN_FOR_OPAREN(symbol) ((symbol) - (SYM_OPAREN - SYM_CPAREN)) // Caller must confirm it is OPAREN or OBRACKET.
 #define SYM_OPAREN_FOR_CPAREN(symbol) ((symbol) + (SYM_OPAREN - SYM_CPAREN)) // Caller must confirm it is CPAREN or CBRACKET.
 #define YIELDS_AN_OPERAND(symbol) ((symbol) < SYM_OPAREN) // CPAREN also covers the tail end of a function call.  Post-inc/dec yields an operand for things like Var++ + 2.  Definitely needs the parentheses around symbol.
-	, SYM_ASSIGN, SYM_ASSIGN_ADD, SYM_ASSIGN_SUBTRACT, SYM_ASSIGN_MULTIPLY, SYM_ASSIGN_DIVIDE, SYM_ASSIGN_FLOORDIVIDE
+	, SYM_ASSIGN, SYM_ASSIGN_ADD, SYM_ASSIGN_SUBTRACT, SYM_ASSIGN_MULTIPLY, SYM_ASSIGN_DIVIDE, SYM_ASSIGN_INTEGERDIVIDE
 	, SYM_ASSIGN_BITOR, SYM_ASSIGN_BITXOR, SYM_ASSIGN_BITAND, SYM_ASSIGN_BITSHIFTLEFT, SYM_ASSIGN_BITSHIFTRIGHT
 	, SYM_ASSIGN_CONCAT // THIS MUST BE KEPT AS THE LAST (AND SYM_ASSIGN THE FIRST) BECAUSE THEY'RE USED IN A RANGE-CHECK.
 #define IS_ASSIGNMENT_EXCEPT_POST_AND_PRE(symbol) (symbol <= SYM_ASSIGN_CONCAT && symbol >= SYM_ASSIGN) // Check upper bound first for short-circuit performance.
@@ -192,7 +192,7 @@ enum SymbolType // For use with ExpandExpression() and IsNumeric().
 	, SYM_BITSHIFTLEFT, SYM_BITSHIFTRIGHT // << >>  ALSO: SYM_BITSHIFTRIGHT MUST BE KEPT LAST AMONG THE BIT OPERATORS BECAUSE IT'S USED IN A RANGE-CHECK.
 #define IS_BIT_OPERATOR(symbol) ((symbol) <= SYM_BITSHIFTRIGHT && (symbol) >= SYM_BITOR) // Check upper bound first for short-circuit performance (because operators like +-*/ are much more frequently used).
 	, SYM_ADD, SYM_SUBTRACT
-	, SYM_MULTIPLY, SYM_DIVIDE, SYM_FLOORDIVIDE
+	, SYM_MULTIPLY, SYM_DIVIDE, SYM_INTEGERDIVIDE
 	, SYM_POWER
 	, SYM_LOWNOT  // LOWNOT is the word "not", the low precedence counterpart of !
 	, SYM_NEGATIVE, SYM_POSITIVE, SYM_HIGHNOT, SYM_BITNOT, SYM_ADDRESS  // Don't change position or order of these because Infix-to-postfix converter's special handling for SYM_POWER relies on them being adjacent to each other.
--- a/source/defines.h
+++ b/source/defines.h
@@ -188,13 +188,13 @@ enum SymbolType // For use with ExpandExpression() and IsNumeric().
 	, SYM_LOW_CONCAT // Zero-precedence concat, used so that "x%y=z%" is equivalent to "x%(y=z)%".
 	
 	// INTEGER OPERATORS START
-	// Integer operators truncates float operands to integers.
+	// Note that the below macro does not consider SYM_BITNOT or the bit-assignment operators.
 	, SYM_BITOR // Seems more intuitive to have these higher in prec. than the above, unlike C and Perl, but like Python.
 	, SYM_BITXOR // SYM_BITOR (ABOVE) MUST BE KEPT FIRST AMONG THE INTEGER OPERATORS BECAUSE IT'S USED IN A RANGE-CHECK (see IS_INTEGER_OPERATOR).
 	, SYM_BITAND
 	, SYM_BITSHIFTLEFT, SYM_BITSHIFTRIGHT // << >>  
 	, SYM_INTEGERDIVIDE	// eg, x // y ALSO : SYM_INTEGERDIVIDE MUST BE KEPT LAST AMONG THE INTEGER OPERATORS BECAUSE IT'S USED IN A RANGE-CHECK (see IS_INTEGER_OPERATOR).
-#define IS_INTEGER_OPERATOR(symbol) ((symbol) <= SYM_INTEGERDIVIDE && (symbol) >= SYM_BITOR) // Check upper bound first for short-circuit performance (because operators like +-*/ are much more frequently used).
+#define IS_INTEGER_OPERATOR(symbol) ((symbol) <= SYM_INTEGERDIVIDE && (symbol) >= SYM_BITOR) // Currently not considered: SYM_BITNOT and the bit-assignment operators.
 	// INTEGER OPERATORS END
 
 	, SYM_ADD, SYM_SUBTRACT
--- a/source/defines.h
+++ b/source/defines.h
@@ -186,13 +186,19 @@ enum SymbolType // For use with ExpandExpression() and IsNumeric().
 	, SYM_REGEXMATCH // ~=, equivalent to a RegExMatch call in two-parameter mode.
 	, SYM_CONCAT
 	, SYM_LOW_CONCAT // Zero-precedence concat, used so that "x%y=z%" is equivalent to "x%(y=z)%".
+	
+	// INTEGER OPERATORS START
+	// Integer operators truncates float operands to integers.
 	, SYM_BITOR // Seems more intuitive to have these higher in prec. than the above, unlike C and Perl, but like Python.
-	, SYM_BITXOR // SYM_BITOR (ABOVE) MUST BE KEPT FIRST AMONG THE BIT OPERATORS BECAUSE IT'S USED IN A RANGE-CHECK.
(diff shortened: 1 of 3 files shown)


Lexikos:
I do not think it is useful or intuitive for `a//b` to truncate `a` and `b` prior to the division. For instance, `3//1.45` will return 3 instead of 2.0 (`n//1.45` will always return `Integer(n)`).

As I see it, the problem with `//` is that it handles integers and the equivalent floating-point numbers inconsistently (effectively rounding toward zero for integers and rounding to the left for floating-point numbers). So in other words, the problem is merely that the method of rounding is inconsistent. The obvious solution is to pick one method and use it for both cases.

If we round toward zero, the behaviour for integers doesn't change, so we can still use "high-performance" integer division behind the scenes when both inputs are integers. `a//b` would complement `Mod(a,b)`; i.e. `(a//b) * b + Mod(a,b) = a` should be true (it already is when `a` and `b` have the same sign or are both integers).

I suppose that this would also "simplify" the behaviour and documentation, but I do not see "complexity" as the problem, but specifically this inconsistency.

HelgeffegleH:
Hello :wave:.
> I do not think it is useful or intuitive for a//b to truncate a and b prior to the division

I didn't mean for it to be useful, I was adhering to the behaviour of the other integer opreators, that is the bitwise operators.  I intended to make the separate suggestion to throw an exception when passing floats to integer operators (bitwise and `//`).

The suggestion is to make `//` an integer-only operator because that is the only useful thing it does now. I do not see why we need an operator doing (eg) `floor(x/y)`, it is a waste of code and documentation, as far as I can tell. 

> So in other words, the problem is merely that the method of rounding is inconsistent.

If someone needs to round true division towards zero, it is trivial to write a function for it, eg,
```autohotkey
fix(z){
	return z < 0 ? ceil(z) : floor(z)
}
```

___

> `(a//b) * b + Mod(a,b) = a` should be true

This is the documented relatation for c++ integer operators `/` and `%`, see [Multiplicative Operators and the Modulus Operator
](https://msdn.microsoft.com/en-us/library/ty2ax9z9.aspx)Looking at [fmod](https://msdn.microsoft.com/en-us/library/20dckbeh.aspx) we see that what the corresponding relation is not stated to be  `floor(x/y) * y + fmod(x,y)`, instead it states,

> The fmod function calculates the floating-point remainder `f` of `x / y` such that `x = i * y + f`, where `i` is an integer

Specifically, `i` is not mentioned to be `floor(x/y)`, and it is not `fix(x/y)` either, example,
```autohotkey
a := 5000000000000250000 ; or 5000000000000249856.0
b := 5000.0
msgbox a//b * b + Mod(a,b) = a ; 0
```

The above relation is easy to solve for `i`, so I guess this implementation would _work_, disregarding divison by zero,
```
case SYM_FLOORDIVIDE:	this_token.value_double = (left_double - fmod(left_double, right_double)) / right_double; break;
```
This is also trivial to implement in script if really needed.

I do not know about `qmath`.

Cheers.

Lexikos:
Throwing an exception would be preferable to silently truncating the input values (but not preferable overall).

For the bitwise operators, truncating the input is IMO the only sensible course (other than throwing an exception). That isn't so for integer division, which can meaningfully be applied to any numeric type.

You indicate that `//` is only useful with pure integers, so I guess you're saying that it is only useful because it keeps the 64-bit integer precision and may be faster than normal division. In that case, I would argue that it is not useful. I have reviewed every use of `//` in my script collection and found none where the precision or performance difference matter. There are also many instances where `/` could have been used, since the result would be truncated anyway (e.g. by Loop or VarSetCapacity). If we really need an `Integer` result, we can use `Integer(a/b)`.

How is it a waste of documentation? There is none yet; only documentation for the similar-looking floor divide operator. The new operator can be described as follows (or similar):

> *(If it returns Integer or Float; same wording as Multiply)*
> **Integer divide (//):** Performs division and returns the quotient minus the remainder. The result is an integer if both inputs are integers; otherwise, it is a floating point number.

> *(If it returns Integer, like `Integer(a/b)`)*
> **Integer divide (//):** Performs division and returns the quotient minus the remainder. The result is always of type Integer.

It is wasteful to have an operator for integer division and then require something more verbose to perform the same operation just because the inputs were floating-point or *could be* floating-point.

HelgeffegleH:
> You indicate that `//` is only useful with pure integers, so I guess you're saying that it is only useful because it keeps the 64-bit integer precision and may be faster than normal division.

No need to guess, I mentioned it in my first comment,

>  Integer division is not easily achieved in any other way, and is important to keep for accurate division of integers which cannot be represented as floats.

This is the only reason `//` exist in the first place of course, it is needed to complement `/`, which only does float divison.

___

> I have reviewed every use of `//` in my script collection and found none where the precision or performance difference matter. __[...]__   If we really need an Integer result, we can use `Integer(a/b)`.

You get an integer indeed, but it wouldn't be the correct one in general. I'm not sure what your point is exactly. I think the basic math operations should be fully supported for both numerical types, that is `+`, `-`, `*` and `/`. Now, since `/` is chosen to do float division, one adds `//` to do integer division. 

Reserving `//` for integers has the benefit of predictability and catching mistakes. Having it fulfill

```autohotkey
(a//b) * b + Mod(a,b) = a
```

for both floats and integers might be a _pleasant_ symmetry. But when did you (or anyone) need that float result? It is basically just a less accurate float division, with no apparent use.

___

> For the bitwise operators, truncating the input is IMO the only sensible course (other than throwing an exception).

At least for `<<`, `>>` and `~`, there are sensible corresponding float operations, implemented in script,

```autohotkey
float_shift_left(f, e) {
	return f*2.0**e
}
float_shift_right(f, e) {
	return float_shift_left(f, -e)
}
float_not(f){
	return -f-1.0
}
```
Implementing shifting like this have the _pleasing_ result of causing (eg) `1.0 << 64 == 18446744073709551616.0` instead of the (perhaps?) less _pleasing_ `1`. Imo, handling floats like this for `<<`, `>>` and `~` is on the same level of deviation from the simple basic meaning of these integer operations, which are the primary reason these operators exist, as handling `//` for floats as,

```autohotkey
float_operation(a, b){
	return (a - mod(a, b)) / b
}
```

As far as I am aware, `float_operation` isn't a _thing_, at least `float_shift_xxx` has a corresponding CRT function, [ldexp](https://msdn.microsoft.com/en-us/library/zx52ds7f.aspx) (although only `int` exponent).

___

> __Integer divide__ (//): Performs division and returns the quotient minus the remainder. The result is an integer if both inputs are integers; otherwise, it is a floating point number.

Would you really call that _Integer divide_? Also, saying, _Performs division and returns the quotient minus the remainder_, might be misleading in the context of integer division, which results in two parts, __the quotient__ and __the remainder__, the remainder is discarded and the quotient is returned. I do not know what to call the _float operation_.  I suggest,

> __Integer divide__ (//): Performs integer division and returns the quotient, discarding the remainder. Any non-integer operands causes an exception to be thrown.

___

> It is wasteful to have an operator for integer division and then require something more verbose to perform the same operation just because the inputs were floating-point or could be floating-point.

It is wasteful to have an integer operator perform a float operation, and then require something more verbose to detect mistakes when one wants the integer operation, just because the inputs can be floating-point.

Cheers.

jeeswg:
I'd recommend floor divide *and* integer divide operators, e.g. // and idiv, for easy/reliable formula translation.

// is floor divide in Python, so I'd make // do floor divide.

Based on searches, // as integer divide is unheard of, and would break the principle of least astonishment.

[Discussion](https://autohotkey.com/boards/viewtopic.php?f=37&t=56113).

HelgeffegleH:
Since it has been agreed that

> Throwing an exception would be preferable to silently truncating the input values 

I have implemented this behaviour for all integer operators, i.e., `//` and the bitwise operators. The `IS_INTEGER_OPERATOR` is kept to help correctly identify type mismatch and help catch bugs.

The most _reasonable_ alternative to this would be to implement the `float_operation` and something like the `float_shift_xxx` and `float_not` for float operands. In such case I think they should always produce an integer (float without fractional part), and you should only allow floats without any fractional part, cf. `**` when raise negative base. Then the integer operators can _work_ for floats in a _similar_ way, although I do not see any real value in this. It only makes it harder to catch mistakes, and especially since mixed type operands can occur it would be a poor choice.

__Bonus__, we now do not have to use the `IS_BIT_OPERATOR` / `IS_INTEGER_OPERATOR` macros for every float operation 👍 .

Cheers.

Lexikos:
> No need to guess, I mentioned it in my first comment,

There is often need to "guess", because we do not communicate perfectly, nor think the same way. What I wrote has additional meaning that you did not mention (at least in what you quoted): i.e. it is useful for no other reasons, and it is faster.

> It is basically just a less accurate float division, with no apparent use.

Would you say the same about floor division? It is a different operation, with a different purpose and result. If the operator gives exactly the desired result, it cannot be considered *less accurate*. I think my point of view could be summarized like this: the focus should be on the math, not the machine. The difference between 1 and 1.0 is 0.0.

> This is the only reason // exist in the first place of course

Let's just say that this is *your* interpretation of the purpose of `//`. Mine is different, *of course*, hence my prior disagreement.

I would point out that the only reason given in the source code comments is that "performance is much higher for integer vs. float division" (although it also alludes to other unmentioned reasons), and the documentation focuses on this while completely omitting any concerns about the higher precision of integers for certain ranges of numbers.

> You get an integer indeed, but it wouldn't be the correct one in general. 

Maybe I'm missing something significant, but it seems to me that it would be the correct one *in general*. Scripts *in general* do not require the full 64-bit range. There may be some edge cases where the precision matters, not relevant to my own use of the operator (as I mentioned). However, even if I am unable to come up with an example of where the precision is important, I can easily assume that one exists. I at least know that there are areas of math that are completely outside my range of experience.

> Reserving // for integers has the benefit of predictability  [...]

I disagree that it is *more predictable* than giving the same mathematical result regardless of type (within the limits of precision).

> At least for <<, >> and ~, there are sensible corresponding float operations, 

Only if they were described in terms of arithmetic rather than in terms of bits (as *bitwise* operators).

> Would you really call that *Integer divide*? Also, [the description you suggest] might be misleading [...]

I did not put much thought into the name or wording. My point was about the alleged "waste of documentation"; your suggested documentation serves my point just as well or better.

> It is wasteful to have an integer operator perform a float operation, and then require something more verbose to detect mistakes when one wants the integer operation, [...]

From my perspective, no such requirement exists. It is not inherently a mistake to pass a floating-point number to an arithmetic operator.

> But when did you (or anyone) need that float result? 

When did I need `//` at all? I already implied that I don't. As for anyone else, I'm not going to question what they use *math* for.


One point in favour of "throw for floats" is that more restrictive behaviour can be relaxed later, if it turns out to be too inconvenient.

HelgeffegleH:
> One point in favour of "throw for floats" is that more restrictive behaviour can be relaxed later, if it turns out to be too inconvenient.

I started writing a longer response, but actually, __that is perfect__, we do not need to agree on anything else, I'd be happy to end this discussion with that 👍.

Cheers.

jeeswg:
I wouldn't normally write on GitHub, but this PR really concerns me.
For consistency, // should always be floor divide or integer divide.
I don't see why you would choose integer divide:
(a) it's already called 'floor divide' in the documentation, suggesting Chris's intentions, he probably based it off Python (Python is mentioned re. case SYM_FLOORDIVIDE in script_expression.cpp)
(b) in Python, // is used for **floor** divide, so due to the resulting confusion that would occur, // as integer divide is a really bad idea, Chris chose operators in accordance with other languages (e.g. C++ and Python), he didn't do something weird for the sake of it (in C++/Java/C#/JavaScript/PHP, // is used for comments)
(c) just pick another operator symbol for integer division, have both, cf. AutoIt, Chris added numerous mathematical operators for convenience, add wanted both floor and integer division to be available (they're especially useful for using complex formulas across multiple languages)

A summary of proposals:
`a // b --> Int(Int(a) / Int(b))` [what is this!?]
`a // b --> Int(a / b)` [fails if either a or b is a float][why not handle floats!?][at least handle floats equal to integers]
`a // b --> Int(a / b)` [accepts floats or integers][why int. div. instead of floor div.!?][// is commonly used as **floor** divide]
`a // b --> Floor(a / b)` [accepts floats or integers][my proposal][consistent with Python]
Across PRs, you appear to desire to make operators stricter, but these changes aren't really protecting/helping the user, and are greatly reducing convenience, and go against what I interpret as Chris's vision of convenient operators.
You've suggested removing <>, and removing string comparison via operators, and removing float support for //. All quite undesirable in my view.

Formulas were mentioned. AFAIK (and I did some tests):
`Floor(a/b) * b + FloorMod(a,b) = a`
`Int(a/b) * b + Mod(a,b) = a`
I would add FloorMod to AHK, it's very handy, it always gives a positive/zero result.

Compare operators in AutoHotkey and AutoIt. Chris added many operators.
[AutoIt](https://www.autoitscript.com/autoit3/docs/intro/lang_operators.htm).
[AutoHotkey](https://autohotkey.com/docs/Variables.htm#Operators).

My proposals re. division, modulus and mathematical convenience:
New functions: Between (like AHK_H), FloorMod (cf. Java and Python's %), Pow (cf. AHK v1's Transform), Sign (cf. Excel).
New operators: restore <> (cf. AutoIt, SQL), // consistent as floor divide (cf. Python), idiv (or something else for integer division), mod (mod operator, cf. Excel VBA).

PR #117, opened 2018-09-23
Adding option validation for UserMenu::UpdateOptions.

HelgeffegleH:
Hello 👋 .

Throws `ERR_INVALID_OPTION` on invalid option.

Reason, facilitates finding common mistakes.

Invalid priority options are not detected, due to rarity and for brevity. Hence, eg `Pxyz`, is equivalent to `P0`.

### DIFF
--- a/source/script_menu.cpp
+++ b/source/script_menu.cpp
@@ -497,7 +497,7 @@ ResultType UserMenu::AddItem(LPTSTR aName, UINT aMenuID, IObject *aCallback, Use
 		name_dynamic = Var::sEmptyString; // So that it can be detected as a non-allocated empty string.
 	UserMenuItem *menu_item = new UserMenuItem(name_dynamic, length + 1, aMenuID, aCallback, aSubmenu, this);
 	if ( !menu_item // Should also be very rare.
-		 || (*aOptions && !UpdateOptions(menu_item, aOptions)) ) // update options
+		 || (*aOptions && UpdateOptions(menu_item, aOptions) != OK) ) // update options
 	{
 		if (name_dynamic != Var::sEmptyString)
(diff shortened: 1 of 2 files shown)


Lexikos:
The convention used in the majority of current functions which return either FAIL or OK is to evaluate as boolean. I do not feel that `x != OK` or `x == FAIL` is any clearer than `!x` in these cases.

> // FAIL = 0 to remind that FAIL should have the value zero instead of something arbitrary
// because some callers may simply evaluate the return result as true or false
// (and false is a failure):

source/script_menu.cpp
```diff
@@ -496,11 +496,16 @@ ResultType UserMenu::AddItem(LPTSTR aName, UINT aMenuID, IObject *aCallback, Use
...
 	UserMenuItem *menu_item = new UserMenuItem(name_dynamic, length + 1, aMenuID, aCallback, aSubmenu, this);
-	if (!menu_item) // Should also be very rare.
+	if ( !menu_item // Should also be very rare.
+		 || (*aOptions && !UpdateOptions(menu_item, aOptions)) ) // update options
 	{
 		if (name_dynamic != Var::sEmptyString)
 			free(name_dynamic);
-		return g_script.ScriptError(ERR_OUTOFMEM);
+		if (menu_item)
+			delete menu_item;
```
Lexikos: Calling the UserMenuItem destructor (via delete) after calling `free(name_dynamic)` is not safe. We must `free(name_dynamic)` when and only when `!menu_item`. Commit 7a7e17af fixes this.

HelgeffegleH: I see my mistake, thanks.

PR #118, opened 2018-11-04
Changing ** (power) operator to throw exception when both operands are 0

HelgeffegleH:
Reason, it is mathematically undefined, an exception is consistent with `x/0` (divide by zero). Secondary, consistent behaviour between `32` and `64` builds (currently, `0**0` is `0` on `32` build and `1`  on `64` build.)

Example,

```autohotkey
(0**0) ; Error
```

It is not uncommon to let `0**0` be `1` in computing, and some times also in mathematics. That would be an acceptable alternative.

Cheers.

### DIFF
--- a/source/script_expression.cpp
+++ b/source/script_expression.cpp
@@ -1142,6 +1142,10 @@ LPTSTR Line::ExpandExpression(int aArgIndex, ResultType &aResult, ResultToken *a
 						// Throw an exception rather than returning something undefined:
 						goto divide_by_zero;
 					}
+					else if (left_int64 == 0 && right_int64 == 0)	// 0**0, not defined.
+					{
+						goto abort_with_exception;
+					}
 					else // We have a valid base and exponent and both are integers, so the calculation will always have a defined result.
 					{
 						if (left_was_negative = (left_int64 < 0))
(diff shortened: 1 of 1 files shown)


Ursi:
I have a pretty strong mathematics background. I would argue that the most mathematically sound way to handle it would be to have 0^0 = 1 if the 0's are both integers, and undefined if at least one of them is a float. This is because in the integers, you can very logically define exponentiation such that any number to the 0 is 1, and a lot of things in discrete math have formulas that work for all cases as long as 0^0 = 1, otherwise, they would work for every case but 0. Now as far as calculus is concerned 0^0 is undefined, and calculus works in the real numbers, which are approximated by floats. I'm assuming 0^0 isn't going to be implemented in this way (but I do think it's the best), so other than this, my vote is 0^0 = 1, because then at least you won't make an exception when you want 0^0 to equal 1, which is not uncommon.

HelgeffegleH:
Hello __Ursi__, I have a pretty strong  potato peeling background 👋 .

Originally, I intended to suggest `0^0 = 1`, but I'm not interested enough to try to make an argument for it, there are plenty of arguments. However, the fact that the behaviour of `0^0` has probably been undefined in AHK forever, makes the exception the least controversial suggestion.  So there is no need to discuss it, it should be an exception if it is going to continue to be undefined. If some one wants to define it, I'm not going to oppose any sensible suggestions. I know sometimes there is different behaviour for floats and integers, I think that is inappropriate for AHK, mainly since AHK is a loosely typed language.

Cheers ☕️ .

HelgeffegleH:
Hello :wave: .

>  At least one user had not even considered that the result could be anything but 1. 

I'm sure at least one user would not even considered that the result could be anything but `0`. `0^0` looks much more like `0`  than `1`, I do not expect most AHK users to be familiar with the common arguments for `0^0 = 1`. Beside, the arguments are mostly detached from anything that would naturally arise in AHK-programming, that is my guess at least. 

My personal reason for prefering `0^0 = 1` is probably only because I'm more used to it, not that I really need it to be. 

Cheers.

PR #119, opened 2018-11-04
Changing operator ** to not cast integer operands to float...

HelgeffegleH:
...and then cast result back to integer for exponents `>= 0`.

__Reasons,__ allow correct results for all `x**y` (`x,y` integer, `y>0`) which is within the range of `__int64`.

Example.

```autohotkey
msgbox 1234**6 ; yields correct result in this branch
```

### Overflow.
Overflow is detected and causes the result to be `_I64_MIN`,  which is useful because it enables, eg, `2**63-1 == 9223372036854775807` which is the correct result. While other operators do not handle overflow, the `**` operator should beacuse it causes overflow even for quite small operands, eg `100**100` is _huge_.

I have implemented the behaviour in this branch in a much simpler way, less complex, less code, it perfroms slightly slower. If there is an interest I'll provide the alternative version.

Not detecting overflow simplifies the implementation, and reduces code size.

__Note:__ Introduces a different inconsistency, which needs separate handling, that is, `0**0.0` is `0.0` on `32` build while `1.0` on `64`, `0**0` is `1` on both builds. Currently, both `0**0` and `0.0**0` is `0` on `32` and `1` on `64` build. See #118.

Cheers.

### DIFF
--- a/source/script.h
+++ b/source/script.h
@@ -400,6 +400,9 @@ struct ArgStruct
 	ExprTokenType *postfix;  // An array of tokens in postfix order.
 };
 
+__int64 pow_ll(__int64 base, __int64 exp);
+bool mul_ll(unsigned __int64 x, unsigned __int64 y, unsigned __int64* res);
+
 #define BIF_DECL_PARAMS ResultToken &aResultToken, ExprTokenType *aParam[], int aParamCount
 
 // The following macro is used for definitions and declarations of built-in functions:
--- a/source/script2.cpp
+++ b/source/script2.cpp
@@ -17524,3 +17524,71 @@ DWORD GetProcessName(DWORD aProcessID, LPTSTR aBuf, DWORD aBufSize, bool aGetNam
 	CloseHandle(hproc);
 	return buf_length;
 }
+
+__int64 pow_ll(__int64 base, __int64 exp)
+{
+	/*
+	Caller must ensure exp >= 0
+	Below uses 'a^b' to denote 'raising a to the power of b'.
+	Computes and returns base^exp, returns _I64_MIN on overflow. By convention, x^0 returns 1, even when x == 0,
+	caller should ensure base is non-zero when exp is zero to handle 0^0.
+	*/
+	if (exp == 0)
+		return 1ll;
+	 
+	// For simplicity, use positive numbers below. 
+	BOOL base_is_negative = (base < 0);
+	BOOL result_is_negative = (base_is_negative && exp % 2);
+
+	base = base_is_negative ? -base : base;
+
+	// based on: https://en.wikipedia.org/wiki/Exponentiation_by_squaring (2018-11-03)
+	unsigned __int64 result = 1;
+	while (exp > 1)
+	{
+		if ((exp % 2) // exp is odd
+			&& !mul_ll(result, base, &result))
+			return _I64_MIN;
+		exp /= 2;
+		if (base >= 3037000500)
+			// Compairison is made to avoid calling mul_ll.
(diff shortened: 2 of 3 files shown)


Lexikos:
If the result cannot be represented as an integer but can be approximated as float, this might cause an unacceptable loss of flexibility.

Lexikos:
On second thought, one can explicitly use floating-point operands. Still, it would be more flexible and intuitive to revert to floating-point in the case of overflow.

Lexikos:
On third thought, it would seem incongruous to do so for `**` but no other operator.

I don't think I care enough about this right now to think it through or decide to merge it.

HelgeffegleH:
> it would be more flexible and intuitive to revert to floating-point in the case of overflow.

I think that is OK. However, if overflow should invoke some other behaviour than the the proposed one, note that `mul_ll` will incorrectly deem `(-2)**63` as overflowing, it doesn't matter for the current proposed behaviour since the result is `_I64_MIN` anyways. I might fix it.

> I don't think I care enough about this right now to think it through or decide to merge it.

I can't argue against that 😉 

Cheers.

Lexikos:
> Overflow is detected and causes the result to be _I64_MIN, which is useful because it enables, eg, 2**63-1 == 9223372036854775807 which is the correct result.

I think that there should be a better reason than this. Do you mean "this is useful because it enables 2**63-1 == 9223372036854775807", or are there more reasons (as implied by "eg")? There are other ways to achieve this result, or eliminate the need for it.

I think there are basically four ways to handle overflow:
  - Wrap.
  - Saturate (return the minimum or maximum possible value).
  - Throw an exception.
  - Leave it as undefined behaviour (but maybe this doesn't count as *handling* it).

You may know that `Integer(2.0**63)` already returns `_I64_MIN`. As I mentioned in #101, this is [actually undefined behaviour](http://blog.frama-c.com/index.php?post/2013/10/09/Overflow-float-integer). The article has a pretty good explanation (at least if you already understand the terminology):

> Generally speaking, `cvttsd2si` always produces `-0x80000000` in cases of overflow. That is almost like saturation, except that floating-point numbers that are too positive are wrapped to `INT_MIN`. One may think of it as saturating to either `-0x80000000` or `0x80000000`, and in the latter case, wrapping around to `-0x80000000` because of two's complement. 

`2**63` would also produce `_I64_MIN` if `**` implements wrapping overflow, but larger values would produce the low 64 bits of the actual result.

I think `2**63-1` is probably only written as shorthand for the maximum int64 value.  In that case, a better solution would be to provide named constants which contain the appropriate min/max values.

> While other operators do not handle overflow, the ** operator should beacuse it causes overflow even for quite small operands, eg 100**100 is huge.

Does this make it *more likely* to overflow? I would expect script authors to be passing values suitable for this operator. Anyway, `n**2` or `n*n` can also easily overflow if n is near `_I64_MAX`.

Even if it is more likely to overflow, I do not see how returning `_I64_MIN` will be more useful than the alternatives.

HelgeffegleH:
I forgot this.

> Do you mean "this is useful because it enables 2**63-1 == 9223372036854775807", or are there more reasons (as implied by "eg")?

I only meant `2**63-1` to be an example of `2**63-x`.

> There are other ways to achieve this result, or eliminate the need for it.

I agree.

> Does this make it more likely to overflow?

I was thinking that a relatively small mistake, such as one excessive loop iteration, would yield a larger amplitude of the error and hence more likely to overflow. The chance of making the mistake in the first place, i.e., the excessive loop iteration would not be greater I guess.

> I would expect script authors to be passing values suitable for this operator.

Thanks for the confidence, I don't deserve it though 😉 

Cheers.

HelgeffegleH:
Closing due to #165.

PR #121, opened 2018-11-04
Disabling unquoted argument and return types for BIF_DllCall.

HelgeffegleH:
Reasons, makes syntax more consistent (no other functions allow unquoted text as input), consequently makes scripts and documentation easier to read / understand. Removes some potential bugs, especially since the introduction of nested functions / closures, specifically, demaning quotes reduces the risk that nested functions becomes closures due to having `dllcall`s with same argument types,

```autothotkey
f(){
	return type(() => dllcall(0, ptr))
	dllcall 0, ptr
}
msgbox f()	; Closure, which ofc is correct, but probably not intended.
```
Also, in this branch, using `#warn LocalSameAsGlobal` will cause warnings if there exists a global variable named `ptr` in the above script.

Simplifies implementation, reduces code size. 

### About the implementation.
 `case SYM_STRING:` was put before `case SYM_VAR:` in two switches, due to the assumption that argument types are more often specified by something like `'str'` rather than `var := 'str'`. Added call to `Var::MaybeWarnUninitialized()` if appropriate before throwing `ERR_INVALID_RETURN_TYPE`, since similar is done for invalid arg types. Obsolete comments removed (w.r.t the changes), some might remain. Function `bool IsDllArgTypeName(LPTSTR name)`  removed since not called anymore.

Cheers.

### DIFF
--- a/source/script.cpp
+++ b/source/script.cpp
@@ -13337,15 +13337,6 @@ void Script::MaybeWarnLocalSameAsGlobal(Func &func, Var &var)
 {
 	if (!g_Warn_LocalSameAsGlobal)
 		return;
-
-#ifdef ENABLE_DLLCALL
-	if (IsDllArgTypeName(var.mName))
-		// Exclude unquoted DllCall arg type names.  Although variable names like "str" and "ptr"
-		// might be used for other purposes, it seems far more likely that both this var and its
-		// global counterpart (if it exists) are blank vars which were used as DllCall arg types.
-		return;
-#endif
-
 	Line *line = func.mJumpToLine;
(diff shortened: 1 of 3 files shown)


kczx3:
👍  This inconsistency always bothered me

lvalkov:
Hoping to see this make it in.

joedf:
Somewhat agree, but this will break old scripts. If this does make it in, I do hope that there is a clear error message for this.

HelgeffegleH:
@joedf , don't worry, this PR is for the [alpha](https://github.com/Lexikos/AutoHotkey_L/tree/alpha) branch. In v2, invalid arg types yields an error message 👍. In _v1_, `errorlevel` is set instead 👎.

Cheers.

PR #122, opened 2018-11-23
Replacing F option for BIF_Sort with Function parameter.

HelgeffegleH:
Reason, to enable passing functors, for convenince and consistency with other functions which calls user functions. (Regex callouts aside.)

### New:

```autohotkey
Sort String [,Options, Function]
```

`Function`, string or object, a function name or reference.

### About the implementation.

No attempts to validate _parameter count_, _byref_ of _isbuiltin_ for the functor is done.

`SortUDF` returns `0` on `FAIL` and `INVOKE_NOT_HANDLED`. On `INVOKE_NOT_HANDLED`, no more attemps to call the user's function are made and an invalid parameter exception is thrown after `qsort` returns. Handling `INVOKE_NOT_HANDLED` is useful to avoid something like `sort str,, []`, to silently sort the string deeming all items being equal.

Consider using [`qsort_s`](https://msdn.microsoft.com/en-us/library/4xc60xas.aspx) instead, to avoid the use of the `g_xxx` variables. 

Cheers.

### DIFF
--- a/source/globaldata.cpp
+++ b/source/globaldata.cpp
@@ -146,8 +146,9 @@ MsgMonitorList g_MsgMonitor;
 UCHAR g_SortCaseSensitive;
 bool g_SortNumeric;
 bool g_SortReverse;
+bool g_SortFuncIsCallable;
 int g_SortColumnOffset;
-Func *g_SortFunc;
+IObject *g_SortFunc;
 
 // Hot-string vars (initialized when ResetHook() is first called):
 TCHAR g_HSBuf[HS_BUF_SIZE];
--- a/source/globaldata.h
+++ b/source/globaldata.h
@@ -123,8 +123,9 @@ extern MsgMonitorList g_MsgMonitor;
 extern UCHAR g_SortCaseSensitive;
 extern bool g_SortNumeric;
 extern bool g_SortReverse;
+extern bool g_SortFuncIsCallable;
 extern int g_SortColumnOffset;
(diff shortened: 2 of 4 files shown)


Lexikos:
See ac30f425.  I think my reason for not integrating it was that I felt that this may not be the final syntax for the Sort function (partly because it should support arrays).  The problem is: I really don't care about Sort.

I took the opportunity to fix propagation of `Exit` and exceptions. Are there other significant differences between my implementation and yours?

HelgeffegleH:
Hello.

> Are there other significant differences between my implementation and yours?

Significant, no.

I think _my exception_ when passing something like `[]` is probably rarely needed, but helpful when it occurs.

I call `AddRef()` on the functor for the (fantastic) case where the user _deletes_ the functor from the UDF or (perhaps more likely) from an interruption of the UDF.

Your branch gives a trivial compile error on __Win32__. (needs one `INT_PTR` to be `__int64`)

It seems we both might miss to `free(item)` when we `goto end;` right after `qsort` returns.

I will add `free(item)` to my branch for the sake of completeness.

>  it should support arrays

If a dedicated linear array type is added, it could have a sort method. I do not think `Sort` needs to support the current object implementation.

> The problem is: I really don't care about Sort.

It is OK 😜 .

Cheers.

Lexikos:
Thanks for the analysis.  I think we have the best of both now.

PR #125, opened 2018-12-16
changing operators << and >> to do logical shift, zero-filling.

HelgeffegleH:
new:

`a << b`, performs logical left bit-shift of `a` by `b` positions.
`a >> b`, performs logical right bit-shift of `a` by `b` positions.

In both cases, vacant bit-positions are zero-filled.

Reasons,
__1)__ Logical shift is generally more useful, simple to understand and work with.
__2)__ The behaviour of left shifting a (signed) non-negative number such that the result is not representable in the corresponding unsigned type, or left shifting a negative number, is undefined.

reference:
https://docs.microsoft.com/en-us/cpp/cpp/left-shift-and-right-shift-operators-input-and-output?view=vs-2017 (2018-12-11)

### DIFF
--- a/source/script_expression.cpp
+++ b/source/script_expression.cpp
@@ -1123,8 +1123,8 @@ LPTSTR Line::ExpandExpression(int aArgIndex, ResultType &aResult, ResultToken *a
 				case SYM_BITAND:		this_token.value_int64 = left_int64 & right_int64; break;
 				case SYM_BITOR:			this_token.value_int64 = left_int64 | right_int64; break;
 				case SYM_BITXOR:		this_token.value_int64 = left_int64 ^ right_int64; break;
-				case SYM_BITSHIFTLEFT:  this_token.value_int64 = left_int64 << right_int64; break;
-				case SYM_BITSHIFTRIGHT: this_token.value_int64 = left_int64 >> right_int64; break;
+				case SYM_BITSHIFTLEFT:  this_token.value_int64 = (unsigned __int64) left_int64 << right_int64; break;
+				case SYM_BITSHIFTRIGHT: this_token.value_int64 = (unsigned __int64) left_int64 >> right_int64; break;
 				case SYM_FLOORDIVIDE:
 					// Since it's integer division, no need for explicit floor() of the result.
 					// Also, performance is much higher for integer vs. float division, which is part


nnnik:
This has been bugging me for years

Lexikos:
Putting aside what the C/C++ standards say, I have no problem with the current behaviour; I have scripts that rely on it. Perhaps you could explain how logical right shift is more useful with signed values.

I think that understanding the result of bit shifting a value of a signed type necessitates understanding its bit representation (i.e. the significance of the sign bit). Regardless of how the bit shifting operators behave, the value we are working with has a signed type. Left shift will still have the possibility of a negative result. If one understands the signed type, I do not see how zero-filling would be easier (or harder) to understand than extending the sign bit. If one does not understand the signed type, one has a more general problem.

As for C++ "defined" behaviour...

"A left shift is a logical shift (the bits that are shifted off the end are discarded, including the sign bit)," says Microsoft. To give such a clear description of what happens, and then say that the result is "undefined" for negative signed values, is silly (I assume they say it's "undefined" merely because the C++ standard says it's "undefined"). As a matter of fact, MSVC++ produces the same machine code for left-shifting signed and unsigned __int64, which makes sense given that it's logical shift in either case.

It might interest you to know that your alternative, with the implicit unsigned to signed conversion, technically relies on "implementation defined behaviour". Right-shifting a negative number is also implementation-dependent.

Standard C++ is not our target platform.

HelgeffegleH:
> "A left shift is a logical shift (the bits that are shifted off the end are discarded, including the sign bit)," says Microsoft. To give such a clear description of what happens, and then say that the result is "undefined" for negative signed values, is silly 

that is just what Microsoft does, a few lines below,

> If you left-shift a signed number so that the sign bit is affected, the result is undefined.

I guess I got stuck on that line the last time I read it, reading it again, I agree that they define the behaviour. So it is fine. 

> Perhaps you could explain how logical right shift is more useful with signed values.

Perhaps I could, but since, 

>  I have no problem with the current behaviour; I have scripts that rely on it.

I can admit this PR was a mistake, there is no reason to change the behaviour, I will return with a suggestion to add a (pair of) logical shift operator(s) instead. (possibly `>>>` (and `<<<`, which would behave identical as `<<` ))

I know what my _implementation_ does, and I do not have any problems with implementation defined behaviour, it is clear we are not working on anything portable here.

Cheers.

Lexikos:
> that is just what Microsoft does, a few lines below,

I know.  That's exactly what I was referring to.

PR #128, opened 2019-01-08
Update mt19937ar-cok.h

a-raccoon:
MT19937 Mersenne Twister random number generator website has moved since.  Also, this function has since been updated in 2004, and there is also a 64-bit machine version from 2005.  Also, they no longer have any license requirements as of 2001.

### DIFF
--- a/source/mt19937ar-cok.h
+++ b/source/mt19937ar-cok.h
@@ -38,7 +38,8 @@
    SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
    Any feedback is very welcome.
-   http://www.math.keio.ac.jp/matumoto/emt.html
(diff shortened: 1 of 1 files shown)


Lexikos:
The correct URL was already in the cpp file, along with the correct email address.

All of the versions I checked on the website contain license requirements similar (perhaps identical) to those already listed in the files contained by this project, which show "Copyright (C) 1997 - **2002**".

It is unclear exactly what the 2004 changes were; there is one comment indicating there was a change to one of the (unused) functions in 2004, but there are other changes (at least some of them are clearly superficial).

"64-bit machine version" apparently means a version which generates 64-bit random numbers (using standard C which should compile on x86 as well).

There is also a faster version which uses SIMD instructions and supports both 32-bit and 64-bit output.

PR #129, opened 2019-01-08
Changing FTOA to use g specifier.

HelgeffegleH:
__Reasons__, readability when printing  and preserving the numeric value for floats closer to zero.

Example, float close to zero,

```autohotkey
d := 1e-18
msgbox (float(string(d)) == d) . '`n' . d
```

The comparison is `true` in this branch, and the number is printed in _scientific notation_. The current behaviour yields `false` for the compairson and the number prints as `0.0` 

___

Note, this branch also fixes truncation errors due to the function being called with too small buffer, example,

```autohotkey
d := 1e300
msgbox (float(string(d)) == d) . '`n' . d 
```

The comparison is `true` in this branch, and the number prints in _scientific notation_. The current behaviour yields `false` for the compairson and a truncated output. The truncation is due to `MAX_NUMBER_SIZE` being too small, needed size is over `300` compared to the current `256`.

### About the implementation

If the number is not printed in _scientific notation_, it always has a decimal point followed by at least one `0`. Some numbers are printed in _scientific notation_ without a decimal point, if something like #101 is implemented, this has to be changed. However, I think it is probably best to always treat _scientific notation_ as float, and #101 should be closed (But the problem with `Integer(sci_not_string)` needs to be remembered).

To consider, many callers calls `FTOA` with a buffer of size  `MAX_NUMBER_SIZE`, that is (in this branch) much more than needed. Around `32` characters is probably always sufficient.

Cheers.

__Edit__: Spelling, formatting.

### DIFF
--- a/source/util.cpp
+++ b/source/util.cpp
@@ -3041,22 +3041,24 @@ LPTSTR InStrAny(LPTSTR aStr, LPTSTR aNeedle[], int aNeedleCount, size_t &aFoundL
 
 int FTOA(double aValue, LPTSTR aBuf, int aBufSize)
 // Converts aValue to a string while trying to ensure that conversion back to double will
-// produce the same value.  Trailing 0s after the decimal point are stripped for brevity.
-// Caller must ensure there is sufficient buffer size to avoid truncating the decimal point.
+// produce the same value.  Trailing 0s after the decimal point are stripped for brevity, when not printed in scientific notation.
+// Numbers printed in scientific notation may not contain a decimal point.
+// Caller must ensure there is sufficient buffer size to avoid truncating the output.
 {
-	int result = sntprintf(aBuf, aBufSize, _T("%0.17f"), aValue);
-	for (int i = result; i > 0; --i)
+	int result = sntprintf(aBuf, aBufSize, _T("%.17g"), aValue);
+	
+	// the 'g' specifier might cause the result to lack a decimal point. 
+	// If the number is not written in scientific notation, and lacks a decimal point,
+	// add ".0" to make the string look like a float.
+	size_t search_result = _tcscspn(aBuf, _T(".e")); 
+	if (search_result == result			// if true, no decimal point, '.', or 'e' was found, add ".0",
+		&& result + 3 <= aBufSize		// but only if the buffer has room for two more characters and the null terminator,
+		&& isdigit(aBuf[result - 1]))	// and the number isn't some variation of inf or NaN.
 	{
-		if (aBuf[i - 1] != '0')
-		{
-			if (i < result)
-			{
-				if (aBuf[i - 1] == '.')
-					++i;
-				aBuf[i] = '\0';
-			}
-			return i;
-		}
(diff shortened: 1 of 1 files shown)


Lexikos:
I had taken MAX\_NUMBER\_SIZE at face value, and not realized that floating-point numbers can exceed it.

I don't think I'd realized the precision specifier controls how high the exponent can be before scientific notation is used.  Allowing 16 digits of precision before collapsing the string seems very reasonable.

This does cause the result to differ in some cases, even when scientific notation is not being used. For example:

    MsgBox 1234567890123456.789
    ; before: 1234567890123456.75
    ; after:  1234567890123456.8

The goal of the previous format was to get the string as close to accurate as possible given the imprecision of representing a decimal number with a binary type, rather than hiding the imprecision by rounding.

HelgeffegleH:
I noted that, _before_ is __1234567890123456.7__ on Win32 build, while _after_ is __1234567890123456.8__ on both builds. However, the _binary_ value is accurate, both _before_ and _after_, on all builds.

The number __1234567890123456.75__ is exactly representable as a float (since it is equal to `4938271560493827/2^2`). So it seems a better (visually) result would be achieved with `"%.18g"`  rather than the proposed `"%.17g"`. This will yield __1234567890123456.75__ on both builds. However, take __0.1__, this is not exactly representable as a float, as is commonly known, instead it yields __0.1000000000000000055511151231257827021181583404541015625__, for this we would need `"%.55g"`. The exact number from the result of `2.0**-1022` would require `"%.715g"` (tested with `format`).

Cheers.

Lexikos:
> I noted that, before is 1234567890123456.7 on Win32 build

That's interesting.

> However, the binary value is accurate, both before and after, on all builds.

I think you mean to say that the binary value is *the same* on both builds. On both builds it is inaccurate in the same way.

> This will yield 1234567890123456.75 on both builds. 

No, it yields `1234567890123456.7` on 32-bit builds. So does `Format("{:.18g}", 1234567890123456.789)`, of course.

> However, take 0.1

My conclusion is that neither approach has the advantage of perfect accuracy. My original goal was to make the inaccuracy of the floating-point format reasonably obvious (rather than hiding it by formatting to 6 decimal places), and it seems both approaches are good enough at that.

HelgeffegleH:
> No, it yields 1234567890123456.7 on 32-bit builds. So does Format("{:.18g}", 1234567890123456.789), of course.

I mentioned format because I got different result when using it, exactly as you did above. Further investigating it I see the difference comes from using the latest alpha release vs. building this branch on my PC.

```autohotkey
msgbox Format("{:.18g}", 1234567890123456.789)
```

This gives me `1234567890123456.7` on the latest Win32 alpha release, but gives me 1234567890123456.75 when I build this branch myself (for both x64 and Win32 (Release)), I use 

```
Microsoft Visual Studio Community 2015
Version 14.0.25431.01 Update 3.
```

I do not think it is very important.

> My conclusion is that neither approach has the advantage of perfect accuracy

Agree.

PR #130, opened 2019-01-22
Adding verification that string is null-terminated for varsetcapacity(var, -1)

HelgeffegleH:
### New

throws exception if string is not null-terminated.

__Reason__, helps detect mistakes, avoids reading outside the variable's string buffer.

### Example

```autohotkey
varsetcapacity var, 8
strput 'abcde', &var, 5
varsetcapacity var, -1 ; Error:  String not null-terminated.
```

Cheers.

### DIFF
--- a/source/script.cpp
+++ b/source/script.cpp
@@ -13070,7 +13070,13 @@ int Line::FormatError(LPTSTR aBuf, int aBufSize, ResultType aErrorType, LPCTSTR 
 	return (int)(aBuf - aBuf_orig);
 }
 
-
+ResultType Script::CriticalScriptError(LPCTSTR aErrorText, LPCTSTR aExtraInfo)
+{
+	// Displays an error message and terminates the script.
+	g->ExcptMode = EXCPTMODE_NONE; // Do not throw an exception.
+	ScriptError(aErrorText, aExtraInfo);
+	return ExitApp(EXIT_CRITICAL); // Called this way, it will run the OnExit function, which is debatable because it could cause more good than harm, but might avoid loss of data if the OnExit function does something important.
+}
 
 ResultType Script::ScriptError(LPCTSTR aErrorText, LPCTSTR aExtraInfo) //, ResultType aErrorType)
 // Even though this is a Script method, including it here since it shares
--- a/source/script.h
+++ b/source/script.h
@@ -236,7 +236,7 @@ enum CommandIDs {CONTROL_ID_FIRST = IDCANCEL + 1
 #define ERR_INVALID_HWND _T("Invalid HWND.")
 #define ERR_INVALID_USAGE _T("Invalid usage.")
 #define ERR_INTERNAL_CALL _T("An internal function call failed.")
-
+#define ERR_STRING_NOT_TERMINATED _T("String not null-terminated.")
 #define WARNING_USE_UNSET_VARIABLE _T("This variable has not been assigned a value.")
 #define WARNING_LOCAL_SAME_AS_GLOBAL _T("This local variable has the same name as a global variable.")
 #define WARNING_USE_ENV_VARIABLE _T("An environment variable is being accessed; see #NoEnv.")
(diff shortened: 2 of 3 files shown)


Lexikos:
It's a good idea. However, I wonder if heavier consequences would be more appropriate. If the string is not null-terminated, that effectively means something has written outside the bounds of the variable's capacity (as reported by VarSetCapacity). All variables **must** be null-terminated. If the variable is not null-terminated, it may be unsafe for the program to continue executing, even if the thread is exited - because the variable might be read by another thread, or other data may have been corrupted.

Related: DllCall exits the program if `Var::sEmptyString` has been overwritten with a non-null value (i.e. because the address of a zero-capacity variable has been passed to the function for output).

HelgeffegleH:
Yes.

Related:  I guess `'str', var` should check this too,

HelgeffegleH:
Regarding 

> (attempt to)

An `onexit` function can stop the script termination as documented. Perhaps this is not suitable, at least not for the case when `Var::sEmptyString` was written to, perhaps `EXIT_DESTROY` should be passed to `Script::ExitApp` to avoid this.

If this isn't accepted, note that `BIF_DllCall` should use `_f_return_FAIL` after calling `g_script.ExitApp(EXIT_CRITICAL)` to avoid that the thread continues if an `onexit` functions stops the script termination. Issue example,

```autohotkey
onexit ()=>1
dllcall callbackcreate((p)=>numput(1, p)), 'str', v
msgbox
```

Lexikos:
Prior to commit b912b6ab, ExitApp() had a mode where it displays a "Critical Error" message and terminates, bypassing OnExit. I removed it to reduce code size, since it wasn't actually in use. Perhaps it was supposed to be used by DllCall (there are comments in DllCall indicating the OnExit routine is called, but maybe these were added after the original intention was forgotten).

The dialog displayed both "The program is now unstable and will exit." and "The current thread will exit.", which could technically be true if both the program and the thread exited, but sometimes one or both did not, as you pointed out. I consider these bugs, in both v1 and v2.  f3ef98c9 fixes them. 7daa585e fixes a silly error I noticed while looking at this.

LineError interprets CRITICAL_ERROR as a signal to exit the script after displaying the message, but it was not signalling ExitApp to exit unconditionally, and it turns out CRITICAL_ERROR was never actually passed to LineError. I fixed that, and used it to replace "The current thread will exit." with "The program is now unstable and will exit.", also eliminating the need to duplicate this part of the error string.

I also removed the _f_critical_error macro, since there's never any need or opportunity to return after terminating the process, nor any dependency on the built-in "ABI".

<hr />

Running the OnExit thread might not be the right thing to do. One can predict a few outcomes from a null terminator being overwritten (not all mutually exclusive):
  - Attempts to read the variable might read into subsequent bytes, with the value receiving an unexpected suffix.
  - Attempts to read the variable might scan for a null terminator and find none, eventually scanning into an uncommitted page. In this case, the program crashes due to "access violation".
  - The variable is never read again, or not read before being assigned some other value. In that case, the problem *might* go unnoticed, but...
  - Maybe subsequent bytes were also overwritten. It is difficult to know what - could be other variables, objects, or anything else AutoHotkey might allocate dynamic memory for.
  - If the variable had zero capacity, every variable that has zero capacity may be affected. In that case, the script's logic can be broken in a fairly predictable way:
```AutoHotkey
NumPut(1, &v, "char")
;...
if (uninitialized = "")
    MsgBox "Empty"
else
    MsgBox "Not empty"
```
Since this has the potential to totally break the logic of the script (not just preventing it from working, but potentially causing it to do things that it shouldn't), it's probably better to not try to execute any more script.

Chris opined:
> Called this way, it will run the OnExit routine, which is debatable because it could cause more good than harm, but might avoid loss of data if the OnExit routine does something important.

But conversely, attempting to execute the OnExit routine under unreliable conditions could cause more harm than good, and cause more loss of data than not executing it. If the problem went unnoticed and later caused an access violation, the process would terminate without executing the OnExit routine anyway.

PR #131, opened 2019-01-22
Adding ability to update the internally stored string length for strings stored in objects.

HelgeffegleH:
### New

pass `-1` for the second paramter of `SetCapacity` to update the string length.

__Reason__, the user might change the string via `dllcall` or `strput`.

__About__, throws exception if string is not null-terminated, similar to #130 .

### Example

```autohotkey
o := []
o.setcapacity 1, 8
strput 'abc', o.getaddress(1), 4
o.setcapacity 1, -1		; new
msgbox o.1 			; 'abc' in this branch, else blank.
```

### Note

the use of `-1` is chosen for symmetry with `varsetcapacity`, not because it is considered better than having, more flexible, separate methods/functions for updating string lengths.

Cheers.

### DIFF
--- a/source/script.h
+++ b/source/script.h
@@ -236,7 +236,7 @@ enum CommandIDs {CONTROL_ID_FIRST = IDCANCEL + 1
 #define ERR_INVALID_HWND _T("Invalid HWND.")
 #define ERR_INVALID_USAGE _T("Invalid usage.")
 #define ERR_INTERNAL_CALL _T("An internal function call failed.")
-
+#define ERR_STRING_NOT_TERMINATED _T("String not null-terminated.")
 #define WARNING_USE_UNSET_VARIABLE _T("This variable has not been assigned a value.")
 #define WARNING_LOCAL_SAME_AS_GLOBAL _T("This local variable has the same name as a global variable.")
 #define WARNING_USE_ENV_VARIABLE _T("An environment variable is being accessed; see #NoEnv.")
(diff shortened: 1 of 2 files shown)


Lexikos:
I might accept this as a short-term solution. However, my preference is for one of the following (I am undecided):
  - Add a function and method for setting length and remove the -1 mode of VarSet**Capacity**.
  - Require the use of a separate data type or object for byte or character buffers.  Length should be a property of the value, not of the variable or field.  (I suppose I'm gravitating toward composability.)

nnnik:
> * Require the use of a separate data type or object for byte or character buffers.  Length should be a property of the value, not of the variable or field.  (I suppose I'm gravitating toward composability.)

^Yes please

Lexikos:
SetCapacity is being removed, since it would otherwise require duplication for properties, arrays and associative arrays. Binary data should use the new [Buffer](http://lexikos.github.io/v2/docs/objects/Buffer.htm) object, while strings can just do without for now. SetCapacity for strings was a micro-optimization - generally unnecessary, and it would probably be more effective to redesign the script around variables rather than objects.

Eventually I want to eliminate the concept of "a variable's string buffer", so the string (or string buffer) is just a value that can exist within any container. (I kind of already said this, but at the time I didn't anticipate making such progress in a short span of time.)

The shift to Buffer objects for binary data should also obviate the need for the capability to "update" a string's length, since you should no longer be modifying the string indirectly. A Buffer has no concept of length (only Size = capacity in bytes), and can only be interpreted as a string explicitly, by Buffer.Data or StrGet.

PR #132, opened 2019-02-18
Changing Hotstring::Hotstring to not assign mReplacement when...

HelgeffegleH:
... the `X` option is specified.

Reason, avoids setting the label/function name, as the replacemnent string.

Issue example,

```autohotkey
hotstring(":X:a", "f")
f(){
}
```
which results in the hotstring `a` being replaced by `f` rather than calling the function `f`.


JnLlnd:
If I may ask... Is this somehow related to the issue described here:
https://www.autohotkey.com/boards/viewtopic.php?t=59482

HelgeffegleH:
Yes, it is.

JnLlnd:
Good. Thanks for taking care of this.

PR #133, opened 2019-03-02
Added ATan2() built in function.

a-raccoon:
Is this how it's done?


HelgeffegleH:
> Is this how it's done?

No. You have not defined a function, only enabled `transform var, atan2, value` to be executed and _fail_.

Cheers.

a-raccoon:
But the function is already pre-defined and has existed in code for eons.

jeeswg:
I wrote some C++ code for an ATan2 function. [Functions](https://autohotkey.com/boards/viewtopic.php?f=23&t=54392). [More code](https://autohotkey.com/boards/viewtopic.php?f=23&t=54394).

If requested, I can turn some of this code into AHK v1/v2 pull requests, and do any edits.

a-raccoon:
atan2 is already in math.h, the c++ common library.  no need to write anything.

jeeswg:
I essentially cloned BIF_ATan (from script2.cpp) replacing qmathAtan with qmathAtan2. It's a two-liner.

a-raccoon:
qmath isn't used and should be destroyed.  Scroll to the bottom and you'll see that all the redefines at the bottom.

Lexikos:
qmath is used in 32-bit builds to reduce code size and increase performance.  It is not used in 64-bit builds because it is full of 32-bit asm.

PR #135, opened 2019-04-05
adding operators >>>, <<<, >>>= and <<<=

HelgeffegleH:
### New:

`a <<< b`, performs logical left bit-shift of `a` by `b` positions.
`a >>> b`, performs logical right bit-shift of `a` by `b` positions.

In both cases, vacant bit-positions are zero-filled. The corresponding compound assignment operators are `>>>=` and `<<<=`.

__Reason__: Convenience.

The reason for adding  `<<<` which behaves identical to `<<` is for _symmetry_ only. If `>>>` is used for logical rightshift, `<<<` is unlikely to be used for anything else. I'd be happy to change the operators to something shorter if anyone has a good suggestion.

### DIFF
--- a/source/defines.h
+++ b/source/defines.h
@@ -171,7 +171,7 @@ enum SymbolType // For use with ExpandExpression() and IsNumeric().
 #define SYM_OPAREN_FOR_CPAREN(symbol) ((symbol) + (SYM_OPAREN - SYM_CPAREN)) // Caller must confirm it is CPAREN or CBRACKET.
 #define YIELDS_AN_OPERAND(symbol) ((symbol) < SYM_OPAREN) // CPAREN also covers the tail end of a function call.  Post-inc/dec yields an operand for things like Var++ + 2.  Definitely needs the parentheses around symbol.
 	, SYM_ASSIGN, SYM_ASSIGN_ADD, SYM_ASSIGN_SUBTRACT, SYM_ASSIGN_MULTIPLY, SYM_ASSIGN_DIVIDE, SYM_ASSIGN_FLOORDIVIDE
-	, SYM_ASSIGN_BITOR, SYM_ASSIGN_BITXOR, SYM_ASSIGN_BITAND, SYM_ASSIGN_BITSHIFTLEFT, SYM_ASSIGN_BITSHIFTRIGHT
+	, SYM_ASSIGN_BITOR, SYM_ASSIGN_BITXOR, SYM_ASSIGN_BITAND, SYM_ASSIGN_BITSHIFTLEFT, SYM_ASSIGN_BITSHIFTRIGHT, SYM_ASSIGN_BITSHIFTRIGHT_LOGICAL // SYM_ASSIGN_BITSHIFTLEFT_LOGICAL doesn't exist but <<<= is the same as <<=
 	, SYM_ASSIGN_CONCAT // THIS MUST BE KEPT AS THE LAST (AND SYM_ASSIGN THE FIRST) BECAUSE THEY'RE USED IN A RANGE-CHECK.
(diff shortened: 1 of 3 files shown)


Drugwash2:
Maybe `<|` or `<-` would be shorter.

**Off-topic:**
While we're working on bit-wise operations: any chance for a **Rotate** set of operators or functions? Shifting is alright for specific operations but sometimes bit rotation may be needed. I have an old script that required some form of pseudo-encryption and the only way I could do bit rotation byte-by-byte was through machine code, which unfortunately doesn't work in x64.

Come to think about it, a set of functions would be much more flexible and thus useful than operatiors. One could choose a **buffer** (variable), **rotation mode** (byte/word/dword/qword), **number of rotations**(max bitcount-1) and optionally **starting point** and **length**.

I.e: **ROL(myvar, "B", 5, 10, 150)** would mean *'rotate left the contents of myvar one Byte at a time for five positions (bits), starting with byte 10 for a total length of 150 bytes'*. The rest of `myvar`'s contents would remain unaltered in the result. The result may be the best choice from:
- put back to `myvar` (if it's a ByRef) returning `True` or `False` for success or failure
- returned by the function while setting `ErrorLevel` to `False` or `True` for success or failure
- optionally put in another variable if specified by the user as a sixth parameter, leaving `myvar` unaltered and returning `True` or `False` for success or failure

Sorry for the off-topic, the idea just suddenly came to me and was afraid to lose it.

HelgeffegleH:
> Maybe <| or <- would be shorter.

Certainly, it is shorter. Especially for the compound assignments it would look weird, `<|=` and `<-=`. 

Off topic is better suited elsewhere.

Cheers.

Drugwash2:
Well, personally I look at symbols in a slightly different way: I see them graphically. So to me they both look like arrows. Especially the first one (`<|` and `|>`) looks like an arrow head, nicely completed by the equal sign for assignments, making it a full arrow (`<|=` and `=|>`, respectively). Additionally, the vertical bar could be associated with the idea of **logical** shift by *mentally* reading it as a lowercase L. Anyway, this was my reasoning for suggesting those symbols, for what it's worth. Maybe someone else could come up with something better.
Thanks and good luck!

HelgeffegleH:
the `<` and `>`  indeed helps visualise direction. It would be `<|=` and `|>=` though. Or even `<|` , `>|`, `<|=` and `>|=`, it wasn't (initially) clear what the suggestion would be for the right shift direction.

Cheers and thank you for your suggestions.

Shambles-Dev:
`>>>` is what Java uses for logical shift right and is the closest thing to a standard that I am aware of.  `<<` and `>>` are arithmetic shifts in most programming languages.  There is no difference between arithmetic and logical shift left, which is why there is no `<<<` operator.

Logical shift right is useful for 'chopping up a bit field', like those used in binary files and network packets.  The arithmetic shifts are often used as faster alternatives to multiplication and division by powers of 2.

I am not aware of any popular programming language with rotate operators (probably because they are rarely used outside cryptography and hash functions), but there are many with libraries that provide left and right rotates, both with and without carry.  Most such libraries include [other obscure bitwise operations](https://en.wikipedia.org/wiki/Bit_Manipulation_Instruction_Sets).

jeeswg:
This is a great PR, thanks Helgef.

The symbols >>> and >>>= are elegant, and >>> is relatively common for the purpose stated:
[Logical shift - Wikipedia](https://en.wikipedia.org/wiki/Logical_shift)
[Bitwise operations - Rosetta Code](https://rosettacode.org/wiki/Bitwise_operations)

Since one day <<< might have a new useful meaning, and we'd have to do a breaking change, I wouldn't implement <<< or <<<=.

And as you stated, <<< and <<<= would be functionally identical to << and <<=.

Kind regards.

fincs:
`<<<` and `<<<=` are 100% redundant and they should be removed.

Lexikos:
I chose to remove `<<<` and `<<<=`.

PR #136, opened 2019-04-06
 adding new directive #LoadLibrary

HelgeffegleH:
### New:
```autohotkey
#LoadLibrary path_to_dll
#LoadLibrary dir
```
`#LoadLibrary path_to_dll`: Loads the dll at `path_to_dll`. `path_to_dll` can be a relative path or just a filename. `path_to_dll` may be preceded by `*i`, to prevent load time error if `path_to_dll` doesn't exist. This is useful for optional functionality and `fileinstall`.

`#LoadLibrary dir`: Alters the search path for subsequent uses of the directives which doesn't specify a direct path to a _dll_. Similar to `#include dir`, but this calls `SetDllDirectory`. Omitting _dir_, i.e, a lone `#LoadLibrary` restores the default search path for `LoadLibrary` (calls `SetDllDirectory(NULL)`).

Note: the script can call `FreeLibrary` once per `#LoadLibrary` if wanted, otherwise the system will free the library when the application terminates. If calling `FreeLibrary` the script should not execute any _dllcalls_ which has been subject to loadtime optimisation for the relevant dll.

Above, _dll_ can be _exe_.

__Reasons__: enables load time optimisation for `dllcall("dll\function")` which improves performance and convenience, and readability over manually loading and looking up function address via `LoadLibrary` + `GetProcAddress` at run time.

### About the implementation:

Without any fear, changed `_WIN32_WINNT` from `0x0501` to `0x0502`. If this is unacceptable, `#LoadLibrary dir` can be removed and we can use `#include dir` instead, which will call `SetCurrentDirectory` which is against the recommendation in the documentation for `LoadLibrary`.

Cheers.

### DIFF
--- a/source/script.cpp
+++ b/source/script.cpp
@@ -3524,6 +3524,44 @@ inline ResultType Script::IsDirective(LPTSTR aBuf)
 #endif
 	}
 
+	if (IS_DIRECTIVE_MATCH(_T("#LoadLibrary")))
+	{
+		if (!parameter)
+		{
+			// an empty #LoadLibrary restores the default search order.
+			if (!SetDllDirectory(NULL))
+				return ScriptError(ERR_INTERNAL_CALL);
+			return CONDITION_TRUE;
+		}
+		// ignore failure if path is preceeded by "*i".
+		bool ignore_load_failure = (parameter[0] == '*' && ctoupper(parameter[1]) == 'I'); // Relies on short-circuit boolean order.
+		if (ignore_load_failure)
+		{
+			parameter += 2;
+			if (IS_SPACE_OR_TAB(*parameter)) // Skip over at most one space or tab, since others might be a literal part of the filename.
+				++parameter;
+		}
+		LPTSTR library_path;	// needs to be freed before return, unless DerefInclude return FAIL.
+		if (!DerefInclude(library_path, parameter))
+			return FAIL;
+		DWORD attr = GetFileAttributes(library_path);
+		if (attr != 0xFFFFFFFF && (attr & FILE_ATTRIBUTE_DIRECTORY)) // File exists and its a directory
+		{
+			BOOL result = SetDllDirectory(library_path);
+			free(library_path);
+			if (!result)
+				return ScriptError(ERR_INTERNAL_CALL);
+			return CONDITION_TRUE;
+		}
+		ResultType result = CONDITION_TRUE;	// set default
+		HMODULE hmodule = LoadLibrary(library_path);
+		if (!hmodule					// the library couldn't be loaded
+			&& !ignore_load_failure)	// no *i was specified
+			result = ScriptError(_T("Failed to load DLL."), library_path);
+		free(library_path);
+		return result;
+	} // end #LoadLibrary
+
 	if (IS_DIRECTIVE_MATCH(_T("#NoTrayIcon")))
 	{
 		g_NoTrayIcon = true;
@@ -12423,7 +12461,7 @@ BIF_DECL(BIF_PerformAction)
 
 
 ResultType Script::DerefInclude(LPTSTR &aOutput, LPTSTR aBuf)
-// For #Include and #IncludeAgain.
+// For #Include, #IncludeAgain and #LoadLibrary.
 // Based on Line::Deref above, but with a few differences for backward-compatibility:
 //  1) Percent signs that aren't part of a valid deref are not omitted.
 //  2) Escape sequences aren't recognized (`; is handled elsewhere).
--- a/source/script.cpp
+++ b/source/script.cpp
@@ -3555,7 +3555,13 @@ inline ResultType Script::IsDirective(LPTSTR aBuf)
 		}
 		ResultType result = CONDITION_TRUE;	// set default
 		HMODULE hmodule = LoadLibrary(library_path);
-		
+
+		if (hmodule)
+			// "Pin" the dll so that the script cannot unload it with FreeLibrary.
+			// This is done to avoid undefined behaviour when DllCall optimizations
+			// resolves a proc address in a dll loaded by this directive.
+			GetModuleHandleEx(GET_MODULE_HANDLE_EX_FLAG_PIN, library_path, &hmodule);  // MSDN regarding hmodule: "If the function fails, this parameter is NULL."
+
 		if (!hmodule					// the library couldn't be loaded
 			&& !ignore_load_failure)	// no *i was specified
 			result = ScriptError(_T("Failed to load DLL."), library_path);
--- a/source/script.cpp
+++ b/source/script.cpp
@@ -3524,11 +3524,11 @@ inline ResultType Script::IsDirective(LPTSTR aBuf)
 #endif
 	}
 
-	if (IS_DIRECTIVE_MATCH(_T("#LoadLibrary")))
+	if (IS_DIRECTIVE_MATCH(_T("#DllLoad")))
 	{
 		if (!parameter)
 		{
-			// an empty #LoadLibrary restores the default search order.
+			// an empty #DllLoad restores the default search order.
 			if (!SetDllDirectory(NULL))
 				return ScriptError(ERR_INTERNAL_CALL);
 			return CONDITION_TRUE;
@@ -3555,12 +3555,13 @@ inline ResultType Script::IsDirective(LPTSTR aBuf)
 		}
 		ResultType result = CONDITION_TRUE;	// set default
 		HMODULE hmodule = LoadLibrary(library_path);
+		
 		if (!hmodule					// the library couldn't be loaded
 			&& !ignore_load_failure)	// no *i was specified
 			result = ScriptError(_T("Failed to load DLL."), library_path);
 		free(library_path);
 		return result;
-	} // end #LoadLibrary
+	} // end #DllLoad
 
 	if (IS_DIRECTIVE_MATCH(_T("#NoTrayIcon")))
 	{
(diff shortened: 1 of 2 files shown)


HelgeffegleH:
> we can use `#include dir` instead, which will call `SetCurrentDirectory`

I realised this isn't true without some additional work for the case where `#ifdef AUTOHOTKEYSC` is `true`.

Lexikos:
Requiring XP SP1 or later is acceptable, though I'm not sure whether this header change actually puts such a requirement into effect.

Currently I think the program can load on Windows 2000, but this will change for v2 as it requires the program to be built with VS 2010, which requires us to not use any newer language features.  VS 2019 supports targeting Windows XP, but I'm not sure whether that means XP SP0, XP SP3 or what.

As for the overall PR, I am reluctant to merge this as my long term goals include optimizing DllCall, and I haven't yet decided whether that will involve new syntax (such as [the DllFunction statement](https://autohotkey.com/board/topic/22732-dllcall-function-declaration/), [DynaCall()](https://hotkeyit.github.io/v2/docs/commands/DynaCall.htm) or [#DllImport](https://hotkeyit.github.io/v2/docs/commands/_DllImport.htm)). If the new syntax resolves the function at load time, #LoadLibrary would be fairly redundant.

Although, now that I'm thinking about it, I realize that the newer fat arrow syntax defeats one of the advantages of #DllImport and DllFunction: brevity.

    MyFunc(x,y,z) => dllcall("path\dll","str",x,"UInt",y,"Int",z,"CDECL Int")

Laszlo's suggestion was:

    #define MyFunc(x,y,z) dllcall("path\dll","str",x,"UInt",y,"Int",z,"CDECL Int")

I suppose that only the first one would provide an object for `Func("MyFunc")` filled with appropriate information.

One advantage of new syntax specifically for importing functions is that the function name can be validated at load time. This isn't possible with DllCall, since the function might not be required (such as when the script calls one or another function depending on OS), or the script might be designed to download or FileInstall the dll file.

HelgeffegleH:
> my long term goals include optimizing DllCall

That is good.

```
MyFunc(x,y,z) => dllcall("path\dll","str",x,"UInt",y,"Int",z,"CDECL Int")
```

In my mind. this should be two optimisations, i.e, `f(.) => g(.)` causes `f` to call `g` directly, So it is not specific to `dllcall`. Ofc, this could not be a macro, c.f, `#define` in __c__, they are useful but off topic here I think. 

The second optimisation, which relates specifically to  `dllcall`, is, I imagine, that calls to `dllcall` would in some cases be optimised (at load time) to be swapped for calls to another function, passing it as much pre-resolved data as possible. I suppose that effectively it could be done by replacing it with a call to (something like) a _DynaCall_ object.
___
> the DllFunction statement

I do not see any real suggestions there which relates to optimising dllcall, w.r.t. performance (execution speed). As for optimising dllcall w.r.t. convenince for the script author, I do not think this is important specifically for dllcall, it is already convenient to wrap dllcalls in UDFs, especially since nested functions was made available. Again, `#define` macros are useful, but not limited for use with dllcall.
___

>One advantage of new syntax specifically for importing functions is that the function name can be validated at load time. This isn't possible with DllCall, since the function might not be required (such as when the script calls one or another function depending on OS), or the script might be designed to download or FileInstall the dll file.

> #DLLImport

I do not know its exact behaviour, I would suggest it'd be used together with `#LoadLibrary` and only yield a load time error if a functions is not found but the dll was loaded. Ofc, a run time error would be raised if the script calls the function and it wasn't loaded. That way one  only needs to specify that a dll is optional in one place, i.e., for `#LoadLibrary` and the script has better control of the dll's reference count <b>[see bottom]</b>. A _new syntax_ should allow the script to be designed to download or FileInstall the dll.

> DynaCall

I don't see any real need for a new syntax for _"ParameterDefinition"_ and _default parameters_, I do not think there is a need for parameter reordering and calling the dll function with _object get syntax_. I think `#LoadLibrary` is useful together with this too.

Cheers.

___

Regarding the mention of a dll's reference count, I cut this for brevity but kept it here for clarity:

If `#DllImport` loads a dll, it can call `LoadLibrary()` for each instance of `#DllImport` which reference a particular dll, and then the script author must call `FreeLibrary()` for every one of these instances, if it wants to unload the dll. If `#DllImport` calls `LoadLibrary()` only when the dll isn't loaded, the script author cannot call `FreeLibrary()` without considering if the dll is still used by other script libraries, making the code less maintainable.

Lexikos:
> I do not see any real suggestions there which relates to optimising dllcall, w.r.t. performance (execution speed).

The point was that DllFunction or #DllImport would be *much easier* to optimize (less work and less risk of introducing bugs), so if I choose to implement them for any reason, optimized dll functions will arrive sooner.

> Again, `#define` macros are useful, but not limited for use with dllcall.

I think that was Laszlo's point.  If DllCall can be optimized as it is, #define provides the advantages of DllFunction or #DllImport while also being useful for other things. If I plan to implement #define, there are fewer reasons to implement DllFunction or #DllImport. (However, the latter may have some other advantages over #define, as I mentioned.)

> [#DllImport]: I do not know its exact behaviour, I would suggest it'd be used together with `#LoadLibrary` [...]

I gave examples of existing proposals or functionality, but anything I implement would use syntax and behaviour of my choosing (perhaps influenced by those others). The exact behaviour of the existing solutions is not too relevant.

I think it would be more intuitive for #DllImport to load the dll automatically. Also consider the possibility of a valid function name failing to resolve because the dll is an older version, such as on an older OS. The function may still be optional as far as the script is concerned. I do not think it makes sense to correlate the presence of #LoadLibrary (or the dll in memory) with the function being mandatory.

> A new syntax should allow the script to be designed to download or FileInstall the dll.

I don't see exactly how this relates to your overall argument. #LoadLibrary would not help on the first execution, or if the script deletes the dll after use, as the dll would not exist when the directive is processed. #DllImport or DllFunction could optionally permit the dll to be delay-loaded by "compiling" the function on first call. Then again, doing so by default would mean the author does not need to think about marking the function as optional (but would reduce the potential for load-time validation).

> The second optimisation [...] is, I imagine, [...]

It can go further. Back in 2008 I wrote an "optimization" that assembled x86 machine code to call the specific functions needed to convert the parameters, given the signature of the function. I didn't use it because I wasn't happy with the code size vs. benefit.

I've had a long time to think about it, and have come up with various ideas. For instance:
  - Most calls will not use variables in the parameter types, so the parameter type strings can be converted at load time even if the function is not known.
  - If the parameter types have been converted at load time, they can be eliminated from the expression evaluation phase.
  - The call site can be tagged with a struct which allows information to be cached, without requiring it to be found through static analysis - i.e. the function address can be resolved exactly once even when it is contained by a variable, as long as the value does not change between calls. (However, the use of LoadLibrary/FreeLibrary by the script complicates this.)
  - If we can generate "thunks" for dll functions (i.e. machine code that converts script parameters to native ones and calls the actual function), we can do it for our own functions too. Library functions then would not need to deal with parameter conversion, which means smaller code, easier to maintain, and more portable (easier to make the functions available to other languages).

> If #DllImport loads a dll, it can call LoadLibrary() for each instance of #DllImport which reference a particular dll,

Alternatively, #DllImport may [**pin** the dll](https://docs.microsoft.com/en-us/windows/desktop/api/libloaderapi/nf-libloaderapi-getmodulehandleexa#get_module_handle_ex_flag_pin-0x00000001).

> If #DllImport calls LoadLibrary() only when the dll isn't loaded, the script author cannot call FreeLibrary() without considering if the dll is still used by other script libraries, making the code less maintainable.

The script author should not attempt to unload the dll while the program still has pointers to its functions, or at all if s/he did not load the dll explicitly. That aside, I don't see these conflicts being at all likely to happen.

------

I think that I will merge this. However, I think #LoadLibrary vs. DllCall seems a bit incongruous. What do you think of #DllLoad? On the other hand, it seems you expect scripts to want to use FreeLibrary as well.

Lexikos:
If the script does `#LoadLibrary *i` but later wants to unload the dll, how will it know whether the dll was loaded (or reference count incremented) *by this directive*?

HelgeffegleH:
Hi. 

> If the script does #LoadLibrary *i but later wants to unload the dll, how will it know whether the dll was loaded (or reference count incremented) by this directive?

I made a logical mistake it seems. I originally though that you could check `GetModuleHandle` but ofc that doesn't work. I think that was part of the reason I didn't make a PR for an earlier implementation of this, which also had other issues, but I then forgot about it. I'm currently thinking

```autohotkey 
#LoadLibrary path_to_dll [, outputvar]
``` 

where `outputvar` would be a static variable if the directive is used inside a function, otherwise it would be global. I suppose the `*i` is redundant if you specify the `outputvar`. But it should be allowed to omit the output var and use the `*i`.

`LoadLibary` docs:

> If the function fails, the return value is NULL. To get extended error information, call GetLastError.

The last error would probably be the most useful value to put in the output var. So if `outputvar == 0` the dll was loaded successfully.

___
 
Regarding the name `#LoadLibrary`, I suppose `#DllLoad` is fine, but it forces me to capitalise the _l_ in _load_. I'm quite neutral.

___

I'm all for better optimisations, but if it takes 5-10 years, I think this would be useful in the mean time and that we could live with a deprecated directive. I also think it is fine to make minor breaking changes every few years, like 2.0.xxx -> 2.1.xxx. We do not have to wait for 3.0 before braking even the smallest of details.

Cheers.

HelgeffegleH:
I changed the directive name to `#DllLoad` and changed it to "pin" the dll to avoid problems if the script calls `FreeLibrary`.  The idea with the `outputvar` didn't seem worth the bother. Also, it is not entirely clear how directives with multiple parameters should be _used_. I suppose that the path parameter would need to escape any commas. But there could be other ways to do it too, each with its drawbacks and benefits I guess.

Cheers.

HelgeffegleH:
I intend to submit a suggestion for the documentation.

PR #142, opened 2019-07-04
introducing namespaces.

HelgeffegleH:
# Goal and motivation

The problems associated with large or shared namespaces and shared state across a large application are considered well known. The best known solution to the problem is to avoid it, by simply allowing _names_, as well as _state_, to be isolated within more manageable blocks or files.

The ultimate goal is to facilitate writing code which is much less __dependent__ on the script in which it is used, code that is more __modular__, and code which is more __predictable__, __readable__ and __maintainable__. This is code which is __faster to develop and integrate__, this is code which is __share friendly__, just like the _AHK community_.

___
# Design

See the [documentation](https://github.com/HelgeffegleH/NS_FILES_AND_DOCS/blob/master/AutoHotkey.chm) (Open __Usage and Syntax -> Namespaces__ ). The documentation is not complete, several aspects of the final design are undecided. Executables are also available [here](https://github.com/HelgeffegleH/NS_FILES_AND_DOCS)

## Threads

_Threads_ is __the__ breaking change, this will also affect scripts which doesn't define any additional namespaces. Fundamentally, it doesn't makes sense that a certain _thread_ has a certain setting for, eg, case sensitivity or string encoding, these kinds of settings are more suited for limited scopes rather than a thread which can call any function in any library. Settings are now confined to a namespace block or an _#imported_ file. 

Although namespaces gives script authors better control of settings (and often favourable performance), it would probably be better if all settings were function parameters for the relevant functions instead. Taking that step from this branch will not be more difficult than from the current _alpha_. This is another discussion though.

## Auto-execute sections

The new design allows better control of _one time_  tasks and initialisation of libraries and scripts. The confusing and unpredictable auto-exec. timer has been removed. The new design also allows multiple scripts to be merged much more easily, than the current design. 

## Function references

Currently, a reference to a function can be obtained via `Func('namespace_name', 'funcName')`, this is not matched by `isFunc` since there might be better ways to do this. One idea is to have a function, eg, `namespace(names*)` to return a _namespace object_. A namespace object could return a  func reference via a property or method. Variable references could be returned as _outputvars_ I guess, or the index operator could be used to set or get values of a namespace' variables.

## Directives

At present, only `#if`s belongs to namespaces, directives such as `#warn` are still applied to the entire script. If eg `#warn` would be applied to specific namespaces, perhaps it should be positional or _inherited down_ from a namespace to its nested ones. __Edit:__ Also `#include` belongs to namespaces, but I didn't handle the changing of working directory yet.

## Misc - unfinished / undecided

* __Window groups__, not namespace specific. I guessed the implementation regarding groups will be changed, so I postponed work on this.
* __Default base object__, not namespace specific. I think it should be but I have postponed this for now.
* __Ahk2Exe__, no work on the _compiler_ has been done. In my mind, `#import x.ahk -> y`, would simply be replaced with `namespace y { #include x.ahk } `. Optional imports would need additional work though, due to load time optimisations.
* It might be nice to add means to _use_ names from other namespaces, eg, `#use a() in b` to allow using `a()` without prefixing `b->`. If someone wants to lay out a complete suggestion, please do so.
* A lone scope resolution at the beginning of a line requires `()` to be a function call.
* `#Import` does not implement `<name>` yet, cf, `#Include <libName>`.
* `ErrorLevel` and `A_Args` are only added to the _Script_ namespace. Imo, `ErrorLevel` should be removed anyways. For `A_Args` I think this is fine, it could also be a _BIV_.
___

# About the implementation

This is a brief introduction to the implementation.

## `class NameSpace`

Each namespace defined in the script yields an instance of this class. 

Some important members and methods, which are recongnised from `Script::` and have the same meaning except they are now limited to their specific namespace,

* `mVar` and `mLazyVar`, each namespace contains its own set of vars. 
    * `FindVar()` and `FindOrAddVar()`, searches in invoking `NameSpace` instance. These return local variables if there is a _current func_.
* `mFuncs`, each namespace has its own set of functions.
    * `FindFunc()`, finds a func in the namespace.
* `mFirstLine`, `mLastLine`, `mFirstStaticLine` and `mLastStaticLine`, used to control auto-exec section and static var initialisation and preprocessing at load time.

Note that `NameSpace::FindXXX` calls (slightly modified) `Script::FindXXX`, this should be improved eventually.

Other important members and methods,

* `mSettings` ( `global_stuct`),  each namespace has its own _script settings_, that is most of the A_ variables.
* `mNestedNamespaces` (`class NameSpaceList`), a list of the namespace' nested namespaces. Get with `GetNestedNameSpaces()`.
* `SetCurrentNameSpace()`, used to change _this_ namespace to be the current one. More on this below.
* `mName`, get the namespace name with `GetName()`

__Labels__, are namespace specific, but namespaces do not store their own labels, these are still stored via `Script::mFirstLabel`. Labels instead have a new member, `mNameSpace`, defining which namespace the label belongs to and allows duplicate labels across namespaces. `Goto` and `gosub` can not jump to a label in another namespace, this is intended. 

Some important global variables:

* `NameSpace *g_CurrentNameSpace`. Always refers to the _current namespace_ which most often means the namespace which owns the current executing or loading line (`Line`). At some stages during loading, there is no natural candidate for which namespace this should be. 
* `NameSpace *g_DefaultNameSpace`. Refers to the _Script_ namespace, in which the entire script is defined. The script's namespaces are all nested namespaces of the default namespace.
* `NameSpace *g_StandardNameSpace`. The _standard namespace_, can be used to access overridden BIFs. This namespace is parallel to the _default namespace_, and accessible everywhere via `std -> ...`.

# Threads and settings

*A_* settings now belong to the namespaces and not threads. This means that any change to an *A_* setting takes effect immediately for all threads, which runs in the relevant namespace. 

## `struct global_struct` and `struct ScriptThread`

The struct `global_struct` has been split into two structs. `global_struct` now only contains settings which belongs to namespaces, such as _WinDelay_, _DetectHiddenWindows_ and most (perhaps all) other settings which are set via A_ variables. Whereas `ScriptThread` contains thread specific settings only, eg, _CurrentFunc_. 

Important global variables:

* `g` (`global_struct`) is kept, but is no longer an array, and only refers to the relevant settings of the current namespace.
* `t` (`ScriptThread`), this is the new _thread array_, its base is defined by `::t0` (prev: `g_array`). Its size is defined by `g_MaxThreadsTotal` and its _position_ is defined by `g_nThreads`. Default settings are defined in `::t_default`.

_Script properties_ are unaffected, at present.

## Setting the current namespace

To set the current namespace, `g_CurrentNameSpace` should not be assigned to directly. But instead by calling `NameSpace::SetCurrentNameSpace()`. This also updates `::g`  to refer to the relevant `mSettings`.

At run time, the namespace is changed when _jumping_ to a line, when calling a built-in function and when getting or setting a BIV. Typically, when setting the namespace at run time, one must save and restore the previous namespace. 

Some classes have a new instance member which indicates which namespace they belong to. It is typically called `mNameSpace` (`NameSpace*`). This is used for various purposes, most importantly is to set the correct namespace, but also for debugging, for example, `Line::mNameSpace` is used by `listlines` to show when _entering_ a new namespace.

Changing namespace is very low cost and doesn't affect performance.

## Scope resolution operator

The scope operator is implemented as a short-circuit operator. The following `SymbolType`s has been added

* `SYM_NAMESPACE`, indicates that `ExprTokenType::resolved_namespace` is used. This new (union) member holds a pointer to a namespace which has either been resolved by the expression evaluator or by load time optimisations.
* `SYM_SCOPE`, means the `->`. In expression evaluation, `SYM_SCOPE` requires a `SYM_NAMESPACE` from the stack.
* `SYM_SCOPE_BEGIN` , is the first `->` in a series of scope resolutions. The beginning is marked to indicate that this scope op does not consume a `SYM_NAMESPACE` from the stack. This allows nested scope ops, eg, `a -> ( b -> c ) -> d`. Where the second `->` cannot consume the `SYM_NAMESPACE` resolved by the first one.
* `SYM_UNRESOLVED_NAMESPACE_VAR` and `SYM_UNRESOLVED_NAMESPACE_DYNAMIC_VAR`, these are the ending variables in a scope resolution series, that is, `b` in `a -> b` or `%b%` in `a -> %b%`.
* `SYM_UNRESOLVED_NAMESPACE_FUNC` and `SYM_UNRESOLVED_NAMESPACE_DYNAMIC_FUNC`, these are the ending functions in a scope resolution series, that is, `b()` in `a -> b()` or `%b%()` in `a -> %b%()`.

Load time optimisations resolves scope resolutions as far a possible, if completely non-dynamic, the final variable or func reference is resolved and there is no performance penalty for using the scope op vs a direct variable or func reference. If a non-dynamic reference to a non-existent func or namespace is made, a load time error occurs. A notable exception is for a optional namespace, example,
```autohotkey
#import *i a.ahk -> b
b -> f() ; if a.ahk couldn't be loaded, this line will only yield a runtime error if executed due to *i.
```
this will present some issues with the compiler.

Important functions relating to the scope operator:
* `Line::ProcessScopeResolution()`, called in `ExpressionToPostfix`. Calls:
    * `FindScopeBegin()`, converts relevant `SYM_SCOPE` to `SYM_SCOPE_BEGIN`, handles the `new` operator and does some syntax verification.
    * `ResolveScopeOperands()`, load time optimisations, resolves namespaces and / or variables and func references from scope resolution series.

## `HotCriterionAllowsFiring()`

This function passes a namespace' `mSettings` to `WinExist` and `WinActive` which I believe, strictly speaking, isn't thread safe, since the _script thread_ can write to `mSettings` at any time while the _hook thread_ reads from it, and reading and writing concurrently is undefined behaviour. However, it seems that the current implementation, which passes `g_default` suffers from this potential unsafe issue too, since `g_default` is written to in `AutoExecSectionTimeout()` and `Script::AutoExecSection()` when hotkeys are active.

Disregarding undefined behaviour on the c++ side, this can potentially cause unexpected behaviour on the script side. An example is a script which uses `settitlematchmode` and `detecthiddenwindows` in two combinations:

```autohotkey 
setting1:
settitlematchmode 1
detecthiddenwindows true
; or
setting2:
settitlematchmode 2
detecthiddenwindows false
``` 

Let's say the script wants to set `setting1`, when `setting2` is active, then it is possible for the hook thread to read the current setting between the two lines, so that the situations becomes,

```autohotkey
settitlematchmode 1
; hook thread comes in here
detecthiddenwindows false
```

## Blocks

Namespace blocks cannot be enclosed in any other block than another namespace block. For consistency, and sanity, class bodies has now also been prohibited from being defined in any block. Example,

```autohotkey
if x {
	namespace { ; or class ... yeilds a load time error
	}
}
```
However, currently this is tolerated, and works,
```autohotkey
if x
	namespace { ; this namespace or any of its lines has nothing to do with the above "if x"
	}
	msgbox ; this line belongs to the above "if x"
```
The same goes for `#import`. No changes has been made to function definitions in this regard.

## Remappings

Remappings now save and restore the relevant keydelay, I think this works but is meant to be a temporary solution, example,

```autothokey
a::lbutton
; Yields:
*a::
std->remap_delay := A_MouseDelay, A_MouseDelay := -1, !GetKeyState("lbutton") ? Send("{Blind}{lbutton DownR}") : '', A_MouseDelay := std->remap_delay
return
*a up::
std->remap_delay := A_MouseDelay, A_MouseDelay := -1, Send("{Blind}{lbutton Up}"), A_MouseDelay := std->remap_delay
return
```

A better solution would not make a code replacement but instead let an internal function _carry out_ the remapping, this could also allow a built in function to enable scripts to easily enable / disable remappings at run time, eg, `Remap('a','b', onoff)`. A better code replacement solution would be easy if settings were function parameters, eg , `send(str, keydelay, pressdur)`. 

## Debugger

Added context name _Namespace_, `PropertyContextType::PC_Namespace`. `context_get -c 2 -d x` now returns the name of the namespace at the given stack depth (`x`) and all names of its enclosing namespaces. Example,
```text
context_get -c 2 -d 0
<?xml version="1.0" encoding="UTF-8"?><response command="context_get" context="2" transaction_id=""><property name="n3"></property><property name="n2"></property><property name="n1"></property><property name="Script"></property></response>
```
from:
```autohotkey
namespace n1 {
	namespace n2 {
		namespace n3 {
			msgbox 'context_get here'
		}
	}
}
```
`proptery_get` and `property_set` can resolve variables with the scope operator. Example,
```
property_get -n "n1 -> n2 -> n3 -> abc"
<?xml version="1.0" encoding="UTF-8"?><response command="property_get" transaction_id=""><property name="n1 -&gt; n2 -&gt; n3 -&gt; abc" fullname="n1 -&gt; n2 -&gt; n3 -&gt; abc" type="integer" facet="" children="0" encoding="base64" size="3">MTIz</property></response>
-
```
from:
```autohotkey
namespace n1 {
	namespace n2 {
		namespace n3 {
			abc := 123
		}
	}
}
msgbox 'property_get here'
```
This also works with the dot and indexing operators.
Note that the scope of the var is resolved relative to the context of the relevant stack depth, to get a specific var independent of this context, start the scope resolution with `Script -> ...`.

### Stack depth context

At some stack depths, the context is not what might be expected, example,
```
namespace n1 {
	abc := 123
}
x::
	msgbox 'break'
return
}
```
If the __x thread__ interrupts the __Auto-execute thread__ of `namespace n1`, the namespace context of that stack item is still __n1__, but the following stack item, __x sub__, will be in the __Script__ namespace. This is due to pushing onto the stack before the namespace has changed. See `DbgStack::current_namespace` for some additional comments. In such cases there should always be a following stack item with the correct context.

# How to continue

If there is no desire to have this, for whatever reason, _at least I tried_ 😄 . If there is an interest, I could finish the work according to some agreement. I would also be very happy to leave the remaining development of this in the hands of someone else, eg, __Lexikos__. Then I'd continue to contribute via PRs _as usual_. 

Cheers ☕️

### DIFF
--- a/AutoHotkeyx.vcxproj
+++ b/AutoHotkeyx.vcxproj
@@ -223,6 +223,7 @@
     <ClCompile Include="source\hotkey.cpp" />
     <ClCompile Include="source\keyboard_mouse.cpp" />
     <ClCompile Include="source\mt19937ar-cok.cpp" />
+    <ClCompile Include="source\NameSpace.cpp" />
     <ClCompile Include="source\os_version.cpp" />
     <ClCompile Include="source\script.cpp" />
     <ClCompile Include="source\script2.cpp" />
@@ -260,6 +261,8 @@
     <ClInclude Include="source\KuString.h" />
     <ClInclude Include="source\lib_pcre\pcre\pcret.h" />
     <ClInclude Include="source\mt19937ar-cok.h" />
+    <ClInclude Include="source\NameSpace.h" />
+    <ClInclude Include="source\NameSpaceDefines.h" />
     <ClInclude Include="source\os_version.h" />
     <ClInclude Include="source\lib_pcre\pcre\pcre.h" />
     <ClInclude Include="source\qmath.h" />
--- a/AutoHotkeyx.vcxproj.filters
+++ b/AutoHotkeyx.vcxproj.filters
@@ -103,6 +103,9 @@
     <ClCompile Include="source\script_com.cpp">
       <Filter>Source Files</Filter>
     </ClCompile>
+    <ClCompile Include="source\NameSpace.cpp">
+      <Filter>Source Files</Filter>
+    </ClCompile>
   </ItemGroup>
   <ItemGroup>
     <ClInclude Include="source\application.h">
@@ -193,6 +196,12 @@
       <Filter>Header Files</Filter>
     </ClInclude>
     <ClInclude Include="source\script_func_impl.h">
+      <Filter>Header Files</Filter>
+    </ClInclude>
+    <ClInclude Include="source\NameSpace.h">
+      <Filter>Header Files</Filter>
+    </ClInclude>
+    <ClInclude Include="source\NameSpaceDefines.h">
       <Filter>Header Files</Filter>
     </ClInclude>
   </ItemGroup>
--- a/source/AutoHotkey.cpp
+++ b/source/AutoHotkey.cpp
@@ -141,39 +141,19 @@ int WINAPI _tWinMain (HINSTANCE hInstance, HINSTANCE hPrevInstance, LPTSTR lpCmd
 			break; // No more switches allowed after this point.
 		}
 	}
-
-	if (Var *var = g_script.FindOrAddVar(_T("A_Args"), 6, VAR_DECLARE_SUPER_GLOBAL))
-	{
-		// Store the remaining args in an array and assign it to "Args".
-		// If there are no args, assign an empty array so that A_Args[1]
-		// and A_Args.MaxIndex() don't cause an error.
-		Object *args = Object::CreateFromArgV(__targv + i, __argc - i);
-		if (!args)
-			return CRITICAL_ERROR;  // Realistically should never happen.
-		var->AssignSkipAddRef(args);
-	}
-	else
+	
+	// Create A_Args and ErrorLevel:
+	if (!g_script.CreateArgsArrayAndErrorLevel(__targv + i, __argc - i))
 		return CRITICAL_ERROR;
 
-	global_init(*g);  // Set defaults.
 
 // Set up the basics of the script:
 #ifdef AUTOHOTKEYSC
-	if (g_script.Init(*g, _T(""), restart_mode) != OK) 
+	if (g_script.Init(_T(""), restart_mode) != OK) 
 #else
-	if (g_script.Init(*g, script_filespec, restart_mode) != OK)  // Set up the basics of the script, using the above.
+	if (g_script.Init(script_filespec, restart_mode) != OK)  // Set up the basics of the script, using the above.
 #endif
 		return CRITICAL_ERROR;
-
-	// Set g_default now, reflecting any changes made to "g" above, in case AutoExecSection(), below,
-	// never returns, perhaps because it contains an infinite loop (intentional or not):
-	CopyMemory(&g_default, g, sizeof(global_struct));
-
-	// Use FindOrAdd vs Add for maintainability, although it shouldn't already exist:
-	if (   !(g_ErrorLevel = g_script.FindOrAddVar(_T("ErrorLevel")))   )
-		return CRITICAL_ERROR; // Error.  Above already displayed it for us.
-	// Initialize the var state to zero:
-	g_ErrorLevel->Assign(ERRORLEVEL_NONE);
 
 	// Could use CreateMutex() but that seems pointless because we have to discover the
 	// hWnd of the existing process so that we can close or restart it, so we would have
@@ -184,7 +164,8 @@ int WINAPI _tWinMain (HINSTANCE hInstance, HINSTANCE hPrevInstance, LPTSTR lpCmd
 	// instance terminates, so it should work ok:
 	//CreateMutex(NULL, FALSE, script_filespec); // script_filespec seems a good choice for uniqueness.
 	//if (!g_ForceLaunch && !restart_mode && GetLastError() == ERROR_ALREADY_EXISTS)
-
+	
+	ASSERT(g_CurrentNameSpace == g_DefaultNameSpace); // Make sure that the default namespace really becomes the default namespace.
 	UINT load_result = g_script.LoadFromFile();
 	if (load_result == LOADING_FAILED) // Error during load (was already displayed by the function call).
 	{
@@ -256,6 +237,9 @@ int WINAPI _tWinMain (HINSTANCE hInstance, HINSTANCE hPrevInstance, LPTSTR lpCmd
 	// because otherwise the change to the "focus stealing" setting would never be undone:
 	SetForegroundLockTimeout();
 
+	// Before creating our windows and the tray icon, initilize the ScriptThread array since the MainWindowProc relies on it existing.
+	if (!InitScriptThreads())
+		return CRITICAL_ERROR;
 	// Create all our windows and the tray icon.  This is done after all other chances
 	// to return early due to an error have passed, above.
 	if (g_script.CreateWindows() != OK)
@@ -320,9 +304,7 @@ int WINAPI _tWinMain (HINSTANCE hInstance, HINSTANCE hPrevInstance, LPTSTR lpCmd
 	g_HSSameLineAction = false; // `#Hotstring X` should not affect Hotstring().
 	g_SuspendExempt = false; // #SuspendExempt should not affect Hotkey()/Hotstring().
 
-	Var *clipboard_var = g_script.FindOrAddVar(_T("Clipboard")); // Add it if it doesn't exist, in case the script accesses "Clipboard" via a dynamic variable.
-
-	// Run the auto-execute part at the top of the script (this call might never return):
+	// Run the auto-execute part at the top of each namespace (this call might never return):
 	if (!g_script.AutoExecSection()) // Can't run script at all. Due to rarity, just abort.
 		return CRITICAL_ERROR;
 	// REMEMBER: The call above will never return if one of the following happens:
--- a/source/Debugger.cpp
+++ b/source/Debugger.cpp
@@ -19,6 +19,7 @@ freely, without restriction.
 #include "script_object.h"
 #include "script_com.h"
 #include "TextIO.h"
+#include "NameSpace.h"
 //#include "Debugger.h" // included by globaldata.h
 
 #ifdef CONFIG_DEBUGGER
@@ -637,37 +638,52 @@ DEBUGGER_COMMAND(Debugger::breakpoint_set)
 	// Due to the introduction of expressions in static initializers, lines aren't necessarily in
 	// line number order.  First determine if any static initializers match the requested lineno.
 	// If not, use the first non-static line at or following that line number.
-
-	if (g_script.mFirstStaticLine)
-		for (line = g_script.mFirstStaticLine; ; line = line->mNextLine)
-		{
-			if (line->mFileIndex == file_index && line->mLineNumber == lineno) // Exact match, unlike normal lines.
+	
+	FOR_EACH_NAMESPACE(ns)
+	{
+		if (ns->mFirstStaticLine)
+			for (line = ns->mFirstStaticLine; ; line = line->mNextLine)
 			{
-				found_line = line;
-				break;
-			}
-			if (line == g_script.mLastStaticLine)
-				break;
-		}
-	if (!found_line)
-		// If line is non-NULL, above has left it set to mLastStaticLine, which we want to exclude:
-		for (line = line ? line->mNextLine : g_script.mFirstLine; line; line = line->mNextLine)
-			if (line->mFileIndex == file_index && line->mLineNumber >= lineno)
-			{
-				// ACT_ELSE and ACT_BLOCK_BEGIN generally don't cause PreExecLine() to be called,
-				// so any breakpoint set on one of those lines would never be hit.  Attempting to
-				// set a breakpoint on one of these should act like setting a breakpoint on a line
-				// which contains no code: put the breakpoint at the next line instead.
-				// Without this check, setting a breakpoint on a line like "else Exit" would not work.
-				if (line->mActionType == ACT_ELSE || line->mActionType == ACT_BLOCK_BEGIN)
-					continue;
-				// Use the first line of code at or after lineno, like Visual Studio.
-				// To display the breakpoint correctly, an IDE should use breakpoint_get.
-				if (!found_line || found_line->mLineNumber > line->mLineNumber)
+				if (line->mFileIndex == file_index && line->mLineNumber == lineno) // Exact match, unlike normal lines.
+				{
 					found_line = line;
-				// Must keep searching, since class var initializers can cause lines to be listed out of order.
-				//break;
+					goto break_twice;	// leave the outer loop too
+				}
+				if (line == ns->mLastStaticLine)
+					break;	// continue FOR_EACH_NAMESPACE
 			}
+	}
+	break_twice:
+	if (!found_line)
+	{
+		FOR_EACH_NAMESPACE(ns)
+		{
+			line = ns->mLastStaticLine; // can be null, but must be excluded
+			for (line = line ? line->mNextLine : ns->mFirstLine; line; line = line->mNextLine)
+			{
+				if (line->mFileIndex == file_index && line->mLineNumber >= lineno)
+				{
+					// ACT_ELSE and ACT_BLOCK_BEGIN generally don't cause PreExecLine() to be called,
+					// so any breakpoint set on one of those lines would never be hit.  Attempting to
+					// set a breakpoint on one of these should act like setting a breakpoint on a line
+					// which contains no code: put the breakpoint at the next line instead.
+					// Without this check, setting a breakpoint on a line like "else Exit" would not work.
+					if (line->mActionType == ACT_ELSE || line->mActionType == ACT_BLOCK_BEGIN)
+						continue;
+					// Use the first line of code at or after lineno, like Visual Studio.
+					// To display the breakpoint correctly, an IDE should use breakpoint_get.
+					if (!found_line || found_line->mLineNumber > line->mLineNumber)
+						found_line = line;
+					// Must keep searching, since class var initializers can cause lines to be listed out of order.				
+					//break;
+				}
+			}
+			// Must continue searching since another namespace might provide a better match. But if...
+			if (found_line 
+				&& found_line->mLineNumber == lineno) // ...an exact match was found, it is safe to break.
+				break; // FOR_EACH_NAMESPACE(ns)
+		}
+	}
 	if (found_line)
 	{
 		if (!found_line->mBreakpoint)
@@ -700,18 +716,20 @@ DEBUGGER_COMMAND(Debugger::breakpoint_get)
 	int breakpoint_id = atoi(ArgValue(aArgV, 0));
 
 	Line *line;
-	for (line = g_script.mFirstLine; line; line = line->mNextLine)
+	FOR_EACH_NAMESPACE(ns)
 	{
-		if (line->mBreakpoint && line->mBreakpoint->id == breakpoint_id)
+		for (line = ns->mFirstLine; line; line = line->mNextLine)
 		{
-			mResponseBuf.WriteF("<response command=\"breakpoint_get\" transaction_id=\"%e\">", aTransactionId);
-			WriteBreakpointXml(line->mBreakpoint, line);
-			mResponseBuf.Write("</response>");
+			if (line->mBreakpoint && line->mBreakpoint->id == breakpoint_id)
+			{
+				mResponseBuf.WriteF("<response command=\"breakpoint_get\" transaction_id=\"%e\">", aTransactionId);
+				WriteBreakpointXml(line->mBreakpoint, line);
+				mResponseBuf.Write("</response>");
 
-			return DEBUGGER_E_OK;
+				return DEBUGGER_E_OK;
+			}
 		}
 	}
-
 	return DEBUGGER_E_BREAKPOINT_NOT_FOUND;
 }
 
@@ -760,43 +778,45 @@ DEBUGGER_COMMAND(Debugger::breakpoint_update)
 		return DEBUGGER_E_INVALID_OPTIONS;
 
 	Line *line;
-	for (line = g_script.mFirstLine; line; line = line->mNextLine)
+	FOR_EACH_NAMESPACE(ns)
 	{
-		Breakpoint *bp = line->mBreakpoint;
-
-		if (bp && bp->id == breakpoint_id)
+		for (line = ns->mFirstLine; line; line = line->mNextLine)
 		{
-			if (lineno && line->mLineNumber != lineno)
-			{
-				// Move the breakpoint within its current file.
-				int file_index = line->mFileIndex;
-				Line *old_line = line;
+			Breakpoint *bp = line->mBreakpoint;
 
-				for (line = g_script.mFirstLine; line; line = line->mNextLine)
+			if (bp && bp->id == breakpoint_id)
+			{
+				if (lineno && line->mLineNumber != lineno)
 				{
-					if (line->mFileIndex == file_index && line->mLineNumber >= lineno)
+					// Move the breakpoint within its current file.
+					int file_index = line->mFileIndex;
+					Line *old_line = line;
+
+					for (line = ns->mFirstLine; line; line = line->mNextLine)
 					{
-						line->mBreakpoint = bp;
-						break;
+						if (line->mFileIndex == file_index && line->mLineNumber >= lineno)
+						{
+							line->mBreakpoint = bp;
+							break;
+						}
 					}
+
+					// If line is NULL, the line was not found.
+					if (!line)
+						return DEBUGGER_E_BREAKPOINT_INVALID;
+
+					// Seems best to only remove the breakpoint from its previous line
+					// once we know the breakpoint_update has succeeded.
+					old_line->mBreakpoint = NULL;
 				}
 
-				// If line is NULL, the line was not found.
-				if (!line)
-					return DEBUGGER_E_BREAKPOINT_INVALID;
+				if (state != -1)
+					bp->state = state;
 
-				// Seems best to only remove the breakpoint from its previous line
-				// once we know the breakpoint_update has succeeded.
-				old_line->mBreakpoint = NULL;
+				return DEBUGGER_E_OK;
 			}
-
-			if (state != -1)
-				bp->state = state;
-
-			return DEBUGGER_E_OK;
 		}
 	}
-
 	return DEBUGGER_E_BREAKPOINT_NOT_FOUND;
 }
 
@@ -809,17 +829,19 @@ DEBUGGER_COMMAND(Debugger::breakpoint_remove)
 	int breakpoint_id = atoi(ArgValue(aArgV, 0));
 
 	Line *line;
-	for (line = g_script.mFirstLine; line; line = line->mNextLine)
+	FOR_EACH_NAMESPACE(ns)
 	{
-		if (line->mBreakpoint && line->mBreakpoint->id == breakpoint_id)
+		for (line = ns->mFirstLine; line; line = line->mNextLine)
 		{
-			delete line->mBreakpoint;
-			line->mBreakpoint = NULL;
+			if (line->mBreakpoint && line->mBreakpoint->id == breakpoint_id)
+			{
+				delete line->mBreakpoint;
+				line->mBreakpoint = NULL;
 
-			return DEBUGGER_E_OK;
+				return DEBUGGER_E_OK;
+			}
 		}
 	}
-
 	return DEBUGGER_E_BREAKPOINT_NOT_FOUND;
 }
 
@@ -831,14 +853,16 @@ DEBUGGER_COMMAND(Debugger::breakpoint_list)
 	mResponseBuf.WriteF("<response command=\"breakpoint_list\" transaction_id=\"%e\">", aTransactionId);
 	
 	Line *line;
-	for (line = g_script.mFirstLine; line; line = line->mNextLine)
+	FOR_EACH_NAMESPACE(ns)
 	{
-		if (line->mBreakpoint)
+		for (line = ns->mFirstLine; line; line = line->mNextLine)
 		{
-			WriteBreakpointXml(line->mBreakpoint, line);
+			if (line->mBreakpoint)
+			{
+				WriteBreakpointXml(line->mBreakpoint, line);
+			}
 		}
 	}
-
 	return mResponseBuf.Write("</response>");
 }
 
@@ -929,7 +953,7 @@ DEBUGGER_COMMAND(Debugger::context_names)
 		return DEBUGGER_E_INVALID_OPTIONS;
 
 	return mResponseBuf.WriteF(
-		"<response command=\"context_names\" transaction_id=\"%e\"><context name=\"Local\" id=\"0\"/><context name=\"Global\" id=\"1\"/></response>"
+		"<response command=\"context_names\" transaction_id=\"%e\"><context name=\"Local\" id=\"0\"/><context name=\"Global\" id=\"1\"/><context name=\"Namespace\" id=\"2\"/></response>"
 		, aTransactionId);
 }
 
@@ -956,10 +980,9 @@ DEBUGGER_COMMAND(Debugger::context_get)
 			return DEBUGGER_E_INVALID_OPTIONS;
 		}
 	}
-
+	
 	Var **var = NULL, **var_end = NULL; // An array of pointers-to-var.
 	VarBkp *bkp = NULL, *bkp_end = NULL;
-	
 	// TODO: Include the lazy-var arrays for completeness. Low priority since lazy-var arrays are used only for 10001+ variables, and most conventional debugger interfaces would generally not be useful with that many variables.
 	if (context_id == PC_Local)
 	{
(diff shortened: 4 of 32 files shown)


nnnik:
Wow awesome work.
A mechanism for that definitively is a necesity going onwards with AHK.

I have read the definitions (although not too deeply) and there are a few things I think still might need changing/tweaking.

Generally I don't like the -> operator in the slightest. It is uneccessary clutter. We already have an operator for accessing objects that operator will work for namespaces too.
(Unless you want to allow colliding function names, )
It might be a good idea to remove func(funcName) or %varOrClass% and similar in favor of:
namespace.%funcName% or nameSpace[varOrClass].

I kind of dislike how you differentiate between libraries and all other namespaces creating artificial barriers that probably lead to a restricted behavior later on.

I look forward to future developments and can't wait to see what you come up with and what we end up with finally.
Now that you took the first step I cannot see this not being part of AHKs future :)

HelgeffegleH:
Hello __nnnik__ 👋. 

The scope resolution operator (`->`), allows separation of namespace names and variable names. Without it there is ambiguity, for example, does `a.b.c` resolve the variable `c` in the namespace `b` which is nested in namespace `a`, or is it object access, and if so, in which object? With the scope op, it is clear what it means, for example  `a->b->c` resolves the variable `c` while `a->b.c` resolves the variable `b` and gets property `c`. Using a single character symbol would reduce _clutter_, I thought about `#` but I got used to `->` pretty fast and kind of like it now 😄.

Regarding _libraries_, this feature will probably be removed. I think the reserved name `outer` will suffice in most cases. Perhaps something more generally useful could be achieved with something like a `#use` directive, although I didn't contemplate this yet.

Thanks for your feedback and optimism, cheers.

nnnik:
Yeah thats exactly the thing - imo variable and namespace shouldn't be seperate "scopes".
If there is a namespace definition and a class definition in the same namespace throw an error.

HelgeffegleH:
I see, it could work. We could let a _property get_ on a namespace name get a nested namespace. _Item get_ can resolve a variable. A _method call_, calls the function. So, eg, `a.b['c']` would be what is now `a->b->c`. While the former is, at present, more recognisable than the latter, I do not see any real benefit. I can think of the following drawbacks,

* The choice of names for namespaces and variable gets more limited.
* For each new variable, we have to check for a name clash.
* There is no visual distinction between _scope resolution_ and _object access_. This makes code less readable imo. 
    * For example,  if  `a['b']` is resolved to a variable reference, it can be passed _byref_ to a function, but if it does _item get_ on an object, the result cannot be passed byref.
    * A similar inconsistency would arise with the address-of operator (`&`).
* _Clutter_, if you like, `a['b']` requires more characters than `a->b`, even more so if we would reduce the scope resolution operator to a one character symbol.


Cheers.

nnnik:
> The choice of names for namespaces and variable gets more limited.

On the other hand allowing variables functions and classes to have the same name can increase confusion.
With the introduction of namespaces we also have essentially unlimited space for new names.
I personally think that we should flatten the language and make sure one name only refers to one thing.
Regardless of whether we finally end up using -> or . it would be better if we prevent this from happening:
A namespace shouldnt have the same name as a class variable or function.

> For each new variable, we have to check for a name clash.
We already do the same with classes and the class overwrite warning - whats the difference here?

> There is no visual distinction between scope resolution and object access. This makes code less readable imo.
Is there a need for visual distinction if they are almost functionally identical.

> For example, if a['b'] is resolved to a variable reference, it can be passed byref to a function, but if it does item get on an object, the result cannot be passed byref.
The issue that objects cannot be passed byref is not something that can be retained in the language for future releases. We either need to remove byref completely or make it compatible with object syntax.

> A similar inconsistency would arise with the address-of operator (&).
I didn't think of that. Why do we need to retain the adress of operator in the first place?
In AutoHotky v1 this was required to give users access to basic features - such as information about object types and binary data.
With AutoHotkey v2 we can actually decide to give our users valid intended features that take care of that. The buffer object is one such feature - more of that might be needed.
This would also remove all the inconsistencies this operator has caused.
Not every data sink or source might have a pointer associated with them that delivers valid reasonable information. That assumption is just plain wrong. Even worse it promotes the use and abuse of internal structures - even though they are subject to change.
We could however replace the operator with valid alternatives.

HelgeffegleH:
> On the other hand allowing variables functions and classes to have the same name can increase confusion.

Sure, but with a scope operator it is always clear what is a namespace and what is variable, so there is no confusion regardless.

> We already do the same with classes and the class overwrite warning - whats the difference here?

Class overwrite is only checked once, at load time, this greatly limits the number of checks needed. All variables created a runtime would need to be checked and the language is supposed to be _designed to support at least several million variables without a significant drop in performance_. With future improvements, such as hashing, this wouldn't be a problem though. And ofc, no one should make a million variables anyways.

Overall, fair points __nnnik__ :thumbsup:, although a bit outside the _scope_ of this PR, but ofc, if there are future changes to the program which this PR needs to be adapted to, then, so be it.

Cheers.

Lexikos:
This deserves more thought than I'm willing to put into it right now.

> these kinds of settings are more suited for limited scopes rather than a thread which can call any function in any library. 

I can see the sense in that, but I disagree with tying this to "namespaces". For any given hotkey or other thread procedure which is designed to have different settings than the default and/or other procedures, would you propose that it be enclosed in a namespace? Is that permitted for hotkeys? A namespace would hypothetically allowing grouping procedures which share common settings, but also imposes requirements on accessing those procedures.

It seems that you are conflating namespaces and modules.

HelgeffegleH:
> It seems that you are conflating namespaces and modules.

Yes. A `Namespace {...}` is more than one _namespace_, it has its own function/var/label namespaces, its own settings, hotkeys, hotstrings, and its own auto exec section. As such, it could be called a _module_, I chose `Namespace {...}` because the word _module_, to me, implies a _self-containment_ which is possible but not required for a `Namespace {...}`. The way I imagined things, is that one would use `#import` for what is intended to be a _module_, i.e., something _self-contained_ (but which could be used by something external). And `Namespace {...}` would be used for things which you want to isolate and / or group, but which isn't necessarily self-contained or doesn't have any purpose on its own.

> I disagree with tying this to "namespaces".

I agree with your disagreement, in isolation, that is, it is not really _sane_ to say that settings belongs to a _namespace_. However, as AHK settings _are_ tightly coupled with variables, I think it is quite natural that these variables (and hence settings) _belong_ to a variable namespace. Threads doesn't have a variable namespace. Functions do, so we could let settings belong to functions, but we'd also need to consider code outside functions. Anyways, I think the idea of setting _defaults_ for functions is flawed to begin with. Evidently we have ended up with constructs such as,

```autohotkey
a_x := y
bif()
```
and 

```autohotkey
px := a_x
a_x := y
bif()
a_x := px
```
and worse, when we really want do
```autohotkey
bif(y) 
; or
bif()
```
___
>  For any given hotkey or other thread procedure which is designed to have different settings than the default and/or other procedures, would you propose that it be enclosed in a namespace?

For thread callbacks, with the current behaviour, consider this example,
```autohotkey
x::
	a_detecthiddenwindows := true
	; ...
return
```
This might be perceived as convenient, but here we have put the responsibility for setting / restoring settings on every function / library we call, and restoration could be required at multiple possible exit points, both for returns and in case of exceptions, so there is a lot of inconvenience implied before we arrive at this _convenience_. And this inconvenience is implied even if the thread callback doesn't change any settings, because the functions and libraries cannot (always) know which thread they'll run in.

With the new design we will have to do it a bit differently. There are several options, which one you do would depend on preference and circumstances. Let's take this again,

```autohotkey
x::
	a_detecthiddenwindows := true
	; ...
return
```

If we know that the thread finishes quickly, or won't be interrupted by anything which relies on, or changes the settings, then we could just leave it as is, but then each procedure (in that namespace) has to set the setting if it relies on it, or we can restore a default value or save and restore the current value,  eg,

```autohotkey
; auto exec:
dhw_default := false
; ...
x::
	a_detecthiddenwindows := true
	; ...
	a_detecthiddenwindows := dhw_default
return
```

We can also use `critical` to ensure there are no interruptions. As described in the documentation, `critical` was changed, it is now both a _namespace_ and thread setting. That is, _namespaces_ can set a default for the threads which they _spawn_, and a thread can set its critical setting to apply only for the remainder of its existence (or until changed again).

Further, we can enclose a (group of) procedure(s) in a separate namespace, eg,

```autohotkey
namespace
{
	a_detecthiddenwindows := true
	x::
		; ...
	return
}
```

The benefit here is that we only need to set the setting once, but as mentioned, it imposes other requirements. For example, if we want to use anything outside the namespace, we'd need to prefix the names with `outer->`, eg `outer -> f()`. 

Another way is to have a namespace with just one (set of) setting(s), or an _empty_ namespace which implies the default settings, for example,

```autohotkey
namespace dhw_on
{
	a_detecthiddenwindows := true
}
namespace default
{
}
x::
	; ...
	dhw_on -> sendmessage(...)
	default -> winexist(...)
	; ...
return
```

Ofc, while this construct is easy to use, performs well and is safe, it really shows the desire to set the settings via function parameters. This construct is less suitable for numeric settings, such as `a_keydelay` or other more _elaborate_ setting combinations. 

We can also define custom functions which accepts parameters for the settings. The ability to override and access bifs makes this more convenient in this branch. Example,

```autohotkey
winexist(x, dhw := false) {
	local
	old_dhw := a_detecthiddenwindows
	, a_detecthiddenwindows := dhw
	, hwnd := std -> winexist(x)
	, a_detecthiddenwindows := old_dhw
	return hwnd
}
```

### Possible improvements

* Make all settings function parameters, with _A_Setting_ as the default, eg, `SendEvent(str [, KeyDelay := A_KeyDelay, PressDuration := A_PressDuration])` (`sendlevel` too I guess). Then we have the possibility to set defaults within _namespaces_, which I suppose can be convenient. Then we can safely and easily use many different settings when we cannot commit to one particular set of settings. 

* Make all settings parameters only. 

I think one of the above are necessary, regardless of whether we will use something like this PR or not.

___

> Is that permitted for hotkeys?

Yes, a namespace can contain hotkeys. Duplicate hotkeys are disallowed, but duplicate variants are permitted across namespaces, because the `#if`s belongs to the namespaces. For example, neither

```autohotkey
#if exprA
x::return
#if exprA
x::return
```
nor
```autohotkey
namespace
{
	x::return
}
x::return
```
are permitted, but this,
```autohotkey
namespace
{
	#if exprA
	x::return
}
namespace
{
	#if exprA
	x::return
}
x::return ; not directly affected by the above #ifs
#if exprA
x::return
```
is allowed. Here, only the first variant which is _true_ will trigger, they are evaluated from the top and down.

___

### Example

As an example, I rewrote a small library for  testing / demo: ([old](https://github.com/HelgeffegleH/buf/blob/master/buf.ahk) / [new](https://github.com/HelgeffegleH/buf/blob/buf_ns/buf.ahk)). 

In the old version, there are six functions in the global namespace, named with a hardcoded prefix: `buf_`. In the new version, the user can use `#import` to specify whichever _prefix_ it likes, eg `buf` (`buf->xyz()`). The new version has a nested namespace which does all of the library initialisation tasks and also allows access to the function pointers, while isolating all _temporary_ vars and functions from the end user if it prefers to use `#include`. The old version uses an awkward static initialisation method with _position dependency_, and lots of code repeating.

___

Cheers.

Lexikos:
This is sounding much more complicated than a *namespace* or even module should be.

> `#if`s belongs to the namespaces

Why is that?

# Semantics

> module, to me, implies a self-containment which is possible but not required for a Namespace {...}. 

I don't think I understand your point of view, let alone agree. It seems to be something that *can be* a module, or whatever you'd call a module that isn't self-contained (I'd call it "a module that isn't self-contained"), but either way, it is definitely not just a namespace. A namespace should do no more than define a scope for names.

A module may be expected to establish a scope for the names it contains (i.e. a namespace), but also other aspects that may or may not be used. Modules are not *required* to be self-contained in any language I've ever used.

If neither name is suitable, perhaps the design/concept is at fault. 

> For example, if we want to use anything outside the namespace, we'd need to prefix the names with outer->, eg outer -> f().

So explicit scope resolution is required for functions (and variables) which are defined outside the namespace *even within a scope which encloses the namespace*? In what way are they *not* self-contained (in contrast with a "module")?

I was thinking only of the inverse - accessing procedures defined inside the namespace from outside. It seems that procedures would need to be grouped based on what settings they use, when perhaps they should be grouped some other way or kept in the global namespace.

Allowing the use of a "namespace" just for its settings is a "creative" solution, but not coherent.

A namespace establishes the scope for names within it, allowing a *name* to be bound to a variable or function independently of what it is bound to in other scopes. Within a new scope, either a name is bound to the same variable/function as before, or it is bound to a new variable/function.

`SendMessage` being present but having different behaviour in each namespace could be explained a number of ways (regarding only semantics, not implementation):
  - Each namespace automatically contains copies of all standard functions, re-compiled to refer to variables within that namespace if present. The built-in variables can be redefined in each namespace, but automatically have the same restrictions as the outer variables.
  - The standard function names are bound in every namespace, but to the same functions. These functions then use a variation of dynamic scope to resolve built-in variables, allowing them to be defined within the "current" namespace, although they automatically have the same restrictions wherever they are defined. (Only, if *dynamic scope* utilizes the scope of the caller, it should consider local variables as well, not just the caller's namespace.)
  - Each "namespace" is actually an instance of the standard library (i.e. standard functions are instance methods and built-in variables are instance properties).

# Thread-safety

> Fundamentally, it doesn't makes sense that a certain thread has a certain setting for, eg, case sensitivity or string encoding, 

It makes more sense than what you propose.

*Thread-local* variables are a simple and direct way of avoiding conflicts arising from multiple threads relying on the same global variables. In our case the threads aren't running simultaneously, but a new thread may launch at any time and interrupt a previous one, starting with potentially unpredictable settings and causing the interrupted thread's settings to be changed at unpredictable points in unpredictable ways.

Tying the settings to "namespaces" limits the scope of conflicts, but does not eliminate them, *even if* the user takes care to enclose functions in "namespaces" based on what settings they use, or uses "namespaces" just for settings. (I think it isn't likely that the average user will do that much.) The reason is that within a given namespace, there is only one copy of the settings, not one per thread or function call.

# Thread-local vs namespace-local, or...?

> This might be perceived as convenient, but [...] there is a lot of inconvenience implied before we arrive at this convenience. 

That's one way to spin it. Another is that you wish to make hotkeys less convenient in order to make things more convenient for library authors.

But what makes you think "thread-local" and "namespace-local" are mutually exclusive, or the only two options?

Consider this: a namespace (in any other language) defines the scope of a name, not the extent or storage duration of the variable (although that may be affected by scope). Thread-local variables have a different storage location for each thread, but the variable's *name* could be scoped to a given block or namespace, and there could theoretically be multiple names bound to the same thread-local variable.

The "inconvenience" you write of is **not** implied by the automatic thread-localization, but by a lack of a convenient way to localize changes to the settings. You've decided that namespace-local is more appropriate and more convenient, but surely you see that it has its own limitations?

# Alternatives

There are simpler, more conventional solutions. For instance, based on your previous example,
```AutoHotkey
dhw_on := ahk.clone()
dhw_on.detecthiddenwindows := true
x::
	; ...
	dhw_on.sendmessage(...)
	ahk.winexist(...)
	; ...
return
```
This model is straight-forward and directly translates to other common languages. Supposing that the AutoHotkey library is eventually separated from the interpreter, it could work the same way in most languages.

If the library was object-based, further dividing it could greatly reduce the impact of (or need for) some global settings. Specifically, variables that affect window identification can be limited to just a handful of functions which return a `Window` object, like `WinExist`, with all functions which target a window either being converted to methods or accepting a Window object (or HWND) as a parameter.

I've been reluctant to consider this because:
  - I think there is a perception that "OOP" is a hurdle for users who are new to programming (or new to OOP).
  - Object-based designs are harder to optimize.
  - It would break further away from AutoHotkey's original "identity".

But that aside, good object-based designs inherently avoid a bunch of problems. The more I think about it, the more it seems we'd be better off with this type of design, plus a focus on helping newcomers over the initial hurdles.

HelgeffegleH:
# Semantics

Some definitions, taken from __Wikipedia__,

>In computing, a **namespace** is a set of symbols that are used to organize objects of various kinds, so that these objects may be referred to by name.

[src](https://en.wikipedia.org/wiki/Namespace) (2019-08-03)

> **Modular programming** is a software design technique that emphasizes separating the functionality of a program into independent, interchangeable **modules**, such that each contains everything necessary to execute only one aspect of the desired functionality.

[src](https://en.wikipedia.org/wiki/Modular_programming) (2019-08-03)

These definitions are broad, but quite good summaries in my opinion. In programming, terminology isn't always well defined, unambiguous or completely agreed upon. I do not really care if agree on any exact definitions. A module implies some namespace, if the module has objects of various kinds which may be referred to by name. A namespace is a module if it contains everything necessary for its desired functionality. I didn't say this couldn't be called modules or something else. Changing the _name_ is [trivial](https://github.com/HelgeffegleH/AutoHotkey_L/blob/NSs/source/NameSpaceDefines.h#L43), at least from the script's point of view. I committed much more to the name _namespace_ in the actual implementation, but it is still easy to adjust it.

> A namespace should do no more than define a scope for names.

This doesn't have much practical meaning. The scope for the names is the part of the script where they can be used (without being fully qualified), this is exactly what `namespace {...}` does. It allows all names defined inside the block to be used there (and only there) without being fully qualified and without the risk of any name clashes.

> This is sounding much more complicated than a namespace or even module should be.

Why is that? I do not at all feel limited by any thing I __should__ or __shouldn't__ do, my main intention was to make something useful. I could have made it simpler, more limited, I could have made something else entirely

> > `#if`s belongs to the namespaces

> Why is that?

`#if` applies to hotkeys, which are part of the namespace, all variable and function references in the expression refer the variables and functions of the namespace in which the `#if` appear. It is convenient, helps to _facilitate writing code which is much less dependent on the script in which it is used..._. 

>  In what way are they not self-contained (in contrast with a "module")?

I see your point, my point was merely that _they_ are not self-contained if they are contained in something else on which they rely.

# Thread-saftey and more

> Thread-local variables are a simple and direct way of avoiding conflicts arising from multiple threads relying on the same global variables.

It is a fact, and the [_variables/state_](https://github.com/HelgeffegleH/AutoHotkey_L/blob/NSs/source/defines.h#L884-L921) which makes sense as thread-local are kept as such. The way _our_ function settings work is not like any _normal use_  of thread-local variables, if it were, we would never be required to do anything like,

```autohotkey
px := a_x
a_x := y
bif()
a_x := px
```
That (the code above) is plain ridiculous, imo. 

> That's one way to spin it. Another is that you wish to make hotkeys less convenient in order to make things more convenient for library authors.

How is that a spin? Do you disagree that the _convenience_ comes without cost for script authors? There is not even a consensus within the AHK community, or a recommendation in the documentation, on who is responsible for managing the settings. With the current situation, in a thread callback, we could just as well need to set some setting before each lib function we call, and restore some settings after they return, or we need to analyse each lib function and all functions it calls and to ensure it doesn't change the settings we need and that we do not mess up the settings it needs. And all this work needs to be redone for every update of the libs we use. Also, if it is more convenient to make libraries, that would benefit hotkeys too.

> You've decided that namespace-local is more appropriate and more convenient, but surely you see that it has its own limitations?

In my opinion it is more manageable in a majority of cases. Yes, it has its own limitations, hence, I mentioned in my first post that,

>  it would probably be better if all settings were function parameters for the relevant functions instead. [...] This is another discussion though.

I thought this would be a separate discussion since the problems with global settings is an inherent problem with language which would need a separate discussion / decision regardless.

> But what makes you think "thread-local" and "namespace-local" are mutually exclusive, or the only two options?

I don't. Making the settings both thread-local and namespace-local would have implied much more work and _overhead_ than I am willing to put in in order to _save_ the global settings, I would prefer to eliminate them. And as I mentioned in the first comment, this is not complete. 

> If the library was object-based, further dividing it could greatly reduce the impact of (or need for) some global settings

Object-based or not, I don't see how there is a _need_ for global settings at all, neither in the current alpha nor in this branch. For the _window functions_, replacing the settings with properties for some object is probably a better idea than parameters for the functions 👍 .

Cheers.

Lexikos:
> this is **exactly** what namespace {...} does. 

In other languages, yes, but here that is a **subset** of what namespace {...} does. 

> This doesn't have much practical meaning.

The name of the construct has practical impact. Users are more likely to read the documentation upon encountering a new construct if they do not have clear expectations about it based on prior experience. Even after reading the documentation, it is easier to remember when the construct matches up with prior experience, or at least does not conflict.

> #if applies to hotkeys, which are part of the namespace,

That doesn't make sense either. Hotkeys are global, or "scoped" to the conditions under which they trigger. Having hotkey *labels* belong to a namespace has practical value, as it makes calling or referring to a specific variant possible, but the label is just incidentally created by the hotkey definition, and not a permanent part of the hotkey.

> all variable and function references in the expression refer the variables and functions of the namespace in which the #if appear.

That much makes sense.

I do not think it is necessary or wise to allow duplicate variants. There is a conflict - it is known in advance that only one will work, so there should be an error message. How do you disambiguate with the Hotkey() function? Call `mynamespace->Hotkey()`? (I suppose that would normally be implied, provided that you call the function from within the namespace.)

> The way our function settings work is not like any normal use of thread-local variables, 

I disagree. If a thread-local variable needs to be changed temporarily, how else would you expect it to be done? Perhaps the details might be hidden away in a constructor/destructor or the using/IDisposable pattern, but it's still doing the same thing. At the lower level, functions do the same with thread-specific registers.

> How is that a spin?

I already said. The "inconvenience" you write of is not implied by the automatic thread-localization, but by a lack of a convenient way to localize changes to the settings.

Your proposal adds a way to localize changes, but also increases the need to do that in the first place, the risk of conflicts, and (once the user is aware of the potential issues) the time and cognitive cost of writing hotkeys (or other thread procedures) that rely on or change the settings.

If one needs to change the settings (or just one setting) temporarily *within* a namespace, must one use another namespace? Can the new namespace inherit the current settings of the outer namespace? Or does one use the "plain ridiculous" (simple and obvious) method?

> Also, if it is more convenient to make libraries, that would benefit hotkeys too.

Your proposal prioritizes one over the other, but it seems to me that both could be solved more elegantly by discrete constructs for namespaces and settings.

> There is not even a consensus within the AHK community, or a recommendation in the documentation, on who is responsible for managing the settings. 

I think it is generally understood that a library function which changes a global setting should restore it before returning.

> I thought this would be a separate discussion since the problems with global settings is an inherent problem with language which would need a separate discussion / decision regardless.

You brought it into the discussion by combining namespaces with global settings. They should be separate (and that basically summarizes my objection). *If* global settings are to be eliminated, that's all the more reason not to change the way that they work now.

> Object-based or not, I don't see how there is a *need* for global settings at all, neither in the current alpha nor in this branch.

In the current alpha, there are settings which cannot be specified or put into effect any other way. Do you honestly believe that no one needs these settings?

Any time someone changes or retrieves a setting, it is generally because they had some *need* to do so. Maybe it was the only way to achieve something, the best way, or just the easiest or most obvious. Maybe it wasn't strictly *needed* because there are alternatives or the script itself is just a convenience, but that's just nitpicking.

HelgeffegleH:
Hi again 👋.

I removed the _Library_ feature, and made a few very minor fixes.  (I'll update the downloads in near future)

I see I didn't respond to the last comments, I might return with comments when I find time to reread the conversation. 

Cheers.

HelgeffegleH:
Hi.

> I do not think it is necessary or wise to allow duplicate variants. There is a conflict - it is known in advance that only one will work, so there should be an error message. 

I don't think it is worse than allowing,
```autohotkey
#if true
x::y
#if 1
x::z
```

In contrast to the above, in the following, which could be a deliberate design, each hotkey can be fired depending on the state of the `toggle` variables
```autohotkey
namespace n1 {
	z::toggle := !toggle
	#if toggle
		r::msgbox 'n1'
}
namespace n2 {
	x::toggle := !toggle
	#if toggle
		r::msgbox 'n2'
}
r::msgbox 'script'
```
You could also have something like duplicate `#if winXXX(...)` with different match modes and dhw settings, for example.

> How do you disambiguate with the Hotkey() function? Call `mynamespace->Hotkey()`? (I suppose that would normally be implied, provided that you call the function from within the namespace.)

You suppose correctly, for example in,

```autohotkey
n1 -> hotkey('if', 'toggle')
n1 -> hotkey('z', 'g')

namespace n1 {
	x::toggle := !toggle
	#if toggle
	g() {
		msgbox a_thisfunc
	}
}
```
the `z` hotkey will only fire if you have pressed `x` to _toggle on_ the the `#if` condition. Having 
```autohotkey
toggle := true
#if toggle
```
in the script namespace bears no relevance.

> If one needs to change the settings (or just one setting) temporarily within a namespace, must one use another namespace? 

No, we can choose to explicitly set whichever settings we like when needed, or set and restore if we like, there needs to be considerations in case of interruptions, but only within the namespace, which should be easy to manage because you can write small blocks of code which is easier to overview, and you can also use `critical`. 

>  Can the new namespace inherit the current settings of the outer namespace?

This is not a feature. I originally decided against any kind of `#use` features, but I have changed my mind about that, although not due to wanting to use settings. My memory is not fresh enough to comment on the ease of adding it. I've barely touched this since making the PR, but ofc it should be feasible.

> I think it is generally understood that a library function which changes a global setting should restore it before returning.

I've seen several counter examples though, even by experienced users.

> You brought it into the discussion by combining namespaces with global settings. They should be separate (and that basically summarizes my objection). If global settings are to be eliminated, that's all the more reason not to change the way that they work now.

Indeed it was a mistake. I don't think it is very difficult to restore it though.

> Do you honestly believe that no one needs these settings?

No, but no one needs them to be global (to the script or thread).

___

If I were to work more on this branch, I suppose I'd continue with restoring thread settings, I'm not sure about the auto exec timer though, that always seemed extremely weird to me and it seems a bit disgusting to have one timer per auto exec section- But having only one timer would make behaviour less predictable. I could also substitute (in the docs and implementation) the word __namespace__ for something else, I'm not sure you really suggested that the word __module__ would be better suited though.

Perhaps this have sparked some inspiration and you'd like to implement _something of this kind_ yourself? That could bring a convenient end to this PR :wink:.

Cheers.

HelgeffegleH:
As far as I am concerned, this has been replaced by #162. I close this for that reason.

PR #145, opened 2019-07-11
Updating BIF_PostSendMessage to accept buffer-like objects for wParam/lParam.

HelgeffegleH:
Reason, convenience and symmetry with, eg, `dllcall` and `numget`.

### DIFF
--- a/source/script2.cpp
+++ b/source/script2.cpp
@@ -1988,16 +1988,28 @@ BIF_DECL(BIF_PostSendMessage)
 		case SYM_INTEGER:
 			param[i-1] = (INT_PTR)this_param.value_int64;
 			break;
-		case SYM_STRING:
-			LPTSTR error_marker;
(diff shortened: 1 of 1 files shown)


source/script2.cpp
```diff
@@ -1988,16 +1988,28 @@ BIF_DECL(BIF_PostSendMessage)
 		case SYM_INTEGER:
 			param[i-1] = (INT_PTR)this_param.value_int64;
 			break;
-		case SYM_STRING:
-			LPTSTR error_marker;
-			param[i-1] = (INT_PTR)istrtoi64(this_param.marker, &error_marker);
-			if (!*error_marker) // Valid number or empty string.
+		case SYM_OBJECT: // Support Buffer-like objects, i.e, objects with a "Ptr" property.
+			size_t ptr_property_value;
+			GetBufferObjectPtr(aResultToken, this_param.object, ptr_property_value);
```
HelgeffegleH: If #146 is accepted, this should be updated accordingly.

PR #146, opened 2019-07-11
Improving detection of invalid Buffer-like objects passed to built-in functions and methods.

HelgeffegleH:
Example,

```autohotkey
dllcall callbackcreate((p)=>msgbox(p)), 'ptr', {} ; new: error, old: passes 0
```

Affects, `numget`, `numput`, `strget`, `strput`, `file.rawread`, `file.rawwrite` and `dllcall`.

Reason, to help facilitate error detection.

### DIFF
--- a/source/TextIO.cpp
+++ b/source/TextIO.cpp
@@ -893,9 +893,8 @@ class FileObject : public ObjectBase // fincs: No longer allowing the script to 
 					if (IObject *obj = TokenToObject(target_token))
 					{
 						size_t ptr, size;
-						GetBufferObjectPtr(aResultToken, obj, ptr, size);
-						if (aResultToken.Exited())
-							return aResultToken.Result();
+						if (!GetBufferObjectPtr(aResultToken, obj, ptr, size))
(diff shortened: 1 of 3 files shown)


Lexikos:
This would break handling of EARLY_EXIT.

HelgeffegleH:
Hi. I'm not certain how this was handled before. The `ResultToken` passed to `GetBufferObjectPtr` is only ever _touched_ by `GetObjectPtrProperty`, which as far as I can tell will only set its result to `FAIL`, and under the same conditions as before the change. But even if its set to `EARLY_EXIT`, `GetObjectPtrProperty` would return `FAIL` anyways. And my changes wouldn't have removed the `EARLY_EXIT` result.

I suppose I could add back the `aResultToken.Exited()` checks, but `DetermineTargetControl` which calls `GetObjectPtrProperty` doesn't check  it.

If you could guide me I would be most grateful.

Cheers.

Lexikos:
I see now that some of the work I have done on this code has not been published yet.

Lexikos:
This appears to have been superseded.

PR #147, opened 2019-07-11
Changing BIV_LastError to use _ultot instead of _itot.

HelgeffegleH:
Example,
```autohotkey
a_lasterror := -1
msgbox a_lasterror ; new: 4294967295. old: -1
```

Reason, documentation says it should be unsigned.

### DIFF
--- a/source/script2.cpp
+++ b/source/script2.cpp
@@ -9543,7 +9543,7 @@ VarSizeType BIV_LastError(LPTSTR aBuf, LPTSTR aVarName)
 {
 	TCHAR buf[MAX_INTEGER_SIZE];
 	LPTSTR target_buf = aBuf ? aBuf : buf;
-	_itot(g->LastError, target_buf, 10);  // Always output as decimal vs. hex in this case (so that scripts can use "If var in list" with confidence).
+	_ultot(g->LastError, target_buf, 10);  // Always output as decimal vs. hex in this case (so that scripts can use "If var in list" with confidence).
 	return (VarSizeType)_tcslen(target_buf);
 }
 


Lexikos:
> Reason, documentation says it should be unsigned.

An error in the documentation is hardly a good reason to make a change to the program.

Considering a change to A_LastError, I would probably opt for unsigned hexadecimal, although I'm not sure whether that's actually suitable for GetLastError(). I might be thinking of HRESULT and SEH exception codes like 0xC0000005. The reason for decimal formatting, at least in the case of A_ScriptHwnd, is that if/when it is changed to pure integer, it will be implicitly decimal. However, I think A_LastError is typically viewed or logged and not used numerically (except to pass to FormatMessage), so string makes some sense.

HelgeffegleH:
I didn't realise it was an error in the docs. I see `HRESULT` is signed, I didn't remember this and assumed it was be unsigned, given the claim that `a_lasterror` should be unsigned that is. Wouldn't it be better to return the `HRESULT` from COM invocations via the exception object than using `a_lasterror`? `ComObjError` seems superfluous since `OnError`, especially if you add the `HRESULT` to the exception object so that `onError` callbacks can easily identify Com errors.

Cheers.

Lexikos:
Even if the docs were correct, there has to be a reason for the (documented or actual) behaviour.

HRESULT may be a signed type, but the values are composed as bit fields and documented as unsigned hexadecimal integers. The sign or "severity" bit is [always set for errors](https://en.wikipedia.org/wiki/HRESULT#HRESULT_format).

My point was that the only error values I could think of that set the sign bit are best displayed as unsigned hexadecimal, but I'm not sure that such codes are ever returned by GetLastError(). I was not thinking of A_LastError returning a HRESULT code, rather assuming that it wouldn't.

I suppose that if GetLastError() never returns values with the most significant bit set, _ultot vs. _itot only matters for values set by the script. But why is the script setting that value? In any case, GetLastError returns DWORD, so if we have to pick one, unsigned is probably the more sensible choice.

Exception handlers and OnError callbacks have equal access to A_LastError. But actually, COM errors do not set A_LastError (see 5475c5c2). ComObjQuery sets it, but never raises a true COM error (just a fake one when given an address < 65536).

It is already easy to distinguish between COM errors and non-COM errors based on the message format, more so than between other errors. Eventually they should be recognized by type/class, not by the presence of a HRESULT code.

OnError does not supersede ComObjError.  The latter permits the script to ignore COM errors *and continue*.

HelgeffegleH:
> It is already easy to distinguish between COM errors and non-COM errors based on the message format, more so than between other errors.

I do not know how to do that, perhaps I could find out by causing COM errors and analyze the format of the text message, but I can't recall having seen any documented format I could rely on. Why I highly approve of:

> Eventually they should be recognized by type/class, not by the presence of a HRESULT code.

___

> OnError does not supersede ComObjError. The latter permits the script to ignore COM errors and _continue_.

True, I overlooked that. 👍 

Cheers.

Lexikos:
Almost none of the error messages are documented, so if lack of documentation is a problem, don't distinguish between exceptions (yet). I suppose you must weigh your goals against the value of strictly adhering to documentation.

On the other hand, source code can be considered a form of documentation, though not a guarantee for future versions. Also, an observant person will probably notice that COM error messages are always preceded by the HRESULT code and followed by Source/Description/HelpFile/HelpContext.

PR #151, opened 2019-09-08
Adding call to SetDllDirectory before calling Script::PreparseExpression

HelgeffegleH:
Reason, so that (the last occuring) `#DllLoad directory` doesn't affect all calls to `LoadLibrary` for `DllCall` optimizations in `Line::ExpressionToPostfix`.

### DIFF
--- a/source/script.cpp
+++ b/source/script.cpp
@@ -1559,7 +1559,12 @@ UINT Script::LoadFromFile()
 	if (   LoadIncludedFile(g_RunStdIn ? _T("*") : mFileSpec, false, false) != OK
 		|| !AddLine(ACT_EXIT)) // Add an Exit to ensure lib auto-includes aren't auto-executed, for backward compatibility.
 		return LOADING_FAILED;
-
+#ifdef ENABLE_DLLCALL
+	// So that (the last occuring) "#DllLoad directory" doesn't affect calls to GetDllProcAddress for run time calls to DllCall
+	// or DllCall optimizations in Line::ExpressionToPostfix.
+	if (!SetDllDirectory(NULL))
+		return ScriptError(ERR_INTERNAL_CALL);
+#endif
 	if (!PreparseExpressions(mFirstLine))
 		return LOADING_FAILED; // Error was already displayed by the above call.
 	// ABOVE: In v1.0.47, the above may have auto-included additional files from the userlib/stdlib.


Lexikos:
ExpressionToPostfix isn't supposed to call LoadLibrary, only GetModuleHandle. IIRC, GetModuleHandle may retrieve a loaded module even if the path doesn't match. I don't think SetDllDirectory affects it.
```
if (   !hmodule_to_free  ||  !(hmodule = *hmodule_to_free = LoadLibrary(dll_name))   )
{
	if (hmodule_to_free) // L31: BIF_DllCall wants us to set ErrorLevel.  ExpressionToPostfix passes NULL.
```
As it says, ExpressionToPostfix passes NULL for hmodule_to_free, which short-circuits the LoadLibrary call.

HelgeffegleH:
That makes sense. `#DllLoad dir` does affect `GetModuleHandle`, at least if you specify a relative path.

Example,

```autohotkey
#DllLoad MyDir
#DllLoad Rel\MyDll

dllcall 'Rel\MyDll\MyFunc'
```
In the above, the dllcall is _optimised_, but with an (anywhere) added `#DllLoad` or `#DllLoad MyOtherDir` it will not be. So it becomes unpredictable. Also, `#DllLoad dir` will affect run time calls to `dllcall`.

I updated the commit message and code comments.

Cheers.

Lexikos:
I see. The documentation for GetModuleHandle does not specify how paths are used, and given that it need only look for modules that are already loaded, I would not have assumed that it searches the search path list in the same manner as LoadLibrary.

> If a relative path is specified, the entire relative path is appended to every token in the DLL search path list. 
> (LoadLibrary)

It seems one side-effect is that `".\dllname.dll"` doesn't actually reference the current working directory, but any directory in the list, the same as `"dllname.dll"`. I wonder if this could become an issue.

> IIRC, GetModuleHandle may retrieve a loaded module even if the path doesn't match. 

Perhaps I did not recall correctly. I recall it happening (years ago), but cannot reproduce it.

> if (!SetDllDirectory(NULL))

Can this realistically fail? What are the implications of failure? I'm inclined to omit the error checking to minimize code size. 

> Also, #DllLoad dir will affect run time calls to dllcall.

#DllLoad should have the same effect on DllCall as on the optimization (possibly no effect).

It would make more sense for the effect to be persistent if we were using AddDllDirectory rather than SetDllDirectory, especially if the script is loading some dlls at runtime, or referring to the dll by relative path in both #DllLoad and DllCall. It seems counter-intuitive for this to fail because DllCall is not searching the same directories:
```
#DllLoad MyDir
#DllLoad Rel\MyDll
...
DllCall 'Rel\MyDll\MyFunc'
```
Of course, `Rel\` can be omitted after the dll is loaded, but there may be a reason to specify the path. For instance, to distinguish between `mylib-1.1\mylib.dll` and `mylib-1.2\mylib.dll`.

HelgeffegleH:
> `if (!SetDllDirectory(NULL))`
Can this realistically fail? 

I only based the check on the fact that the [documentation](https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-setdlldirectoryw) implies failure can occur,

> If the function fails, the return value is zero.

Which further states,

> To get extended error information, call GetLastError.

which I suppose would answer,

> What are the implications of failure?

I guess that failure when passing `NULL` is not very likely, and perhaps only possible if there is some system instability, in which case the program would probably fail pretty soon anyways, so the check is probably unnecessary.

> #DllLoad should have the same effect on DllCall as on the optimization (possibly no effect).

Makes sense.

> It would make more sense [...]

I agree.

PR #152, opened 2019-09-12
why not bind coordmode to command?

hyaray:
everytime mouse move and click,
I have to save coordmode and do something, then set coordmode back,
this make things heavy,
add commands like this `MouseMoveByWindow(x, y)` seems easier.
How do you thinging about my suggestion?


Lexikos:
This is not a valid pull request. The GitHub issue tracker is disabled because I don't want to deal with bug reports or suggestions on GitHub. A pull request is not a valid substitute.

In general, suggestions should be posted on the forum. 

However, I will not duplicate the functions in this way. There are at least 11 functions affected by CoordMode in v2 and 4 coord modes. Supposing the original functions and CoordMode are eliminated, that's an extra 33 functions.

A more elegant approach would be to provide functions for translating between CoordModes, and let these be combined with the existing functions. However, it is generally already trivial to perform this conversion; CoordMode is just a convenience.

PR #154, opened 2019-10-17
v2: Replace deprecated <typeinfo.h>.

lvalkov:
As of [MSVC v14.23](https://docs.microsoft.com/en-us/cpp/overview/cpp-conformance-improvements?view=vs-2019#standard-library-improvements-1), the header is no longer part of the toolset.


Lexikos:
Thanks, but the #include itself was obsolete and has been removed.

PR #155, opened 2019-10-18
v2: Fix invalid read of uninitialized InputHook variable.

lvalkov:


Lexikos:
`vk_by_number` must correspond to `vk`; it indicates how `vk` was specified. `vk` is set on each iteration, so initializing `vk_by_number` outside of the loop will bypass the warning but will not fix the error. Something like `"{vk13}a"` would still behave incorrectly because `vk_by_number` is not reset to false for "a".

Thanks for bringing my attention to it.

PR #157, opened 2019-11-25
adding support for __thiscall to BIF_DllCall

mmikeww:
reference Helgef's improvement from this thread:
https://www.autohotkey.com/boards/viewtopic.php?p=265351#p265351

> new:
> specify thiscall for the return type to use calling convention __thiscall. Puts first argument in ecx register, except that, the behaviour is as __stdcall.
> 
> reference: https://docs.microsoft.com/en-us/cpp/cpp/thiscall?view=vs-2017
> 
> Reason, implementation is simple and this is difficult to achieve in script.

### DIFF
--- a/source/script2.cpp
+++ b/source/script2.cpp
@@ -12506,6 +12506,7 @@ VarSizeType BIV_TimeIdlePhysical(LPTSTR aBuf, LPTSTR aVarName)
 #define  DC_BORLAND             0x0001      // Borland compat
 #define  DC_CALL_CDECL          0x0010      // __cdecl
 #define  DC_CALL_STD            0x0020      // __stdcall
+#define  DC_CALL_THISCALL		0x0040		// __thiscall
 #define  DC_RETVAL_MATH4        0x0100      // Return value in ST
 #define  DC_RETVAL_MATH8        0x0200      // Return value in ST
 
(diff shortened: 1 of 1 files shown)


HelgeffegleH:
There were some reasons why I didn't make the pull request myself, other than forgetting about it. At least I remember that supporting unquoted return type is unnecessary since it has been deprecated. Other considerations, should we enforce the first parameter to be ptr or atleast ptr-sized integer. I think gcc allows compiling thiscall with the first parameter as float, so that is weird. Should callbackcreate be updated to match? Probably something else I can't think of right now. 

__con__: probably one use case through out the entire existences of the language. And since 32 bit calling conventions are being less and less prevalent, I don't expect this to become very useful in the future. 

cheers.

mmikeww:
Sorry about that Helgef. I did assume that perhaps you had just forgotten about this when I stumbled across it. I shouldn't have taken the liberty to submit on your behalf

HelgeffegleH:
No need to apologise, as far as I'm concerned you made the pull request on your own behalf. I had forgot, I'm glad you reminded me. 

Cheers.

Lexikos:
I've decided not to merge this, but for anyone who needs to call a *thiscall* function, I've demonstrated [how to call a thiscall function using machine code](https://www.autohotkey.com/boards/viewtopic.php?f=76&t=91105&p=402798#p402798).

PR #160, opened 2020-02-08
Fix memory leaks

HotKeyIt:


Lexikos:
The reference count for objects derived from NativeFunc must always be positive - the Release calls were intentionally omitted. It does not matter whether the count is *correct* as such because the object cannot ever be deleted. With the current design, the added Release calls serve only to increase code size; although they hypothetically would become useful if the memory allocation method is changed to allow dynamic unloading of script or proper cleanup after `eval()`.

PR #162, opened 2020-03-29
introducing namespaces, second version.

HelgeffegleH:
## Introduction
This is an alternative to #142. The goal of this PR is similar to that of #142, but this one does not address the global state of the script. All namespaces now share the same state. This design is much less complete, but perhaps more conventional, and implies no breaking changes except for a few reserved words. I have left many decisions open. This has also been much less tested than #142.

## Documentation and files
There is no documentation, please refer to this post for an introduction to the proposed syntax and usage.  Files are available for ➡️ [__download here__](https://github.com/HelgeffegleH/NS_FILES_AND_DOCS/tree/master/second_version) __(Updated: 2020-06-28)__.

## Syntax and usage
__Edit__, `namespace` was changed to `module`, files has been updated (2020-06-28).

Use the `namespace` keyword to declare a new scope for functions and variables,
```autohotkey
namespace [name] {
    ; any code...
}
```

### Scope resolution

To refer to a __function__ or __super-global variable__ defined in the namespace _name_, from outside of it, use the dot operator, eg,
```autohotkey
namespace ns {
    global x := 0
    f() => 1
}
msgbox ns.x ns.f() ; 01
```
Dynamic scope resolution uses the same syntax as dynamic object/prop/method syntax, eg, `%'ns'%.%'x'%` resolves to the variable `x` in `ns`.

To create a file scope use, `#Import`

```autohotkey
#Import [*i] path [ to name]
```
Works as in #142 but uses _to_ as the separator between _path_ and _name_.

__Functions__ and __super-global variables__ defined in namespace `X`, can be referenced without scope resolution in namespace `Y` if declared with the `#UseFunc`  or `#UseVar` directives in namespace `Y`,
```autohotkey
#UseFunc func_list in name
#UseVar var_list in name
```
where _func_list_ and _var_list_ are comma separated lists of names. Use the marker `*All` to use all names. These directives are only positional relative to each other, example, 
```autohotkey
#UseVar *all in ns1 ; This will not include y since ns1 includes y only after this directive is processed, but it will include x.
namespace ns1 {
	global x
	#UseVar y in ns2
	namespace ns2 {
		global y 
	}
}
```
The `extends` keyword supports (non-dynamic) scope resolution, for example, `class x extends myNamespace.myClass`. Both nested classes and namespaces are supported.

### Reserved names

The names **_Script_**, **_outer_** and **_std_** are reserved as namespace names, with the same meaning as in #142. Variables and namespaces cannot share names within the same namespace. Some names which would otherwise be valid, cannot be used as namespace names, that is any of the BIV names (including super global _BIVs_), and also not `this` or `value`.

### Built-in super globals

These are still added to the _Script_ namespace, adding them to _std_ (or some other hierarchy or set of namespaces) would make more sense imo, but I left this undecided. As a consequence, any namespace other than _Script_ needs to either fully resolve these names or use `#UseVar` to access them. Example,

```autohotkey
namespace {
	msgbox type(array) ; String
}
namespace {
	msgbox type(array) ; Class
	#usevar array in Script
}
namespace {
	msgbox type(Script.array) ; Class
}
```

### Built-in functions

Are still accessible everywhere by default, and can always be accessed via the `std` namespace if overridden _locally_.

### Built-in variables

Can only be accessed without any scope resolution. (Except the super global ones, they are in _Script_).

### Auto execute section and static

No changes, execution flows from top to bottom and into any `namespace` unless stopped by a 
`return` or directed elsewhere by any other means. Example,

```autohotkey
msgbox 1
namespace {
	msgbox 2
	namespace {
		msgbox 3
	}
}
msgbox 4
return
namespace {
	msgbox 5
}
```
Shows __1,2,3__ and __4__. (#142 showed __3,2,5,1__ and __4__, which is more useful imo.)
Static variables and `__new`, no changes.

### Func references

There is (currently) no way  to retrieve a function reference from another namespace unless it explicitly defines function to do it, eg, `func(x) => std.func(x)`.

### Optimisations

Non-dynamic scope resolution is resolved as far as possible at load time. Fully non-dynamic scope resolution implies no run time performance penalties. The same goes for names referenced via the `#UseXXX` directive.

## Misc.

Ideas, comments, limitations, known issues, etc.

* Auto execution could instead jump over namespaces and we could initialise namespaces by other means, such as auto calling a specific function in each namespace, require _manual initialisation_ or add some new syntax, eg,  `auto { ; code } ` . Could also be like #142.

* __Labels__, `goto/gosub` can jump to any label in any namespace, these are still in one global namespace. I have postponed this until the auto execute behaviour is decided.

* A dynamically resolved module, not followed by any other scope resolution operation is not detected as an error. The result should be an empty string. This problem is similar to referring to an object without any following operations, eg, `msgbox obj` . Invalid non dynamic scope resolution is detected at load time, eg, `msgbox myNamespace ; error`.

* The use of `#UseXXX` vs extending the `#Import` syntax was chosen for maintainability and readability. Having the postfix _XXX_ means we do not need to prohibit functions and variables to share names.

* _Debugger_, only minimal changes made, and have not been tested. It is supposed to support context get in the correct global scope but cannot resolve scope. This is __TO DO__.

* Auto-include adds functions from the libs once per namespace which needs it. So each namespace has its own version.

* `#Include` works once per namespace, no need to use `#includeagain a.ahk` if `#include a.ahk` was used in another namespace. 

* Ideas and opinions (misc/misc)
    * Remove the BIV system, let constants be _const_ variables in a built-in namespace. _Dynamic_ BIVs can be replaced by functions for setting/getting, or better yet replace by function parameters / obj props, i.e., remove the global state. Some BIVs, such as `true`/`false`/`a_ptrsize` should not be vars at all, but reserved keywords, resolved at load time. Dynamic access to them are not relevant.

    * `#warn` should be namespace specific, or even better on by default (and treated as error).
    * I'd prefer to remove auto-include. We could instead allow something like `#UseFunc myFunc in lib` where lib is a reserved marker, or `#UseFunc <MyFunc>`. This is slightly less convenient but improves readability and maintainability.
    * The new syntax implies that with future changes we could allow scripts to use old versions of the standard library. For example, in __v2.1__ we might want to do something like `#UseFunc InStr in std.2.0` in a library written for __v2.0__, if `InStr` was subject to a breaking change in __v2.1__.

* Many things forgotten...

# About the implementation.

__TO DO__. I'll either edit it in here or make a new _post_. One comment only:

The implementation now uses the more _generic_ word `ScriptModule`, compared to `NameSpace` in #142. The word _namespace_ is now only committed to in some `#define`.

# How to continue

Collaboration, delegation or a _take-over_, are all fine by me.

Cheers ☕️ .


HelgeffegleH:
I've done some misc. work on this since the PR. I have split parts of the implementation into several files, more about this below. I have attempted to improve detection of invalid dynamic scope resolution, such as `%'MyModule'%` not followed by any other operations. In particular, I think things like `obj.x := obj[x] := var := %'MyModule'% := ... ` (chaining not significant) could cause undefined behaviour and `if %'MyModule'%` was `true`. 

Additionally, you cannot longer access bifs via scope resolution through any other module than `std`. For example, `MyModule.MsgBox()` is now only valid if `MyModule` defines its own `MsgBox` function. You can still refer to bifs without any scope resolution, in any module.

# About the implementation
This is a brief introduction to the implemenation, it is incomplete.

## `class ScriptModule`

For every _namespace_ there is one instance of `class ScriptModule`. Some important members:

* `mVar`, `mVarCount`, `mLazyVar`, `mLazyVarCount`, replaces the corresponing `Script::` members. `Script::FindVar` (and similar) now takes a `ScriptModule` as input, to indicate which variable list to use.
* `mFuncs`, replaces `Script::mFuncs`, `Script::FindFunc` now takes a `ScriptModule` as input to indicate which func list to use.
* `mNested`, stores nested modules.
* `mOuter`, refers to the enclosing module.

I have split parts of the implementation into several files for easier overview, in **source\ScriptModules**,

* **ScriptModules.h**, declarations and such.
* **ScriptModulesDefines.h**, macros, defines names and error messages related to modules.
* **ScriptModules.cpp**, includes,
    * **ScriptModulesAddNames.h**, implementation related to `#UseXXX`.
    * **ScriptModulesMisc.h**
    * **ScriptModulesNested.h**, handles nested modules.
    * **ScriptModulesOptional.h**, relates to the `*i` option for `#Import`.
    * **ScriptModulesReleaseObjects.h**, does similar to what `ReleaseVarObjects` did.
    * **ScriptModulesSourceFileIndex.h**, relates to allowing multiple `#Include` across modules without needing to do `#includeAgain`. Not included in the self-contained builds.
    * **ModuleListMethods.h**, `ModuleList`s are used to store nested modules. Eg, `ScriptModule::mNested` and `Script::mNested`.

## Global variables.

* `g_CurrentModule`, the current module, this is set on every iteration in `ExecUntil`, this is low cost. For example, `ExpandExpression` relies on this. During loading this variable is also important. Each new Line stores this value in `Line::mModule`.
* `g_StandardModule`, reference to the standard module. Added for convenience.

## Variable names and scope resolution.

Variables and namespaces cannot share names due to ambiguity. This also implies that variables cannot be created too early when encountered during parsing, since it is not known if there will be a namespace definition later. Instead, variables are now created in `Line::ExpressionToPostfix`. A notable exception is variables which are declared, this implementation has not been changed, there is no ambiguity in these cases. `ExpressionToPostfix` also handles load time optimisations and resolves non-dynamic scope resolution as far as possible. For dynamic scope resolution, `ExpandExpression` resolves (eg) `%'MyModule'%` to `SYM_MODULE` (a new `SymbolType` which is considered an operator) which then ends up in `Op_ObjInvoke`, which calls `Op_ModuleInvoke` to handle it. 

Name validation is done at several places, for example in ` Script::PreprocessLocalVars` and in the new method `Script::PreprocessGlobalVars`.

To be continued, perhaps.

Cheers.

Lexikos:
> invalid dynamic scope resolution, such as %'MyModule'% 

I assume that is invalid because MyModule is a namespace, not a variable or an object. This was somewhat understandable when the scope resolution operator was different, but now it just looks like you're accessing objects. Is there a good reason not to give the namespace an object interface, and allow it to be used like any object? At the very least, I would expect `x2 := x, x2.y` to be equivalent to `x.y`, aside from load time optimizations.

If a namespace is also an IObject, I suppose that there is no need for SYM_MODULE as SYM_OBJECT will suffice, and in place of Op_ModuleInvoke would be the object's Invoke implementation. (I haven't looked at the code.)

HelgeffegleH:
> I assume that is invalid because MyModule is a namespace, not a variable or an object. 

Correct.


> If a namespace is also an IObject, I suppose that there is no need for SYM_MODULE as SYM_OBJECT will suffice, and in place of Op_ModuleInvoke would be the object's Invoke implementation. (I haven't looked at the code.)

At some point, I considered this as a possibility to facilitate the implementation. I do not recall exactly why I abandoned the idea.

> Is there a good reason not to give the namespace an object interface, and allow it to be used like any object?  At the very least, I would expect x2 := x, x2.y to be equivalent to x.y, aside from load time optimizations.

Imo, disregarding any implementation technicalities:

The _name_ of a namespace is supposed to be just that, _a name_, it is used to qualify other names. The name doesn't hold a value. Doing `x2 := x, x2.y` is an error. It is supposed to be `x2 := 'x', %x2%.y`. Similar, `f := g, f()`, isn't equivalent to `g()`, we do `f := 'g', %f%()`. If we _allow_ `x2 := x`, then one might expect to be able to do `x := z`. But that doesn't permit consistency between `%'x'%.y` and `x.y`, if the latter is optimised. 

Cheers.

Lexikos:
> Auto execute section and static - No changes

I somehow didn't realize what this was saying until now (and haven't done any testing). I was thinking that the "auto execute section" of a module provides a good alternative to the use of static initializers for auto-execution (refer to [Static initializers and auto-execution](https://www.autohotkey.com/boards/viewtopic.php?t=46870)). I think I'll need to r

> The name of a namespace is supposed to be just that, a name, 

I've been thinking of it as a module the whole time, and I realize modules are probably what I wanted (in some form). You've also referred to them as modules repeatedly.

In Lua, modules are just tables, and as such can be aliased with simple assignment. Modules not only can be, but probably always are, loaded dynamically. I anticipate having similar capabilities in AutoHotkey, and suppose that a module needs to be (or be represented by) an object. In Python, everything is an object; you can assign a module object to another name and use it like the original name. In C# there's `using alias_ns = original_ns;`, though it's a directive, not really an assignment.

> The name doesn't hold a value. 

`x := 1` creates a variable. That variable has a name. The name does not hold a value; it is just a way of referring to the variable, which holds a value. A namespace holds a collection of names. I see no reason that it could not be an object, semantically or practically. In some (most) languages, classes are not values. In some languages, functions are not values. It is what we make it to be.

If we can't say that a namespace is an object, we can say there's an object which provides an interface to access whatever is defined within that namespace.

When you use a namespace-qualified name, it looks like you're accessing a member of an object within a variable. We *can* make it act more in line with its appearance, so why wouldn't we?

One small reason not to permit an assignment like `x2 := x` for aliasing - *if there is a way to create a proper alias* - is that `x2.y` won't be optimized, whereas a proper (more explicit) alias could be. However, assignment could have a valid use; the alias may conditionally refer to different namespaces/modules.

> If we allow `x2 := x`, then one might expect to be able to do `x := z`. But that doesn't permit consistency between `%'x'%.y` and `x.y`, if the latter is optimised.

I didn't suggest that we allow `x := z`. There's an obvious reason to disallow it, but that reason does not apply to `x2 := x`. Having certain "names" that aren't valid targets for assignment is not a new concept - we've had read-only built-in variables forever. I also intend to make `#Warn ClassOverwrite` (or something similar) mandatory.

There's also a precedent for names which only work when used with `.` (or `[`): `base.method()`. That came from C#, which also uses `.` for scope resolution. It is analogous to the C++ `the_super_class::method()`, in that case clearly using scope resolution. But I doubt many would connect that to the new namespace-modules. I think most would find it surprising that `x2 := x` is an error. It feels like an idiosyncrasy, perhaps similar to v1's `ClipboardAll`.

I now see you already suggested a namespace object, although exposed by a [c]namespace()[/c] function. (I wasn't ready to spend much time on this topic when you posted the original PR, and did not absorb everything I read, if I even read it all.)

Anyway, I won't reject this PR on the basis of not supporting standalone references to namespaces, or having them yield a "namespace object". I assume that it could be added later with minimal or no impact to compatibility - I assume that `x2 := x` raises an error.

> `#UseFunc` or `#UseVar` directives

I feel that these should be statements, not directives; like `local` or `global` (or `namespace`).

I assume `#UseVar` applies to classes; what about nested namespaces?

> [#Import] uses `to` as the separator 

I think it's conventional to use `as`. Does the path require quotes? I feel #Include should require quotes (except maybe `"<lib>"`), but I have neglected to change it thus far.

> ... any namespace other than Script needs to either fully resolve [the built-in super globals] or use `#UseVar` to access them. 

I still think it's weird for a namespace (or whatever it is) to not automatically have access to the names defined by the scope which encloses it. I'd imagine `namespace Script { ... }` as enclosing the entire script. How else would it contain every name not defined inside some other `namespace`?

I can see why you did it (reminds me of force-local), but maybe it doesn't need to be mandatory. 

If you define a namespace inside another explicit namespace, does the nested namespace "inherit" the names from the outer namespace (as would be the norm), or is it effectively self-contained?

I don't associate *imported* modules as being defined inside the main script, so in that case it makes sense to exclude names that were defined outside the module. That's generally what a library author would want. However, I don't think it makes sense to exclude the standard classes.

I stopped short of implementing a built-in class for every prototype because I was reluctant to suddenly dump so many names into the global namespace. On the other hand, I wanted to be able to write `a is Array` and similar.

> adding [the built-in classes] to std (or some other hierarchy or set of namespaces) would make more sense imo

I agree.

> A dynamically resolved module, not followed by any other scope resolution operation is not detected as an error. The result should be an empty string.

Is that "should be" referring to how you expect the current implementation to act, or the ideal?  I assume the former, since the ideal would be to throw an exception.

I guess that dynamic resolution is done within SYM_DYNAMIC handling. Detecting the following operator is easier in "infix" form, during ExpressionToPostfix. For instance, the `is_lvalue` field is true if ExpressionToPostfix detected an assignment following the SYM_DYNAMIC. It is used to throw for read-only variables. I suppose you can do something like that.

> I'd prefer to remove auto-include.

That's a possibility, but is there some reason you brought it up here? Is it inhibiting anything, or just requiring more work to support?

> This also implies that variables cannot be created too early when encountered during parsing, since it is not known if there will be a namespace definition later. Instead, variables are now created in Line::ExpressionToPostfix. 

There have been multiple reasons I've wanted to do that. I think the first was (the late) #MustDeclare. Probably also super-globals, which are (or were?) Var aliases if referenced in a function before they are declared, or direct pointers if referenced after. Maybe other reasons I'm forgetting.

HelgeffegleH:
I removed `SYM_MODULE` as you suggested, modules are now objects. You can now do `x := myModule, x.fn()` etc.  In addition to your points, this have the benefit of allowing reaching a non-hard-coded _depth_, eg,
```
while ?
	x := x.%expr%
x.??
```
which wasn't possible before. Although that could be (generally) allowed via something like `x.%p*%.??`.

Optional modules can be referenced but an exception will be thrown if you try to invoke it, eg
```autohotkey
#import *i no_file.ahk to n1
x := n1 ; ok
x.fn() ; error
``` 

>  I think I'll need to r

I noted this sentence wasn't complete. I will look into your other queries/comments later.

kczx3:
One of my biggest complaints with this current implementation are that built-in super globals must be referenced via Script or #UseVar. These should be accessible just as built-in variables and functions where no namespace is required but can be referenced by std if previously overridden.

Lexikos:
I'm fairly certain I'll be renaming `namespace` to `module` if you don't.

The current commit seems to be missing `ScriptModulesInvoke.h`.

Also, my first tests with the previous commit weren't very positive.

    namespace x {
        global z
    }
    namespace y {
        ; #usevar *all in x
        fy() => MsgBox(x.z)
    }
    x.z := 1
    y.fy()

Here, I get a "This value of type "String" has no property named "z"."  If I uncomment #usevar, I get a load time error stating namespace x couldn't be found, pointing at the file's last line.

So apparently, no names defined in the outer namespaces are accessible by default, even names of other namespaces. This is great for "maintainability", but, I think, very counter-intuitive. It appears that #usevar can't import a namespace from `outer` or `script`, unless it is importing everything.

It seems that #usevar imports names into the current namespace and treats them as if they were defined there. I think that when module A uses module B, module C should not see module B's members inside module A. #Usevar should only affect what the module itself sees, allowing it to avoid fully qualifying names but not affecting which names it "exports".

> These should be accessible just as built-in variables and functions where no namespace is required but can be referenced by std if previously overridden.

I agree.

> I think I'll need to r

I think I decided it required more thought before sharing, but left part of it by accident.

HelgeffegleH:
I added the missing file 🤦. 

Cheers.

HelgeffegleH:
>I assume `#UseVar` applies to classes; what about nested namespaces?

You can do 
```autohotkey
#UseVar myClass in myMod.myNestedMod

class x extends myClass { 
} 
; alt
class y extends myMod.myNestedMod.myClass {
}
```

I'm not sure if this answers your question.

>  super-globals, which are (or were?) Var aliases if referenced in a function before they are declared, or direct pointers if referenced after.

Speaking from memory, consider
```autohotkey
#usevar x in y
f() {
	global x	
}
namespace y {
	global x
}
```
* When the declaration in `f()` is reached, the variable `x` is created in `Script`.
* When the declaration in `y` is reached, another variable `x` is created.
* Then, before `PreparseExpressions`, `#UseVar` is processed, see `ScriptModule::ResolveUseParams()`. It will find `y.x` and replace `Script.x` with it, by simply overwriting it in the global list. The replaced var is _marked_ by `VAR_REPLACED`. If it pops up in `FindVar`, when searching in a function's list of globals, it is replaced there too. This can happen during runtime only if you declare a variable global inside a function, and only refer to it dynamically. It is done this way regardless of how declarations and directives are positioned. Declaring `x` super-global in `Script` will yield an error (when `#UseVar` is processed).

In short, I didn't use aliases, I think because I was unsure if that would cause some problems. I'll consider if I can improve it.

There should be no change to
```autohotkey
global x
f()=>(x,y)
global y
```

>[`#UseXXX`]I feel that these should be statements, not directives; like local or global (or namespace).

Do you mean basically the same thing but without the `#`? Variables could probably easily be allowed to be _used_ inside specific functions. _Used_ functions would probably not be possible to be referenced dynamically (`%expr%()/func(expr)`) without adding a list of _used_ functions, for every UDF.

> One of my biggest complaints with this current implementation are that built-in super globals must be referenced via Script or #UseVar. These should be accessible just as built-in variables and functions where no namespace is required but can be referenced by std if previously overridden.

@kczx3, as I mentioned in my first post, this behaviour is undecided. The reason these variables exist only in `Script` is almost coincidental.  I would like to be able to _override_ these classes _locally_, eg,
```autohotkey
; pseudo-code
namespace ns {
	class Array extends std.Array {
		g()=>msgbox(123)
	}
	([].g()) ; 123
}
([].g()) ; error
```

If they were available everywhere, like BIVs, then perhaps they should be BIVs. If they are not BIVs, but available in every module, then defining `Class Array extends std.Array` will need some special handling since the class `Array` would already exist (this currently is an error if you do it in `Script`).

One possibility which I have considered. is to confine these classes to a (set of) built-in module(s) by default, and require explicit inclusion. 
Pros,

* Makes it easier to add new classes without considering if its name might already be widely used.
* Users can be ignorant about built in classes when they start coding. No _surprising_ error when you do `map := true` or something of that kind.
* Self-documenting code.
* Simplify implementation. (From where we stand now that is. Ofc, this doesn't have to be the case.)

Cons,

* Could be perceived as verbose. Personally, I can feel a slight annoyance when I have to look up module names in python for example. But I haven't used python frequently for a few years now. I can't recall any annoyance when I did use it frequently.

I suppose it is inconsistent with BIFs. Requiring something like python's `from x import y` for _everything_ might be taking things it too far, I'm confident some would think so at least. Personally, I'm not opposed it, although I never intended it to be like that.

> [auto-include]  is there some reason you brought it up here?

I brought it up because the behaviour is _undecided_. Currently, every auto-include of a function creates a new copy of it. This means that static vars aren't shared, which might be desired, or not. 

> I'm fairly certain I'll be renaming namespace to module if you don't.

👍 I'll do it the next time I touch this code.

> So apparently, no names defined in the outer namespaces are accessible by default, even names of other namespaces. This is great for "maintainability", but, I think, very counter-intuitive. 

This observation is correct, and this is intended. It is not counter-intuitive to me :wink:. Have I understood you correctly in that you agree with the behaviour if we call it a _Module_ instead? Otherwise, I will consider if I can change the behaviour to meet your expectations. As a side note, one can do `#UseXXX *all in outer`, but there is currently no way to _use_ a module name, other than `x := outer.x` (which doesn't benefit from optimisations.)

> It seems that #usevar imports names into the current namespace and treats them as if they were defined there.

Correct.

> I think that when module A uses module B, module C should not see module B's members inside module A. #Usevar should only affect what the module itself sees, allowing it to avoid fully qualifying names but not affecting which names it "exports".

Do you mean like this,

```autohotkey
namespace A {
	#usevar x in outer.B
}
namespace B {
	global x
}
namespace C {
	msgbox outer.A.x ; Wanted: Variable not found
}
```

I'm not sure I see the benefits of this.

> I'd imagine namespace Script { ... } as enclosing the entire script. 

I think that is an accurate image. Basically it is like this, for a script __a.ahk__,

```autohotkey
namespace Script {
    #include a.ahk
}
namespace std {
}
```

`Script` and `std` are added to `g_script.mModules`.

### _Undecided_

I tried to make as few decisions as possible, while still producing something. In the end, @Lexikos will make all the decisions ofc. 

Cheers.

HelgeffegleH:
@Lexikos, would it be a problem if I pulled the latest alpha code? I could rebase my changes to occur after the new code (not like I did for #142 which _destroyed_ the commit history 🤦‍♂️ ).  It is a bit confusing to work on this when much of the code is out of date so to speak.

HelgeffegleH:
I tested to pull and rebase in a separate branch, this is the result ➡️ [sm_rb](https://github.com/HelgeffegleH/AutoHotkey_L/tree/sm_rb). Note the last commit to resolve some issues introduced by the rebase. It compiles but I didn't test it.

Lexikos:
> I'm not sure if this answers your question.

I already assumed classes work the way you demonstrated. What answers my question is:

> but there is currently no way to *use* a module name

It seemed to be possible when importing `*All`, but not individually.

> In short, I didn't use aliases, I think because I was unsure if that would cause some problems. 

That's good, as long as var references in the expressions are not pointing to the replaced var. I think I used aliases to avoid having to iterate through and update all of the variable references, which had already been resolved to the local variable before the global was created. (And I did it knowing I would eventually change how or when names are resolved.)

> Do you mean basically the same thing but without the #?

I meant exactly the same thing (no change in behaviour), but without the #. Inside a function, `global x` imports x from the current module into the function's namespace. It makes little sense to make `#usevar` a directive but `global` a declarative statement. I see no reason that the language should not eventually support importing a variable from a different module into the current function.

> If they were available everywhere, like BIVs, then perhaps they should be BIVs.

I think I'm missing the distinction between BIVs and other variables. BIVs should be treated as any other super-global variable, except that *currently* one is not allowed to redeclare them.

When I changed them to VAR_VIRTUAL, I imagined that if we keep these virtual vars, it may be possible for scripts to create them. I consider it a requirement that the module system (eventually) be able to support scripts adding to the built-in namespaces. For instance,
  - A "core" build of AutoHotkey could omit most of the built-in functionality, and parts could be re-implemented in script. That script could be embedded as a resource, the same way compiled scripts work, but potentially being used with an external script.
  - Some functionality is easier and smaller to implement in script, and being built-in offers no significant advantage other than being readily available to all scripts. For instance, a simple replacement for `Input()` was actually smaller than the (compiled) built-in code I removed that did basically the same thing.
  - AutoHotkey could be patched or improved in some ways without touching C++.

> If they are not BIVs, but available in every module, then defining Class Array extends std.Array will need some special handling since the class Array would already exist (this currently is an error if you do it in Script).

I guess this is a side-effect of there being no distinction between a name being *accessible to* or *used by* the module, and a name being *defined by* or *part of* the module. At least, I think it currently works that way, and it shouldn't. I'll have to consider how best to explain this. I'll start by saying that the external interface of a module should not be affected by which other modules it uses to do its work or how it refers to them (whether it uses #usevar or fully qualified names).

> I'm not sure I see the benefits of this.

For one, if `Array` is accessible to all modules by default but not a part of the module, the module should be able to define its own `Array`.

A module might define its own `x` and import `*All` from a module that coincidentally also defines its own `x`. In that case, the locally-defined `x` should take precedence. (Maybe that's already handled by deferring import and skipping names that are already taken; I don't know.)

> Makes it easier to add new classes without considering if its name might already be widely used.

Hiding away the core classes does not make it easier to add other classes later. Either way, classes can be added into other modules. You already suggested something like `std.2.0`. Obviously, there would be no `std.2.1` module until that version is released, and then whatever we add to that module will not affect existing scripts.

> Users can be ignorant about built in classes when they start coding. 

I would expect anyone to be ignorant about most things when they start, whether we "allow" them or not. They can be ignorant about built-in classes until they have some reason to learn. I see no problem with a "surprise error" being the reason for learning - perhaps it would make the learning more effective.

With them hidden away, one may get into a habit of using `object`, `array`, `gui`, etc. as their own variable names, then later find that they need to be renamed. It also makes code snippets more verbose, or less likely to work verbatim.

> Self-documenting code.

I think this is not a valid argument unless you apply it consistently, and you already know this is inconsistent with BIFs.

Every script will be using the built-in classes, whether they need to refer to them directly or not. So I'm not sure what the purpose of "self-documenting" is here.

> Simplify implementation. 

Maybe you mean "less work from this point" or "let's us be lazy right now". I don't see how this would affect the complexity of the final implementation. (I fully support your right to be lazy, as I don't expect you to do all the work.) 😉

> Currently, every auto-include of a function creates a new copy of it. This means that static vars aren't shared, which might be desired, or not.

I think that goes without saying, by definition. If there's another auto-include, it's because a new copy is being made.

I assume what you mean is that a given function might be auto-included once into each module, rather than auto-included once and then imported into each module. I think it would be better for an auto-include to be in its own *independent* module, and only be *referenced by* the calling module. However, I'm leaning toward removing auto-include.

Lexikos:
> I think that when module A uses module B, module C should not see module B's members inside module A.

I remembered what I thought were the main issues:

If I import *all from A, that does **not** mean I want to import everything from every module that A uses. That would in part defeat the point of having modules, especially because I think most users won't be so cautious that they'll individually import each item they need from each module.

If module C imports module A and module B defines a function f, module C cannot define a function f even though it never explicitly imports module B. Any module should be able to define whatever functions aren't already *defined in* that module, even if it imports *all from some other module which does.
```
module B {
    f() => MsgBox('B.f')
}
module A {
    #usefunc *all in outer.B
    f2() => MsgBox('A.f2')
}
module C {
    ; #usefunc *all in outer.A  ; Causes "Duplicate function definition".
    #usefunc f2 in outer.A
    f() => MsgBox('C.f')
}
c.f()
```
Currently the error points at the end of the file, which makes a counter-intuitive problem that much worse.

Again, having to specify `outer.` is counter-intuitive. It seems that most of the time, one would be importing from an outer module, not one defined inside the current module. Anyway, having to use `outer.` in any situation is counter-intuitive to me, and inconsistent with my choices around super-globals and closures.

kczx3:
Are there two ways that you can use a namespace from another file?  You can either #import the whole file or #include the file and then use the namespace that way?  If you #include a file that contains both globals and a namespace, are the globals in that file still merged with the main script's globals?

I'd really like to see the ability for a library author to _choose_ what they export to the consumer.  Right now, I guess that could be done using Closure's but it'd be nice for a namespace to be able to explicitly export things from a namespace.  All the rest would be private.

Also, and I think Lexikos was leaning this way too, but these really are more like modules to me than a namespace.

HelgeffegleH:
@Lexikos, thank you for your guidance. I think I can fix most issues you have brought up. I'll consider your comments and return with a more specific response later if I have something to say. Again, is it OK if I pull the latest changes into this branch so I can work with the current code? It will make me less confused I think. 

@kczx3 

> these really are more like modules to me than a namespace.

See ➡️ [__Changing "namespace" to "module"__](https://github.com/Lexikos/AutoHotkey_L/pull/162/commits/9df84e97606feb69098880d654237e3b6d09bf9b).

> Are there two ways that you can use a namespace from another file?

No. A file only defines a module if you `#import` it. 

The only change to `#include` is that you can include the same file more than once across different modules, without using `#includeagain`. Eg,
```autohotkey
/* a.ahk:
msgbox 'ok'
*/
module {
	#include a.ahk ; ok
}
module {
	#include a.ahk ; ok
}
#include a.ahk ; ok
#include a.ahk ; must be #includeAgain to do something
```

kczx3:
@HelgeffegleH thanks.

I'm confused on all the different ways to use #include/#import then.  Is the *i required with #import?

myModule.ahk
```
module myModule {
    myGlobal := "myModule::myGlobal"
}
```

main.ahk
```
#import myModule.ahk to x

msgbox(x.myGlobal)
```

Running main.ahk with your latest x64 build results in:

```
---------------------------
AutoHotkey_64.ahk
---------------------------
Error at line 1.

#Import file "C:\Users\user\Google Drive\AHK\AHK-modules build\myModule.ahk to x" cannot be opened.

The program will exit.
```

kczx3:
> I'd really like to see the ability for a library author to _choose_ what they export to the consumer. Right now, I guess that could be done using Closure's but it'd be nice for a namespace to be able to explicitly export things from a namespace. All the rest would be private.

Seems as though all variable declarations in `module A` are private unless they are declared with `global`.  Is that correct?

HelgeffegleH:
@kczx3,

note ➡️ [**Changing macro SMODULES_INCLUDE_DIRECTIVE_FILE_MODULE_SEP to _T("as")**](https://github.com/Lexikos/AutoHotkey_L/pull/162/commits/1263d92d87ebfcced2bf5dd5d7277e78cf9f4abb)

It is now `#Import file.ahk as myModule` . Also, with your code, it should be
```autothotkey
#import myModule.ahk as x

msgbox(x.myModule.myGlobal)
```
You can think of `#import x as y`  as being effectively equivalent to
```
module y {
    #include x
}
```
When you write a library, you should in general not use `module myLib {}` around it, imo. Instead just write it in a file and then let the user either `#include` or `#import` it as it like. I expect _module blocks_ will mostly be used for small _help modules_ or perhaps auto-execute code. Eventually it will be used for compiled scripts too.

You can only resolve super-globals (variables declared global in global scope). _Normal globals_ are not accessible.

Cheers.

kczx3:
> Normal globals are not accessible.

Is there a reason for this?  I can't find one.  It would be nice to have a way to export a variable from a module _without_ needing to make it a super-global or having to create a getter closure.

Lexikos:
There seems to be a bug in the sm_rb branch (not sure about the other one). This code somehow only passes one parameter to FileAppend:

    FileAppend s "`n", "*"

> Again, is it OK if I pull the latest changes into this branch so I can work with the current code? 

Yes.

HelgeffegleH:
> Is there a reason for this?

> It would be nice to have a way to export a variable from a module without needing to make it a super-global

I think it is easy to replace `global` causing variables to be _exportable_ with an actual `export` declaration if we want something like that (it should apply to functions too then). In the mean time, `global` has this double purpose, which I can agree isn't ideal. But I think it is better than _exporting_ all globals (declared or not). Typically, you want to export classes (which are super-globals) but not _for-loop variables_ used in auto-exec code. There is also the fact  that you can create global variables dynamically, these wouldn't be included by `usevar *all` which might be surprising, this could be changed but I see no particular reason to that. I also don't have the mental capacity to think about it at this point 😐 .

> There seems to be a bug in the sm_rb branch

That is a very helpful observation. It seems I messed `sPrecedence` in `ExpressionToPostfix` when I rebased, there were conflicts there because I added and removed `SYM_MODULE`. I somehow missed `SYM_SUPER`, so I ended up with a missing `,0`. I'll try to take better care next time 👍 .

Cheers.

kczx3:
@HelgeffegleH Sounds good re:globals.  I guess I misunderstood that using `global` in a module didn't create a super-global.  So it appears a super-global cannot be created from within a module.  Is that correct?

The global keyword within a module instead allows it to be accessible outside the module.  Agreed that functions should really have this ability as well.

HelgeffegleH:
Declaring a vairable `global` in a module's _global space_ causes it to become __super-global__ in that module, and any module which _imports_ it with. `#usevar`. _Scope resolution_ and `#usevar` can only reference __super-globals__. Example,

```autohotkey
module m1 {
	global v1 := 1
	f()  {
		msgbox v1
	}
	f ; 1
}

#usevar v1 in m1
g(){
	msgbox v1
}
g ; 1
global v2 := 1
module m2 {
	msgbox v1 ; warn useunset / blank
	msgbox v2 ; warn useunset / blank
}
```

Cheers.

Lexikos:
> But I think it is better than exporting all globals (declared or not).

When you say "declared or not", I think you mean "declared *at the module level* or not". `Global x` inside a function still declares x, but doesn't mark it as super-global and therefore doesn't export it from the module. At the moment an implicit "global" variable is hardly global at all; it is a module-level variable which (I think) is not visible anywhere other than the module body by default (as in, without additional declarations).

I have been reconsidering how globals work for a number of reasons, the most recent being the removal of global hotkey subroutines (which prompted the removal of gosub and therefore all global subroutines except auto-execute) and your namespace/module proposals. I'd rather not get into it in detail here, but basically, I think the current scheme (of implicit globals) is less relevant now, and ineffective as a measure of protection due to the use of super-globals.

I think that if temporary variables are needed in the auto-execute portion of a non-trivial script, a function should be used. As an alternative to defining and immediately calling a function, we could add an optional initialization function for the main script and each module.  I suppose that basically only saves a single call to that function (and provides uniformity), but for the main script, it can replace `A_Args`. (I've already implemented `Main(args*)` in a test branch to feel it out as a replacement for `A_Args`.)

Another solution might be to provide a way to create block-local variables without defining a function, but I haven't worked out suitable syntax (and instead choose to defer it indefinitely).

When I say "the main script and each module", what I really mean is "each module". The main script should be a module like any other. It should be possible to load two independent scripts as modules, without any chance of naming conflicts, or any change to how each script refers to itself or its own globals. I'm thinking along these lines:

  - Each module may be nested or top-level.
  - Top-level modules may contain a directive or declaration defining the default module name. Something like `#Module MyScript` or `module MyScript:`, which must precede any other code (excluding comments) if it is present. This shows that the file is intended to be a module, and allows the code to rely on it actually being a (top-level) module.
  - To permit multiple top-level modules in the same file (or for preprocessed scripts), the module declaration could support an additional keyword or name`.` prefix to declare a top-level module.
  - A top-level module (and any modules nested inside it) may use the name which the module defines for itself even if a different name was specified by #Import. This might be useful, for instance, in force-local functions, for referring to module-level variables without a prior declaration inside the function. At least, it would be if a force-local function could refer to the module without first declaring it.
  - `#Import` loads a file (once per file across the entire program) as a top-level module, and associates it with a name in the current module; either the name specified in the #Import directive, or the default name for the module being imported, or perhaps the filename as a fallback. Any name specified with the #Import directive is only an alias, used only in the module which contains that directive.
  - `Outer` would be valid only in nested modules, if at all.
  - Nested modules may be intended as part of the outer module, to be used together with the outer module and other modules nested in that module. As such, module-level names defined in the outer module would be automatically visible, consistent with functions. There can be provision for a sort of force-local module, if needed, although I think that's basically the same as a top-level module (see above).
  - Imported modules are protected from unintended interaction with the script's "global" variables, because they are not nested inside the script module. Perhaps other modules would not be accessible by default; instead, each module would #Import the modules it needs (and if already loaded, they would only be referenced, not loaded again). 

> Typically, you want to export classes (which are super-globals) but not for-loop variables used in auto-exec code. 

I have also considered how to make for-loop variable names actually local to the loop, even outside any function, and I think it is feasible.

Lexikos:
The handling of output vars in sm_rb seems to be a bit broken.
  - `Catch v` causes a var-unset warning because ExpressionToPostfix converts v to ARG_TYPE_INPUT_VAR.
  - Arbitrary expressions such as `(v)` and `v:=0` are permitted/not detected, and cause crashes.

Currently there's a check to skip postfix conversion for ACT_FOR when infix is a single SYM_VAR. I see no reason to give ACT_FOR special treatment - I suppose it would apply equally to all output var args. The simplest solution is probably:
```diff
-	if (mActionType == ACT_FOR
-		&& infix_count == 1
-		&& infix->symbol == SYM_VAR
-		&& aArg.type == ARG_TYPE_OUTPUT_VAR)
+	if (aArg.type == ARG_TYPE_OUTPUT_VAR)
 	{
+		if (infix_count == 1 && infix->symbol == SYM_DYNAMIC && infix->var->IsReadOnly())
+			return VarIsReadOnlyError(infix->var, Script::INVALID_OUTPUT_VAR);
+		else if (infix_count != 1 || infix->symbol != SYM_VAR)
+			return LineError(ERR_EXPR_SYNTAX, FAIL, aArg.text);
-		// This var is an output variable for ACT_FOR. No further processing needed:
+		// This var is an output variable.  No further processing needed:
```

Prior to commit 760094e, output vars (excluding FOR) were handled by ExpressionToPostfix *after* postfix conversion, which allowed things like `Catch (v)` (since parentheses are included in infix but not postfix).

> Catch: OutputVar may optionally be enclosed in parentheses

But that's not working in v2.0-a116. That's fixed by 42c20859. The parentheses are removed at an early stage, so there's no need for infix to postfix conversion.

HelgeffegleH:
I changed `#useXXX` to `useXXX` and [changed the behaviour of `usefunc`](https://github.com/Lexikos/AutoHotkey_L/pull/162/commits/1ae3514795b74ff47920c83833a78ac4dc8b09fe). I did similar work for `usevar` but rolled it back since there where so many changes to variables from a115 to a122, the code wasn't suitable. This is now on top of a115 (I think). I looked into merging a122 but it is a bit of a pain, I'm not sure when I'll find the time to do it (if anyone else wants to do it, feel free 👍 😉 ). 

> The handling of output vars in sm_rb seems to be a bit broken.

Thank you, I have noted your comment but didn't look into it yet. 

Cheers.

HelgeffegleH:
This code is too much out of date, both with respect to implementation and design. A module implementation should be reworked from scratch. I will close this PR for those reasons.

I hope this have at least resulted in some ideas for the future.

PR #163, opened 2020-04-01
Fixing typo in DefineClassVars.

HelgeffegleH:
...to require the end of `__Init` body to fulfill both conditions, rather than either one.

Issue example,

```autohotkey
class x {
	a:=()=>0
	b:=()=>0
	; derefrence nullptr, g->CurrentFunc
} 
```

### DIFF
--- a/source/script.cpp
+++ b/source/script.cpp
@@ -6514,7 +6514,10 @@ ResultType Script::DefineClassVars(LPTSTR aBuf, bool aStatic)
 		{
 			// __Init method already exists, so find the end of its body.
 			for (block_end = init_func->mJumpToLine;
-				block_end->mActionType != ACT_BLOCK_END || !block_end->mAttribute;
+															// The end of the body is a line with:
+				block_end->mActionType != ACT_BLOCK_END		// mActionType == ACT_BLOCK_END
+				&& !block_end->mAttribute;					// and mAttribute == ATTR_TRUE.
+
 				block_end = block_end->mNextLine);
 		}
 		else


Lexikos:
That is not a typo; merely code not designed to handle arrow functions.

Your comment is correct:
> // The end of the body is a line with:
> // mActionType == ACT_BLOCK_END
> // **and** mAttribute == ATTR_TRUE.

However, that condition would be reflected by `! (block_end->mActionType == ACT_BLOCK_END && block_end->mAttribute == ATTR_TRUE)` or `block_end->mActionType != ACT_BLOCK_END || block_end->mAttribute != ATTR_TRUE`, which has the same effect as the original code.

Your code stops on the first line with a non-null mAttribute, *regardless of action type*. For the following code, the order of evaluation is **b**, **d**, **c**, **a**, because each arrow method has an ACT_BLOCK_BEGIN with non-null mAttribute.
````
class x {
	a:=()=>0
	b := 0
	c:=()=>0
	d := 0
}
o := x.new()
````
Methods do not put their blocks within the __init method, so do not need to be considered. Arrow methods do, because they are handled recursively, before `g->CurrentFunc` is reset to `NULL`.

Normally an __init method would only contain expressions, but it can contain a normal block (with `mAttribute == NULL`) or nested function if the script explicitly defines the method before any initializers. In that case, the design is intended to dumbly append the initializer at the end of the method (not at the end of its first block).


The simplest solution seems to be to change ACT_BLOCK_END's mAttribute to contain a pointer to the function (as is already done for ACT_BLOCK_BEGIN), and use `block_end->mAttribute != init_func` instead of `!block_end->mAttribute` (with `||`, not `&&`).

HelgeffegleH:
I see, thank you for the explanation and the proper fix 👍 .

Cheers.

PR #164, opened 2020-04-03
Adding StringCaseSense 'Logical' and sort option 'L' for logical sort.

HelgeffegleH:
Uses  ➡️ [`StrCmpLogicalW`](https://docs.microsoft.com/en-us/windows/win32/api/shlwapi/nf-shlwapi-strcmplogicalw).

Reason, the sort option is for convenience and performance, the `stringcasesense` option is mostly for completeness imo.

This doesn't affect `StrCompare`. I guess there are multiple ways it could be done. I think the `CaseSensitive` parameter should be changed to accept any of the `StringCaseSense` options directly, and not rely on `A_StringCaseSense` in any case. 

__Edit:__ I guess that if the `CaseSenseitive` parameter for `StrCompare` was changed as suggested above, it would be better to let `sort` have its own, similar, `CaseSenseitive` parameter instead of the __C, CL, L__ options.

Example,
```autohotkey
str := '
(
string20
3string
20string
2string
string3
st3ring
st20ring
string2
st2ring
)'
msgbox '2string' < '20string'	; 0
a_stringcasesense := 'Logical'
msgbox '2string' < '20string'	; 1
msgbox sort(str, 'L')
/*
output:
2string
3string
20string
st2ring
st3ring
st20ring
string2
string3
string20
*/
```

__Con__: It doesn't consider negative numbers it seems. 

Cheers.

### DIFF
--- a/source/script.h
+++ b/source/script.h
@@ -1207,6 +1207,7 @@ public:
 		if (!_tcsicmp(aBuf, _T("On")) || !_tcscmp(aBuf, _T("1"))) return SCS_SENSITIVE;
 		if (!_tcsicmp(aBuf, _T("Off")) || !_tcscmp(aBuf, _T("0"))) return SCS_INSENSITIVE;
 		if (!_tcsicmp(aBuf, _T("Locale"))) return SCS_INSENSITIVE_LOCALE;
+		if (!_tcsicmp(aBuf, _T("Logical"))) return SCS_INSENSITIVE_LOGICAL;
 		return SCS_INVALID;
 	}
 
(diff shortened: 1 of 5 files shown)


HelgeffegleH:
Adding the L option seems like a poor solution and the A_StringCaseSense setting needs to be discontinued, not extended. I will close this for those reasons, sorry for the mindless noise.

Lexikos:
> Adding the L option seems like a poor solution 

How so?

It is comparable to the N option, which is not replaced by a CaseSense parameter.

HelgeffegleH:
I have an (perhaps irrational) aversion for option strings. I can agree that we don't need a separate `CaseSense` param for `sort`. But I think it would be better to have an option parameter which takes an object. Eg, `sort str, { casesense : 'Logical' }`. Ofc, similar should then be done for the entire _standard lib_.

PR #165, opened 2020-04-03
Changing operator ** to not cast integer operands to float and then c…

HelgeffegleH:
…ast result back to integer for exponents `>= 0`

__Reasons__, allow correct results for all `x**y` (`x,y` integer, `y>0`) which is within the range of **__int64**.

Behaviour of overflow is undefined.

This is an alternative to #119.

Cheers.

### DIFF
--- a/source/script.h
+++ b/source/script.h
@@ -398,6 +398,8 @@ struct ArgStruct
 	ExprTokenType *postfix;  // An array of tokens in postfix order.
 };
 
+__int64 pow_ll(__int64 base, __int64 exp); // integer power function
+
 #define BIF_DECL_PARAMS ResultToken &aResultToken, ExprTokenType *aParam[], int aParamCount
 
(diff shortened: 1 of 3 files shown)


Lexikos:
So are the differences to #119 that:
  - overflow is undefined instead of defaulting to `_I64_MIN`
  - it does not have the "inconsistency" noted below
?

> that is, `0**0.0` is 0.0 on 32 build while 1.0 on 64, `0**0` is 1 on both builds.

HelgeffegleH:
Yes. 

* Overflow is not detected at all in this branch, so code size and performance is improved (vs #119). The underlying algorithm for computing the result is the same, ➡️ [_exponentiation by squaring_](https://en.wikipedia.org/wiki/Exponentiation_by_squaring).
* The inconsistency was removed by #118.

Cheers.

PR #167, opened 2020-04-13
Changing CaseSensitive parameters to not depend on A_StringCaseSense.

HelgeffegleH:
Also adding option __Logical__, __Locale__, __On__ and __Off__.

New:

* __1__ (`true`) and __On__ means `SCS_SENSITIVE`
* __0__, (`false`) and __Off__, means `SCS_INSENSITIVE`
* __Locale__, means `SCS_INSENSITIVE_LOCALE` 
* __Logical__, means `SCS_INSENSITVE_LOGICAL`.

Any other values yields `SCS_INVALID` and should be considered an error.

`StrCompare` now handles `CaseSensitive := "Logical"`, by calling `StrCmpLogicalW`.

`InStr` throws for `CaseSensitive := "Logical"`, alternatively it could mean `SCS_INSENSITIVE`, but it seems more likely that it would be a mistake and hence better to inform the user.

Cheers.

### DIFF
--- a/source/script.h
+++ b/source/script.h
@@ -3414,6 +3414,7 @@ __int64 TokenToInt64(ExprTokenType &aToken);
 double TokenToDouble(ExprTokenType &aToken, BOOL aCheckForHex = TRUE);
 LPTSTR TokenToString(ExprTokenType &aToken, LPTSTR aBuf = NULL, size_t *aLength = NULL);
 ResultType TokenToDoubleOrInt64(const ExprTokenType &aInput, ExprTokenType &aOutput);
+StringCaseSenseType TokenToStringCase(ExprTokenType& aToken);
 IObject *TokenToObject(ExprTokenType &aToken); // L31
 Func *TokenToFunc(ExprTokenType &aToken);
 IObject *TokenToFunctor(ExprTokenType &aToken);
--- a/source/script2.cpp
+++ b/source/script2.cpp
@@ -11900,6 +11900,8 @@ BIF_DECL(BIF_StrCompare)
 	case SCS_SENSITIVE:				result = _tcscmp(str1, str2); break;
 	case SCS_INSENSITIVE:			result = _tcsicmp(str1, str2); break;
 	case SCS_INSENSITIVE_LOCALE:	result = lstrcmpi(str1, str2); break;
+	case SCS_INVALID:
+		_f_throw(ERR_PARAM3_INVALID);
 	}
 	_f_return_i(result);			// Return 
 }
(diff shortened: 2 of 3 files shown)


Lexikos:
*Why?*

Do you have a counter-argument for this?

> // Rather than adding a third value to the CaseSensitive parameter, it obeys StringCaseSense because: [...]
> // 2) It's more friendly for typical international uses because it avoids having to specify that special/third value
> //    for every call of InStr. [...]

HelgeffegleH:
> Why?

It is more intuitive that the value of a parameter determines the behaviour of the function, rather than the parameter together with a global setting. This change makes scripts more maintainable and predictable. I find the behavoiur extremely unintuitive, example,
```autohotkey
msgbox instr('Ää', 'ä') ; 2
stringcasesense true
msgbox instr('Ää', 'ä') ; 1
```
I doubt many even knows about this. I personally struggle to remember this rule, even though I've read it numerous times, I had to look it up again now. I think I get more confused every time I read it, but maybe that is just aging.

> Do you have a counter-argument for this?

I'd have to first guess what it means; I guess it was written by an american, and that a _typical international user_, is anyone who isn't an american. In that case, I am a _typical international user_, and I have never needed this and never saw any example where anyone else did. Disregarding personal experiences, I think it is very unlikely that someone needs the locale mode for every call to `instr`, unless it is a very small script, in which case it is manageable to pass the parameter anyways. __If__ someone needs the locale mode very often and doesn't want to pass the parameter on every occasion, it can use a function, `myInstr(a,b) => instr(a,b,c)`. Future improvements could allow something like `instr(a,b) =>built-in-instr(a,b,c)`.

Cheers.

Lexikos:
> I find the behavoiur extremely unintuitive, example,

I agree that this example is counter-intuitive, but because *case-sensitive* is combined with *locale mode*.  As I said in #168, "Case sensitivity and locale sensitivity should be separate. (For instance, StringCaseSense true shouldn't make case-insensitive comparisons locale-sensitive.)"

I always thought `'Ä' != 'ä'` was counter-intuitive.

> I guess it was written by an american, and that a typical international user, is anyone who isn't an american.

So you're assuming that the developer had no personal experience or contact with users who needed this option? I admit I am ignorant about most locales other than my own (and I'm not American), but I must assume there was a reason the option was added in the first place.

"user" is not what was written in the comment.

> ... never saw any example where anyone else did. 

I would think that if a user finds `StringCaseSense` and it satisfies their need, you wouldn't hear about it.

> I think it is very unlikely that someone needs the locale mode for every call to instr

I don't understand why anyone would want case insensitive comparisons to default to ignoring locale, or in other words, presumably ignore the rules of their own language. It would make the script behaviour more "consistent" between locales, but I fail to imagine (or observe) the real world consequences.

> unless it is a very small script

I do not see how the need could be present for very small scripts and yet not scale to larger scripts.

PR #168, opened 2020-04-13
Adding CaseSensitive parameter to BIF_Sort

HelgeffegleH:
New:
```autohotkey
SortedString := Sort(String [, Options, Callback, CaseSensitive := false])
```
Also removing the __C__ and __CL__ options.

Also adding handling of `SCS_INSENSITIVE_LOGICAL` to `BIF_Sort`. To be useful #167 or something similar would need to be implemented.

I'm not passionate about the parameter order. It would also be possible to join the `Callback` and `CaseSensitive` parameters, since both can't be meaningful at the same time. But I think it is better to have a separate parameter.

Cheers.

### DIFF
--- a/source/script.cpp
+++ b/source/script.cpp
@@ -241,7 +241,7 @@ FuncEntry g_BIF[] =
 	BIFn(SendMessage, 1, 9, BIF_PostSendMessage),
 	BIF1(SetTimer, 0, 3),
 	BIF1(Sin, 1, 1),
-	BIF1(Sort, 1, 3),
+	BIF1(Sort, 1, 4),
 	BIFn(SoundGetInterface, 1, 3, BIF_Sound),
 	BIFn(SoundGetMute, 0, 2, BIF_Sound),
 	BIFn(SoundGetName, 0, 2, BIF_Sound),
--- a/source/script2.cpp
+++ b/source/script2.cpp
@@ -6546,7 +6546,9 @@ BIF_DECL(BIF_Sort)
 
 	// Resolve options.  First set defaults for options:
 	TCHAR delimiter = '\n';
-	g_SortCaseSensitive = SCS_INSENSITIVE;
+	g_SortCaseSensitive = ParamIndexToCaseSense(3);
+	if (g_SortCaseSensitive == SCS_INVALID)
+		_f_throw(ERR_PARAM4_INVALID);
 	g_SortNumeric = false;
 	g_SortReverse = false;
 	g_SortColumnOffset = 0;
@@ -6559,15 +6561,6 @@ BIF_DECL(BIF_Sort)
 	{
 		switch(_totupper(*cp))
 		{
-		case 'C':
-			if (ctoupper(cp[1]) == 'L') // v1.0.43.03: Locale-insensitive mode, which probably performs considerably worse.
-			{
-				++cp;
-				g_SortCaseSensitive = SCS_INSENSITIVE_LOCALE;
-			}
-			else
-				g_SortCaseSensitive = SCS_SENSITIVE;
-			break;
 		case 'D':
 			if (!cp[1]) // Avoids out-of-bounds when the loop's own ++cp is done.
 				break;
--- a/source/script2.cpp
+++ b/source/script2.cpp
@@ -25,6 +25,7 @@ GNU General Public License for more details.
 #include "resources/resource.h"  // For InputBox.
 #include "TextIO.h"
 #include <Psapi.h> // for GetModuleBaseName.
+#include "shlwapi.h" // for StrCmpLogicalW
 
 #include <mmdeviceapi.h> // for SoundSet/SoundGet.
 #include <endpointvolume.h> // for SoundSet/SoundGet.
@@ -6453,7 +6454,7 @@ int SortWithOptions(const void *a1, const void *a2)
 	}
 	// Otherwise, it's a non-numeric sort.
 	// v1.0.43.03: Added support the new locale-insensitive mode.
-	int result = tcscmp2(sort_item1, sort_item2, g_SortCaseSensitive); // Resolve large macro only once for code size reduction.
+	int result = tcscmp3(sort_item1, sort_item2, g_SortCaseSensitive); // Resolve large macro only once for code size reduction.
 	return g_SortReverse ? -result : result;
 }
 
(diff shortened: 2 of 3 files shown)


jeeswg:
# Delimiter/OmitChars parameters

I like what you're trying to do, but believe other parameters are more important for Sort.

If anything, Sort should have Delimiter/OmitChars parameters, similar to StrSplit.

`Sort(String [, Options, Function, Delimiter, OmitChars])`

This would allow you to sort individual characters (no delimiter), or to sort by multi-character delimiters.
This would also allow you to ignore whitespace or other characters, when sorting.

# Other priorities for Sort

When choosing parameters, some considerations:

I outlined the main priorities for Sort, here: [Wish List 2.0 - AutoHotkey Community](https://autohotkey.com/boards/viewtopic.php?f=13&t=36789).

Priorities:
• handle arrays
• stable sort: by default, on/off option
• reverse without sorting: maybe via the existing 'R' option plus a new 'null sort' option ('R' by itself sorts, then reverses)
• no delimiter: sort individual characters (e.g. alphabetise/reverse/shuffle)
• multi-character delimiter: sort by a multi-character delimiter

Also, this bug needs to be fixed:
[minor bug: Sort command's handling of LFs/CRLFs - AutoHotkey Community](https://autohotkey.com/boards/viewtopic.php?f=14&t=59066)

For reference:
The original AHK v1 options: C CL Dx F N Pn R Random U Z \\.

# Alternatives to a CaseSensitive parameter

If we don't add a CaseSensitive parameter, we could add these options (or similar):
C0, CLogical.
Note: C and C0 are already used with hotstrings.

Possibly you could have something like:
CSCS or CA (based on A_StringCaseSense).

We could add options such as:
C0/COff, C1/COn, CLocale, CLogical.
So that you could just use 'C' followed by A_StringCaseSense.

Possibly, one could change Sort to insist that the options are space/tab-separated.

I'm open to having both a D option and a Delimiter parameter.

I don't find `","` very readable, amongst multiple commas, when used in functions, so would consider a new built-in variable: A_Comma.

# Removing any need for A_StringCaseSense

To either remove, or remove the need for, A_StringCaseSense ...

Your plan would be to add CaseSensitive functionality to InStr/Sort/StrReplace/Switch.
That's fine.
But what about the 'in'/'contains' operators?
You'd need 'in'/'contains' functions to explicitly set the case sensitivity for those.

I've suggested:
`if StrEquals(Haystack, NeedleStringOrArray [, CaseSensitive, Delimiter])`
And likewise: StrContains, StrStarts, StrEnds.

# What about 'Numeric'?

If you had a StrEquals function, equivalent to 'in', you'd need a 'Numeric' option:
E.g. `if StrEquals(var, "0x10,0x20,0x30", "N", ",")`
E.g. `if StrEquals(var, [0x10, 0x20, 0x30], "N", ",")`
Sort and Switch would need a Numeric option too.
So you'd have 5 possible options: 0/1/Locale/Logical/N.
So, what you've been calling 'CaseSensitive' would really be 'ComparisonMode' for those functions/statements.

A further point, if Sort accepted 'N' in the proposed 'CaseSensitive' parameter, you'd have to remove it from the Options parameter.

But anyhow, your suggestion would effectively give us:
`Sort(String [, Options1, Callback, Options2])`
Which isn't the most intuitive.
And isn't the best use of an additional parameter.
With the added ambiguity of where to put 'N'.

kczx3:
@jeeswg
“I outlined the main priorities for Sort, here:”

I think it’s important you understand those are YOUR priorities and nothing more.

Lexikos:
I'm not in favour of splitting options out of the *Options* parameter.

> To be useful #167 or something similar would need to be implemented.

My initial impression upon seeing several related PRs (#167, #168, #169, #170, #171) at once, that all (if only at first glance) appeared to assume StringCaseSense should be removed or ignored without explaining why, was not good. Having the work already done and ready to accept (or reject) might make a proposal more agreeable, but to me it felt aggressive. I see that you've provided multiple options ready to go and I appreciate it to some extent, but I can't say that I liked your approach.

At the moment, I think:
  - Case sensitivity and locale sensitivity should be separate. (For instance, `StringCaseSense true` shouldn't make case-insensitive comparisons locale-sensitive.)
  - If a case sensitivity parameter is present, it should permit specifying:
    1) use all default settings,
    2) case sensitive,
    3) case insensitive according to default locale mode,
    4) case insensitive ignoring locale,
    5) case insensitive respecting locale (or a specific locale?).
  - Overriding of the default locale sensitivity should always be explicit, or not occur. In other words, merely specifying that something should be case-insensitive should not override the default locale sensitivity. This applies to parameters, options and `=`/`==`.
  - Putting locale aside, I'm not really opposed to removing the global case sensitive/insensitive setting.

If `=` and `==` always ignore locale, for instance, users who always want the locale to be respected would have to basically avoid those operators.

HelgeffegleH:
> I'm not in favour of splitting options out of the Options parameter. [...] At the moment, I think:

Fair points.

> If `=` and `==` always ignore locale, for instance, users who always want the locale to be respected would have to basically avoid those operators.

I think locale comparison is seldom needed, in rare cases, you'd have to use `StrCompare`.

> My initial impression upon seeing several related PRs (#167, #168, #169, #170, #171) at once, that all (if only at first glance) appeared to assume StringCaseSense should be removed or ignored without explaining why, was not good. 

I think each one of these changes are useful on their own. I didn't _assume_ anything, but it is true that if all were accepted, plus #103 and #109, `StringCaseSense` wouldn't have any use and could be removed unless other uses for it were planned. This was ofc on purpose, although not planned when I did #103.

> Having the work already done and ready to accept (or reject) might make a proposal more agreeable, but to me it felt aggressive. I see that you've provided multiple options ready to go and I appreciate it to some extent, but I can't say that I liked your approach.

I don't know what _aggressive_ means in this context. I could definitely have presented some of these PRs better, I'll think about that in the future.

Cheers.

Lexikos:
> I think locale comparison is seldom needed

When is it necessary to avoid locale comparison?

> I don't know what aggressive means in this context.

Neither do I, now that I'm finished venting.

HelgeffegleH:
> When is it necessary to avoid locale comparison?

Locale comparison is much slower, according to the manual. I guess performance would be the most common case were one might want to avoid the locale mode. Another example, is if you compare [AHK names](https://lexikos.github.io/v2/docs/Concepts.htm#names),

```autohotkey
funcName := (random(0,1) ? func('aä') : func('aÄ')).name
switch funcName 
{
	case 'Aä': msgbox 1
	case 'AÄ': msgbox 2
}
aä()=>0
aÄ()=>0
```
Here it is __necessary__ to have `stringcasesense false` if the locale mode deems `'ä' = 'Ä'` true. This is ofc contrieved and it is not easy to imagine well written code where you'd need to compare _AHK names_ like this. But I suppose you could deal with other _names_ which have this same _property_ as _AHK names_, which you want to sort or compare for some reason.

Cheers.

HelgeffegleH:
> I'm not in favour of splitting options out of the Options parameter.

I'm closing this for that reason.

PR #170, opened 2020-04-15
Changing Switch to not depend A_StringCaseSense

HelgeffegleH:
Always does case sensitive comparison.

Some examples on how to handle this change,

```autohotkey
Switch StrLower(str)
{
	case 'abc': ; ...
	case 'def': ; ...
}

Switch chr
{
	case 'A', 'a': ; ...
	case 'B', 'b': ; ...
}
```

I am aware these examples have limitations.

I'll also make an alternative PR with an added `CaseSensitive` parameter. __Edit:__ See #171.

Cheers.


Lexikos:
This feels more intuitive to me, but it's probably because I've used other scripting languages with case-sensitive switch. It does seem to be more useful (since it's easier and cleaner to StrLower than temporarily change SCS).

kczx3:
Agreed this seems much more intuitive to me.

PR #171, opened 2020-04-15
Adding CaseSensitive parameter to Switch

HelgeffegleH:
New:

```autohotkey
Switch [Exression, CaseSensitive := false]
```
The `CaseSensitive` parameter works as the other `CaseSensitive` paramters, i.e., as for `StrCompare` and `InStr`.

Specifying the `CaseSensitive` parameter while omitting the expression is an error.

This is an alternative to #170, with #167 and small modification, this will not depend on `A_StrinCaseSense`

Cheers.

### DIFF
--- a/source/globaldata.cpp
+++ b/source/globaldata.cpp
@@ -287,7 +287,7 @@ Action g_act[] =
 	, {_T("Catch"), 0, 1, false, NULL} // fincs: seems best to allow catch without a parameter
 	, {_T("Finally"), 0, 0, false, NULL}
 	, {_T("Throw"), 0, 1, false, {1, 0}}
-	, {_T("Switch"), 0, 1, false, {1, 0}}
+	, {_T("Switch"), 0, 2, false, {1, 0}}
 	, {_T("Case"), 1, MAX_ARGS, false, NULL}
 
 	, {_T("Exit"), 0, 1, false, {1, 0}} // ExitCode
--- a/source/script.cpp
+++ b/source/script.cpp
@@ -23,7 +23,6 @@ GNU General Public License for more details.
 #include "application.h" // for MsgSleep()
 #include "TextIO.h"
 
-
 // These are the common pseudo-Funcs, defined here mostly for readability:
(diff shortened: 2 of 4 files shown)


Lexikos:
I must admit this is convenient, although my initial reaction was something akin to irrational disgust. However, I wonder if it can be generalized somehow. My first thought was vaguely something like:

    Switch Expression, Callback

But I realized once it's written out, having two parameters is redundant.

    switch caseValue => caseValue = str  ; case insensitive
    switch caseValue => inStr(caseValue, str)

But then switch would need to differentiate between function objects and other objects/values, so I suppose it needs a keyword or alternative name.

Maybe it should have the CaseSensitive parameter regardless, for ease of use.

HelgeffegleH:
> my initial reaction was something akin to irrational disgust.

😄 Adding a parameter wouldn't have occurred to me, much less happen, if the implementation didn't permit it to be done so easily.

HelgeffegleH:
>`Switch Expression, Callback`

After not thinking about this for two months I randomly concluded that it is a pretty nice idea. I think the callback should receive two parameters `ExprValue, CaseValue`. That way we preserve the useful property of switch which evaluates the expression only once without needing to store the result (in script) and increase flexibility, Eg,

```
switch x . y, (exprValue, caseValue) => exprValue = caseValue ; case insensitive
{
	; ...
}
```

Cheers 🏖️ 🍹

Lexikos:
The merge commit also implements the current convention for CaseSense parameters.

9ff3d2d changes Switch to unconditionally perform string comparison (never numeric comparison) when *CaseSense* is specified, and throw a TypeError if an object is encountered.

While implementing this, I noticed (and took) some opportunities for optimization, all relating to the use of TokensAreEqual, which was a general function not optimized for use by Switch:
  - IsNumeric only needs to be called once for the switch value, but it was being called once for each case.
  - IsNumeric doesn't need to be called for any of the case values if the switch value is non-numeric, but it was being called anyway.
  - There's no need to account for SYM_VAR (by calling TypeOfToken) because of the way ExpandSingleArg works.

I'm shelving the callback idea for now.

PR #173, opened 2020-04-22
Enable Unicode for stdout

HotKeyIt:


Lexikos:
This isn't enabling Unicode so much as it is forcing UTF-8 for the encoding of error output. That's all well and good for your own custom build, where you don't have to care about breaking every tool that expects ANSI.

joedf:
Perhaps I am misunderstanding, but I thought that's why there's an ANSI build. No?
😵🥴

PR #174, opened 2020-05-30
Changing BIF_CallbackCreate to use ValidateFunctor.

HelgeffegleH:
Fixes variadic functors and error message for parameter 3.

Issue example,
```autohotkey
callbackcreate (*)=>'',, 1 ; Prev: "Error:  Invalid number of parameters." Now: works!
```

Cheers.

### DIFF
--- a/source/script2.cpp
+++ b/source/script2.cpp
@@ -15376,12 +15376,6 @@ BIF_DECL(BIF_CallbackCreate)
 	if (!func)
 		_f_throw(ERR_PARAM1_INVALID);
 
-	// Get func.MinParams (if present) for validation and default parameter count.
-	__int64 minparams;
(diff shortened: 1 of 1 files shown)


Lexikos:
I was avoiding the redundant invocation of `MinParams`, which is used if the parameter count is omitted. Also, it looks like you won't be calling ValidateFunctor if the caller specified a parameter count.

Do my latest changes solve your concerns?

HelgeffegleH:
Note that the example above wasn't meant to show that the error message had improved (for me), but that there was an error, which it shouldn't be,

```autohotkey
callbackcreate (*)=>'',, 1    ; a110: "Error:  Invalid callback function." 
                              ; In this branch: returns a pointer.
```

I can see that I wasn't very clear. 

For the sake of completeness, I updated this branch, see the new commit message.

Cheers.

PR #175, opened 2020-06-05
Changing #If and #IfTimeout to #HotIf and #HotIfTimeout

HelgeffegleH:
__Reason__, `#If` is often confused with `if`. `#HotIf` is more _descriptive_. Changed `#IfTimout` to match `#HotIf`.
Less relevant, sometimes also confused with C-like `#if`. If we want something like C's `#if` in the future, `#if` will be available.

Also changed some relevant comments, some might remain.


Lexikos:
A few days ago, I added to my rough notes "rename #if?". They say great minds think alike.

I'll review this before merging it.

PR #177, opened 2020-06-05
Changing hotkeys and hotstrings to not use labels.

HelgeffegleH:
## New

Mulit-line hotkey and hotstring definitions now require a block following their double-colon (`::`). The block defines a function body to execute when the hotkey or hotstring is triggered. No label is defined. Example,

```autohotkey
x::
{
    ; function body to execute
}
::y::
{
    ; function body to execute
}
```
Same line action hotkeys and hotstrings implicitly defines a function,  for example,

```autohotkey
x::action
:X:y::action
```
is for most *action*s equivalent to
```autohotkey
x::
{
    action
}
::y::
{
    action
}
```
These implicitly created functions are named __\<Hotkey\>__. To call a hotkey function without triggering the hotkey, it has to be explicitly named, i.e., use the usual [hotkey functions](https://lexikos.github.io/v2/docs/Hotkeys.htm#Function). Any call to a hotkey callback passes the name of the hotkey/hotstring to the function as the first parameter. For implicitly created functions, the first parameter is automatically added and (cleverly) named `ThisHotkey`. Example,

```autohotkey
x::msgbox ThisHotkey ; x
:X:y::msgbox ThisHotkey ; :X:y
```

### Stacking

Stacking is supported but has been restricted. There can now be no non-blank lines, including `#if`s, in between stacked hotkeys. Stacking above an auto-replace hotstring, remapping or _AltTab_-hotkey is detected as an error. For example, the following is OK,

```autohotkey
::x::
y::
z::msgbox ThisHotkey ; ::x, y or z depending on which triggered
```
but not,
```autohotkey
x::
y::z
```

### OTB

OTB is supported, eg,

```autohotkey
x:: {
    ; code
}
::y:: {
    ; code
} 
```
To remap a key to `{`, escape it, eg, ``x::`{``. To send one literal `{` with an auto-replace hotstring, use `{{}`, eg, `::y::{{}`. Or use __text__ or __raw__ mode. If there is anything except white space following the opening brace, or if the _X_ option is in effect for a hotstring, the brace is never considered an opening block brace.

### Auto-execute section

The implicit `return` before the first hotkey has been removed, consequently execution flows past any hotkey or hotstring definition as it would past any explicit function definition. Example,
```autohotkey
x::y
z:: {
}
msgbox ; shows when the script starts.
```

### Misc.

* `#if` directives can not be defined inside functions (or classes).
* The `Hotkey()` and `Hotstring()` functions can still set a label as callback. I see no reason to keep this.
* `A_ThisHotkey` is kept for use in _label callbacks_ and `#if expr`. 

## Motivation

I believe the problems associated with _global sub-routine_ callbacks are well known and need not be repeated here in excess. In my opinion, these changes preserves the benefits of the current syntax, while eliminating many of its problems. The new behaviour simplifies the auto-execute section and will eliminate many of the cases of, _this code doesn't work when I paste it in my main script_. Obviously, some users will have issues due to not understanding function scope, but they probably doesn't understand the scope of the current label callback system either, and will have problems regardless. Users will benefit from being required to learn function scope early. 

## About the implementation

Not very complete, and in no particular order,

Code for remappings has been simplified by being confined to one block. Remappings define functions doing the same thing as the previous _sub-routines_. For simplicity, the function body consist of only one line, eg, `SetKeyDelay(-1), Send(...)`.

The **_A label must not point to a function._** code persists and this is also detected for implicit hotkey functions. However, I don't think it is needed (other than for sanity reasons, perhaps.). 

I've taken care to allow referring to hotkeys by their _name_ via the `hotkey()` function, for example, the following works as expected, 

```autohotkey
x::a
hotkey '*x',, 'off'
hotkey '*x up',, 'off'		; disable remapping
hotkey '*x', 'myFunc', 'on'	; do something else
hotkey '*x', '*x', 'on'		; enable remapping
hotkey '*x up', '*x up', 'on'
```

This is done by storing the (implicitly named) hotkey functions in `HotkeyVariant::mOriginalCallback`.
I did not write any such code for hotstrings, as it seems this is not supported now. (I guess that is a question)

The new hotkey functions are not stored in the script's main list (`mFuncs`), instead a separate list is used, `mHotFuncs`. They are kept for local variable processing. I suppose we can delete the list after loading, I forgot. 

I didn't put any love into the name, **\<Hotkey\>**. I thought it would be good to have a name for debugging purposes. It could be named after the hotkey which defines it, but that would be confusing when _stacking_ hotkeys, as they call the same function. It could also be left unnamed. The name is set in `Script::CreateHotFunc`.

I can't recall if _the debugger_ would need some changes to adapt to these changes. I have not considered it.

### DIFF
--- a/source/application.cpp
+++ b/source/application.cpp
@@ -1311,7 +1311,8 @@ bool MsgSleep(int aSleepDuration, MessageMode aMode)
 				g.hWndLastUsed = criterion_found_hwnd; // v1.0.42. Even if the window is invalid for some reason, IsWindow() and such are called whenever the script accesses it (GetValidLastUsedWindow()).
 				g.SendLevel = variant->mInputLevel;
 				g.HotCriterion = variant->mHotCriterion; // v2: Let the Hotkey command use the criterion of this hotkey variant by default.
-				hk->PerformInNewThreadMadeByCaller(*variant);
+				hk->PerformInNewThreadMadeByCaller(*variant, hk->mName);
+				
 			}
 
 			// v1.0.37.06: Call ResumeUnderlyingThread() even if aMode==WAIT_FOR_MESSAGES; this is for
--- a/source/hotkey.cpp
+++ b/source/hotkey.cpp
@@ -867,7 +867,7 @@ void Hotkey::TriggerJoyHotkeys(int aJoystickID, DWORD aButtonsNewlyDown)
 
 
 
-void Hotkey::PerformInNewThreadMadeByCaller(HotkeyVariant &aVariant)
+void Hotkey::PerformInNewThreadMadeByCaller(HotkeyVariant &aVariant, LPTSTR aName)
 // Caller is responsible for having called PerformIsAllowed() before calling us.
 // Caller must have already created a new thread for us, and must close the thread when we return.
 {
@@ -946,7 +946,10 @@ void Hotkey::PerformInNewThreadMadeByCaller(HotkeyVariant &aVariant)
 
 	// LAUNCH HOTKEY SUBROUTINE:
 	++aVariant.mExistingThreads;  // This is the thread count for this particular hotkey only.
-	ResultType result = aVariant.mJumpToLabel->ExecuteInNewThread(g_script.mThisHotkeyName);
+
+	ExprTokenType aParams = { aName };
+	ResultType result = aVariant.mJumpToLabel->ExecuteInNewThread(g_script.mThisHotkeyName, &aParams, 1);
+	
 	--aVariant.mExistingThreads;
 
 	if (result == FAIL)
@@ -1061,8 +1064,8 @@ ResultType Hotkey::Dynamic(LPTSTR aHotkeyName, LPTSTR aLabelName, LPTSTR aOption
 		// Currently using Func* cast rather than ValidateFunctor() because of the need to
 		// support UseErrorLevel (which should be removed when the Error model is revised).
 		Func *func = LabelPtr(aJumpToLabel).ToFunc();
-		if (func && func->mMinParams > 0)
-			RETURN_HOTKEY_ERROR(HOTKEY_EL_BADLABEL, ERR_HOTKEY_FUNC_PARAMS, func->mName);
+		if (func && (func->mMinParams > 1 || (func->mParamCount == 0 && !func->mIsVariadic)))
+			RETURN_HOTKEY_ERROR(HOTKEY_EL_BADLABEL, func->mParamCount == 0 ? ERR_PARAM_REQUIRED : ERR_HOTKEY_FUNC_PARAMS, func->mName);
 	}
 
 	bool suffix_has_tilde, hook_is_mandatory;
@@ -1619,6 +1622,7 @@ HotkeyVariant *Hotkey::AddVariant(IObject *aJumpToLabel, bool aSuffixHasTilde)
 	// aJumpToLabel can be NULL for dynamic hotkeys that are hook actions such as Alt-Tab.
 	// So for maintainability and to help avg-case performance in loops, provide a non-NULL placeholder:
 	v.mJumpToLabel = aJumpToLabel ? aJumpToLabel : g_script.mPlaceholderLabel;
+	v.mOriginalCallback = g_script.mLastHotFunc;
 	v.mMaxThreads = g_MaxThreadsPerHotkey;    // The values of these can vary during load-time.
 	v.mMaxThreadsBuffer = g_MaxThreadsBuffer; //
 	v.mInputLevel = g_InputLevel;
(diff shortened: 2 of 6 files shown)


Lexikos:
Excellent! I had been reluctant to change the syntax and default scope of hotkey subroutines, but having it all done and the reasoning presented so well pushes me past that.

It seems that the single-line form has no way to access (non-super) global variables, but maybe that's okay. It's probably better than having the single-line and multi-line forms using different defaults, and I like that the multi-line form is assume-local.

> To send one literal `{` with an auto-replace hotstring, use `{{}`, [...] if the X option is in effect for a hotstring, the brace is never considered an opening block brace.

After talk of OTB, this confused me, until I tested and realized that `::x::{` was an undetected error, and that the `Missing "{"` message you've added was referring to the braces that go around key names, not OTB.

> These implicitly created functions are named <Hotkey>. 

Maybe that's useful, though it seems a bit inconsistent with fat arrow functions. I suppose it's more likely that ListVars will be used in a hotkey function than in a fat arrow function. For DBGp it's not as important, since the client can show the line number.

Perhaps an alternative would be to give both hotkey and fat arrow functions a name based on the line number. Something with the meaning of "anonymous function at line *n*" (although then maybe it needs part of the filename too).

> A_ThisHotkey is kept for use in [...] #if expr.

That's a bother. I think it can be solved by changing #HotIf to generate an anonymous function of the form `ThisHotkey => expr`. With a bit of refactoring, the code from ParseFatArrow could probably be used for both. The HOT_IF_CALLBACK mode already passes the hotkey name. I suppose the HOT_IF_EXPR mode can then be removed.

You have my approval to do this and remove support for labels with Hotkey() and Hotstring().

Lexikos:
> I've taken care to allow referring to hotkeys by their name via the hotkey() function

I didn't catch the implication: it's now possible to restore a hotkey *variant* to its original function, even when there are multiple variants with the same name. However, there's a bug:

    ^1::MsgBox "original"
    !1::MsgBox "other hotkey"
    ^2::Hotkey "^1", hk => MsgBox("modified")
    ^3::Hotkey "^1", "!1"

I would expect <kbd>Ctrl</kbd>+<kbd>3</kbd> to set `^1` to the original function of `!1` (if the variant exists), or throw an exception. (You could find the specified variant, resolving it the same way that you resolve the first parameter, then take its original function.) What actually happens is that `^1` is restored to *its* original function.

Also, I found a bug that is now doubled up. When you use `Hotkey hk, label`, the label is *called*. Label was never intended to be available to the script directly, so its Invoke implementation just executes the label without doing any checking. I assume there is an extra invocation now, because the label is called twice.

HelgeffegleH:
> After talk of OTB, this confused me, until I tested and realized that `::x::{` was an undetected error, and that the `Missing "{"` message you've added was referring to the braces that go around key names, not OTB.

I'm not sure I understand. A script with `::x::{` and no matching `}` will yield: `Missing "}"`. Because it is supposed to be a mulit-line hotstring definition (OTB style), which then misses its closing brace. In order to have an opening brace following a `::hotstring::`, which does __not__ mean the beginning of a block, one have to,

* use __text__ or __raw__ mode, either via `:T:` or `:R:` or `#Hotstring T/R`. For brevity, I didn't consider any other options (except __X__). Although I guess `:SE:x::{` would mean you want to send `{` using `sendevent`, not that you want to define a mulit-line hotstring function. I'd rather ignore __text__ and __raw__ mode than to handle every possible option.
* use the __X__ option, either via `:X:` or `#hotstring X`. This will yield an error because a single line action cannot begin with a literal object defition, eg, `{x:1}` is not valid on its own.
* have something else (non-white space) following the `{` on the same line. For example, `::x::{enter}`, sends __Enter__ and `::x::{{}` sends __{__.

For hotkeys, `trigger::{` is always a multi-line hotkey definition. Except if there is something else (non-white space) following the `{` on the same line. For example, `x::{x:1}` is a line with a literal object, which, again, is not valid on its own.

> That's a bother.

Yes, I was disappointed when I realised this, because removing (the need for/limitation of) `A_ThisHotkey` was part of the _plan_.

> I think it can be solved by changing #HotIf to generate an anonymous function of the form `ThisHotkey => expr`

> You have my approval to do this and remove support for labels with Hotkey() and Hotstring().

I will look into this (eventually), and the bug you mentioned in the second post.

I feel that the documentation on hotkeys/hotstrings needs more than just modification (even before this change). I'll open an issue in the docs repo and ask about this there.

Cheers.

Lexikos:
I'd appreciate a PR for the documentation of these changes, when you have time.

HelgeffegleH:
I started modifying the existing documentation to match the changes, adding the minimum additionally required info and removing outdated info. In the future, more elaborate improvements is probably desired.

However, I realised I have missed a few things,

* `A_EventInfo` for mouse wheel hotkeys.
* `A_EndChar` for hotstrings.
* Maybe something else.

I will investigate if these can easily be passed to the function via an additional parameter. I suppose that there might be some issues with hotkeys/hotstrings recieving extra info vs hotkeys/hotstrings which doesn't.

PR #181, opened 2020-06-13
Removing HOT_IF_EXPR

HelgeffegleH:
`#HotIf expr` now uses `HOT_IF_CALLBACK`, with callback

```autohotkey
<hotkey>(ThisHotkey) {
    return expr
}
```

The function is assume local for consistency with hotkeys. 

__Reason__, `HOT_IF_CALLBACK` can pass parameters. Removes the need (and limitation) for `A_ThisHotkey`.

## About

`#HotIf var` is not optimized for brevity, questionable benefit and for consitent behaviour. Example,

```autohotkey
; Behaviour if it was optimised (as before):
var := 'abc'
#hotIf var      ; var is seen, 'abc' is considered true.
x::msgbox var   ; var is not seen
#hotIf 'abc'    ; never true (must be numeric)
```

`LPTSTR HotkeyCrieterion::OriginalExpr` is added for storing the original expression, so it can be found again. Alternatively, I guess, `HotkeyCriterion::ExprLine` could be kept for `HOT_IF_ACTIVE/EXIST` and `WinTitle` could be used for `HOT_IF_CALLBACK` (added via `#hotIf`). But it seems more maintainable and readable to have only one (clear) way of finding the original expression. Only downside is one more pointer member per `HotkeyCriterion`, I don't think it matters much.

I did not use fat-arrow since that would require working around fat-arrows using assume global. 

Removed `EvaluateHotCriterionExpression()`, because it is not called anymore.

### DIFF
--- a/source/hotkey.cpp
+++ b/source/hotkey.cpp
@@ -56,7 +56,6 @@ HWND HotCriterionAllowsFiring(HotkeyCriterion *aCriterion, LPTSTR aHotkeyName)
 		found_hwnd = WinExist(g_default, aCriterion->WinTitle, aCriterion->WinText, _T(""), _T(""), false, false); // Thread-safe.
 		break;
 	// L4: Handling of #HotIf (expression) hotkey variants.
-	case HOT_IF_EXPR:
 	case HOT_IF_CALLBACK:
 		// Expression evaluation must be done in the main thread. If the message times out, the hotkey/hotstring is not allowed to fire.
 		DWORD_PTR res;
@@ -104,7 +103,7 @@ HotkeyCriterion *AddHotkeyCriterion(HotCriterionType aType, LPTSTR aWinTitle, LP
 	if (   !(cp = (HotkeyCriterion *)SimpleHeap::Malloc(sizeof(HotkeyCriterion)))   )
 		return NULL;
 	cp->Type = aType;
-	cp->ExprLine = NULL;
+	cp->OriginalExpr = nullptr;
 	if (*aWinTitle)
 	{
 		if (   !(cp->WinTitle = SimpleHeap::Malloc(aWinTitle))   )
@@ -144,6 +143,7 @@ HotkeyCriterion *AddHotkeyIfExpr()
 	if (   !(cp = (HotkeyCriterion *)SimpleHeap::Malloc(sizeof(HotkeyCriterion)))   )
 		return NULL;
 	cp->NextExpr = NULL;
+	cp->OriginalExpr = nullptr;
 	if (g_LastHotExpr)
 		g_LastHotExpr->NextExpr = cp;
 	else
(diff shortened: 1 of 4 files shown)


source/hotkey.cpp
```diff
@@ -155,15 +155,14 @@ HotkeyCriterion *AddHotkeyIfExpr()
...
-			&& !_tcscmp(aExpr, cp->ExprLine->mArg[0].text)) // Case-sensitive since the expression might be.
+	for (HotkeyCriterion* cp = g_FirstHotExpr; cp; cp = cp->NextExpr)
+		if (cp->OriginalExpr && !_tcscmp(aExpr, cp->OriginalExpr)) // Case-sensitive since the expression might be.
 			return cp;
 	return NULL;
 }
 
 
-void Script::PreparseHotkeyIfExpr(Line *aLine)
+void Script::PreparseHotkeyIfExpr(Line* aLine)
```
HelgeffegleH, line 165: @Lexikos.
Visual studio does these changes on its own sometimes, it is not intended by me. I think the most common style is `type *var` vs `type* var`. I certainly miss and/or forget sometimes.

Lexikos:
> The function is assume local for consistency with hotkeys.

A hotkey can potentially assign a global variable by declaring it. #HotIf cannot, so flexibility (or at least convenience) is lost. In order to avoid making the variable accessible everywhere, one has to define a full function and call that from #HotIf. To be reusable the function may have to use dynamic variables, since #HotIf can't pass a global variable by reference.

I think it's far more likely that any variable reference in the expression is intended to be global than local (excluding ThisHotkey). It was the same for global fat arrow functions (which are now essentially obsolete). I also feel that the new behaviour of hotkeys, especially one-line hotkeys, is counter-intuitive, but I think that the issue is more with implicit globals in general.

With all reasons for global subroutines (which would be implicitly creating and using globals) now gone, and modules/namespaces on the way, I think I will need to review how "globals" work.

HelgeffegleH:
> I think it's far more likely that any variable reference in the expression is intended to be global 

Certainly. 

I think that `var` in `x::msgbox var` _looks_ like it is in the global space since it isn't enclosed by any block. I'm not sure if it is unintuitive or if I'm just used to the old ways though. With modules/namespaces, one shouldn't hesitate to declare a variable (super) global, doing it shows that it is used more widely than an _normal_ global. `#HotIf`, and one-line hotkeys, and any other scope for that matter, could access _normal_ globals via something like `global.var`, hypothetically that is. I imagined that with modules/namespaces, one would only use _normal_ globals for _autoexecute_ code. You wouldn't use `global` in functions, other than for perhaps super-globals if you want to use force local. I suppose we would need a way to declare `global partialvar%%` for that to work.

> I think I will need to review how "globals" work.

That is interesting. I think that super-global vs _normal_ global can be a little confusing for beginners, at least it was for me.

Cheers.

PR #182, opened 2020-06-13
Various fixes for #175, #176, #177

HelgeffegleH:
Fixing 
```autohotkey
; Expr line cont.:
#hotif (x,
    y) 
```
and
```autohotkey
hotif 'invalid' ; Parameter #2 must match an existing #If expression.
```
and __KeyHistory__ message,
>  #=Disabled via #IfWinActive/Exist

Also comment fix and variable / parameter names.

See individual commits.


source/globaldata.cpp
```diff
@@ -264,7 +264,6 @@ Action g_act[] =
 	, {_T("{"), 0, 0, false, NULL}, {_T("}"), 0, 0, false, NULL}
 
 	, {_T("Static"), 1, 1, false, {1, 0}} // ACT_STATIC (used only at load time).
-	, {_T("#If"), 0, 1, false, {1, 0}}
```
HelgeffegleH, line 267: It seems this broke the program, I'll fix it

Lexikos:
>  Correcting text in GetHookStatus. 

Good catch. However, it really applies to #HotIf in general, so I've removed "WinActive/Exist".

PR #183, opened 2020-06-25
v2: DriveRetract()

Ragnar-F:
I have never understood why retraction only exists as a parameter.

### DIFF
--- a/source/script.cpp
+++ b/source/script.cpp
@@ -112,7 +112,7 @@ FuncEntry g_BIF[] =
 #ifdef ENABLE_DLLCALL
 	BIFn(DllCall, 1, NA, BIF_DllCall),
 #endif
-	BIFn(DriveEject, 0, 2, BIF_Drive),
+	BIFn(DriveEject, 0, 1, BIF_Drive),
 	BIFn(DriveGetCapacity, 1, 1, BIF_DriveGet),
 	BIFn(DriveGetFilesystem, 1, 1, BIF_DriveGet),
 	BIFn(DriveGetLabel, 1, 1, BIF_DriveGet),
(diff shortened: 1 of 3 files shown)


HelgeffegleH:
Logically it makes more sense 👍 . 

> I have never understood why retraction only exists as a parameter.

I don't feel the word _retract_ is commonly associated with this operation. Usually, you have an _eject_ button, which acts like a toggle (or only eject).  Maybe `DriveClose` is easier to remember / understand / search for.

Ragnar-F:
Sadly I can't say much about this, because English is not my native language, but seeing the MCI commands used in the code also named "open" and "close" for these purposes, DriveOpen() and DriveClose() instead of DriveEject() and DriveRetract() could be a good idea.

jeeswg:
The pair, Eject/Close, I find somewhat inelegant.

Using Open/Close would add ambiguity. In what way are you 'opening'/'closing' the drive, are you locking it? Etc.
Cf. FileOpen/ProcessClose/WinClose.

The original proposal I find to be fine. The only question being: is this overkill, to introduce such a rarely used function.

(Btw Eject/Retract are clearly physical.
This is useful because users may not expect AHK functions to have a **physical** effect.
No other AHK functions have a physical effect.)

jeeswg:
Some possibilities:
```
DriveEject(Drive:="", Mode:=0) ;0/1/-1 for eject/retract/toggle [like AHK v1]
DriveEject(Drive:="", Mode:=1) ;0/1/-1 for retract/eject/toggle [the boolean is opposite to AHK v1]
DriveTrayOpen(Drive:="", Mode:=1) ;0/1/-1 for closed/opened/toggle [refer to state, not action][or another name]
DriveEject(Drive:="", Mode:="") ;blank/R for eject/retract
DriveRetract(Drive:="")
```

In everyday life, the Eject button both ejects and retracts, ergo one function.
That said, 2 separate functions could also work well.
As a native English speaker, Eject/Retract sound quite (i.e. very) natural to me.

Btw the retract functionality doesn't work on my laptop or desktop PC.
A forum poll re. this might be interesting. 'Does DriveEject's retract feature work for you?'

Lexikos:
> I don't feel the word *retract* is commonly associated with this operation. Usually, you have an *eject* button, which acts like a toggle (or only eject).

I was reluctant to merge this because of similar thoughts, but after using it while testing, it seems fairly natural.

I considered adding the "CD" suffix, like DriveGetStatusCD, since these functions did not work with any other drive types.

On the other hand, I compared the code size of the previous approach to one using DeviceIoControl (7f9bf01), and found that the latter was slightly smaller, and also works for other drive types that have the "Eject" option in Explorer (although it's not the same as using Safely Remove Hardware). My research and testing was unable to turn up any reason to use the old approach.

PR #185, opened 2020-06-29
v1: IsSet()

jeeswg:
IsSet for AHK v1.1.

For reference: the v2 commit: [Added IsSet(Var)](https://github.com/Lexikos/AutoHotkey_L/commit/916fe75fb3670a9884276ba2c0f88d39808f6d9f).

Btw, I noticed that in AHK v2, the Exception.What property contains the function name, but not so in AHK v1.1.
E.g. try the Hotstring function.


jeeswg:
It seems it's more common to create a new branch for each PR ... so I'll do that.

PR #187, opened 2020-07-04
Changing SimpleHeap::Malloc to terminate the program on allocation failure

HelgeffegleH:
__New__

Use `SimpleHeap:Malloc` (and do not verify the return value) when it is acceptable that the program terminates on failure. Else call `SimpleHeap::Alloc`, and verify the return value.

__About__

Calls `g_script.TerminateApp(EXIT_DESTROY, CRITICAL_ERROR)` to terminate the script. `Malloc` _shouldn't_ be called during run time.

__Reasons__

Brevity, convenience, maintainability and presumably (future?) code size.

__Misc__

Also fixes potential undefined behaviour,

* in `CreateClass/DefineMembers` which didn't verify return from `operator new`.
* when `DefineMethod` is called via `CreateRootPrototypes`.

I'll change the method names `Malloc/Alloc` if desired.

### DIFF
--- a/source/SimpleHeap.cpp
+++ b/source/SimpleHeap.cpp
@@ -24,7 +24,34 @@ SimpleHeap *SimpleHeap::sLast  = NULL;
 char *SimpleHeap::sMostRecentlyAllocated = NULL;
 UINT SimpleHeap::sBlockCount = 0;
 
+
+// Malloc never returns NULL. If allocation fails a message is shown
+// and the program terminates. Typically called only during load time.
+
 LPTSTR SimpleHeap::Malloc(LPTSTR aBuf, size_t aLength)
+{
+	LPTSTR ret = Alloc(aBuf, aLength);
+	if (!ret)
+		g_script.TerminateApp(EXIT_DESTROY, CRITICAL_ERROR); // Alloc has already shown the error message.
+	return ret;
+}
+
+void* SimpleHeap::Malloc(size_t aSize)
+{
+	void* ret = Alloc(aSize);
+	if (!ret)
+	{
+		g_script.ScriptError(ERR_OUTOFMEM);
+		g_script.TerminateApp(EXIT_DESTROY, CRITICAL_ERROR);
+	}
+	return ret;
+}
+
+// Alloc can return NULL. Callers should verify the return value.
+// Typically used during run time, when an allocation failure shouldn't
+// terminate the program, cf Malloc.
+
+LPTSTR SimpleHeap::Alloc(LPTSTR aBuf, size_t aLength)
 // v1.0.44.14: Added aLength to improve performance in cases where callers already know the length.
 // If aLength is at its default of -1, the length will be calculated here.
 // Caller must ensure that aBuf isn't NULL.
@@ -34,7 +61,7 @@ LPTSTR SimpleHeap::Malloc(LPTSTR aBuf, size_t aLength)
 	if (aLength == -1) // Caller wanted us to calculate it.  Compare directly to -1 since aLength is unsigned.
 		aLength = _tcslen(aBuf);
 	LPTSTR new_buf;
-	if (   !(new_buf = (LPTSTR)SimpleHeap::Malloc((aLength + 1) * sizeof(TCHAR)))   ) // +1 for the zero terminator.
+	if (   !(new_buf = (LPTSTR)SimpleHeap::Alloc((aLength + 1) * sizeof(TCHAR)))   ) // +1 for the zero terminator.
 	{
 		g_script.ScriptError(ERR_OUTOFMEM, aBuf);
 		return NULL; // Callers may rely on NULL vs. "" being returned in the event of failure.
@@ -46,7 +73,7 @@ LPTSTR SimpleHeap::Malloc(LPTSTR aBuf, size_t aLength)
 	return new_buf;
 }
 
-void* SimpleHeap::Malloc(size_t aSize)
+void* SimpleHeap::Alloc(size_t aSize)
 // This could be made more memory efficient by searching old blocks for sufficient
 // free space to handle <size> prior to creating a new block.  But the whole point
 // of this class is that it's only called to allocate relatively small objects,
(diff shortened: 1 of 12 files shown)


Lexikos:
This is one of the maintenance items I've been deferring.

I think the undefined behaviour you mentioned is not undefined. `new XXX()` in modern versions of Visual C++ always throws an exception if the allocator returns null unless you use `(std::nothrow)`. I've had fixes on my TODO list for years, but never cared enough to prioritize it, so either kept with the old pattern or just ignored the possibility of failure (as you've noticed in the functions you mentioned). It effectively just terminates the program anyway.

The fix mostly just relates to reduced code size and increased maintainability - i.e. not checking for `new` failure (but still checking for malloc/realloc failure).

Lexikos:
So there's a new function SimpleHeap::Alloc that has the old behaviour consistent with malloc, and SimpleHeap::Malloc has new behaviour inconsistent with malloc...

How do you decide when to terminate and when to check for failure?

My idea was to terminate for basically anything small, because by the time small allocations fail, either the process is almost completely out of virtual address space (and it's 32-bit), or the system has used all of its RAM and most of its pagefile (or has no pagefile). I think the latter case might never happen, because the system terminates processes to free up memory. For either case (if the script isn't terminated), there's a fair chance that not only the current operation will fail, but the script will basically be unable to execute (error handling or subsequent threads) and will just encounter a series of failures.

Allocations of large or unknown amounts (like for FileRead or VarSetStrCapacity) should throw a script exception, because there's a much better chance the script is actually capable of continuing (and that it's a product of an error in the script, or something unexpected that isn't total resource exhaustion). Generally malloc is used for these cases, so there's no need for change.

What is the purpose of the added null checks in the functions that construct the built-in object prototypes? If these fail, the program should terminate. (And it will, because `new` will throw `bad_alloc`, not return `nullptr`.) When you mentioned fixing undefined behaviour, I assumed you meant by having the program terminate instead of continuing under the assumption that the allocation succeeded (even though it wouldn't actually continue). Even if it could continue, it's probably better to dereference the null pointer and terminate (from the access violation exception) than to continue with the prototype or class objects being null.

Another thing I think we should do while we're changing these parts of the code - replace code like
```C++
= (LPTSTR)SimpleHeap::Malloc((this_new_arg.length + 1) * sizeof(TCHAR));
= (DerefType*)SimpleHeap::Malloc((deref_count + 1) * sizeof(DerefType));
```
with something like
```C++
= SimpleHeap::Malloc<TCHAR>(this_new_arg.length + 1);
= SimpleHeap::Malloc<DerefType>(deref_count + 1);
```
The parameter default can be 1, so
```C++
= (HotkeyVariant *)SimpleHeap::Alloc(sizeof(HotkeyVariant))
```
becomes
```C++
= SimpleHeap::Alloc<HotkeyVariant>()
```

> return sAnyPrototype; // out of mem, the program should fail soon somewhere else.

I think it's more reliable to dereference the null pointer. But I also think that if FindFunc fails (due to a memory allocation failure), the program should terminate. But I also think the chance of that happening at this stage (when CreateRootPrototypes is called) is vanishingly small.

HelgeffegleH:
> How do you decide when to terminate and when to check for failure?

The changes was supposed to preserve the current behaviour, i.e,, allocation failure during load time terminates, but during run time an exception is thrown. So `Malloc` should only be used during load time, and `Alloc` during run time. Some code run both on load and run time, then use `Alloc`. `SimpleHeap` is rarely used during runtime, so I thought this would be managable. But,

> My idea was to terminate for basically anything small

this is probably simpler and more maintainable 👍 .

> Another thing I think we should do

I agree.

> What is the purpose of the added null checks in the functions that construct the built-in object prototypes?

> I think the undefined behaviour you mentioned is not undefined. new XXX() in modern versions of Visual C++ always throws an exception if the allocator returns null unless you use (std::nothrow).

I added code like this,

```c++
auto func = new BuiltInMethod(SimpleHeap::Malloc(full_name));
if (!func)
	return nullptr;
```

becuase `new` is overloaded, and uses `SimpleHeap::alloc`, which can return `NULL`. I couldn't find any documentation suggesting that `operator new` will throw if you return `nullptr` from it. I tested `void* operator new(size_t s) { return nullptr; }` and it returns `nullptr` without calling the constructor. I agree that there should be no null checks when using `::new`.

> When you mentioned fixing undefined behaviour, I assumed you meant by having the program terminate instead of continuing under the assumption that the allocation succeeded

For the same reason you mentioned,

> I also think the chance of that happening at this stage (when CreateRootPrototypes is called) is vanishingly small.

I didn't find it worth to add code to do anything else than return, I assumed the program would find its way out somewhere else pretty soon anyways, and long before these objects would be used anyways (I could ofc be mistaken).

> So there's a new function SimpleHeap::Alloc that has the old behaviour consistent with malloc, and SimpleHeap::Malloc has new behaviour inconsistent with malloc...

The names isn't very good. Hence why I mentioned that I'll change them on request. Which also was meant to imply that I had no good suggestions myself. I lack imagination.

Cheers.

Lexikos:
>  I couldn't find any documentation suggesting that operator new will throw if you return nullptr from it. I tested void* operator new(size_t s) { return nullptr; } and it returns nullptr without calling the constructor.

That makes sense now that you say it, and now I understand how (easily) `new (std::nothrow) x` would be implemented. There was a disconnect in my mind between the definition of the overloaded operator and the actual call, since obviously the overloaded operator does not call the constructor or return a pointer to the object. It didn't occur to me that the part being overloaded is the part that normally throws, not the part that calls the constructor.

> I didn't find it worth to add code to do anything else than return

But I don't see why you even do that much, especially given that it doesn't offer an improvement in behaviour. Why not just rely on the allocation function to terminate the program?

> The names isn't very good.

If you at least swap the names, then
  - Malloc does what it always did, and no one who already knew what it did will be surprised.
  - Alloc will be a new function, so anyone who's already familiar with Malloc will know something new is happening.
  - The diff will not show changes to lines that called Malloc, so maybe there's less to review, and `git blame` might be more useful.
  - The diff will show changes to lines that now call Alloc, but maybe that was already the case, due to the removal of nullptr checks.

Lexikos:
> Some code run both on load and run time, 

When we have the capability to eval code or dynamically load modules, virtually all load time code will also be used at run time. It might already be the case with AutoHotkey_H. The reason I haven't just added eval and similar functionality (like HotKeyIt has) is that I think memory management should be reviewed first to permit script fragments to be freed/unloaded after use or if an error occurs parsing it.

PR #188, opened 2020-07-04
Renaming some objects with "label" in their name.

HelgeffegleH:
__New__

* `class LabelRef` renamed to `IObjectRef`.
* `class LabelPtr` renamed to `IObjectPtr`.
* all `aJumpToLabel` renamed to `aCallback`.
* all `mJumpToLabel` renamed to `mCallback`.

__About__

`LabelPtr/Ref` and `mJumpToLabel` was renamed by VS' __Rename__ feature. `aJumpToLabel` was renamed by VS' __Find and replace__. Also removed `LabelPtr::ToLabel()`.

__Reason__

These objects doesn't reference labels.

### DIFF
--- a/source/application.cpp
+++ b/source/application.cpp
@@ -1257,7 +1257,7 @@ bool MsgSleep(int aSleepDuration, MessageMode aMode)
 			case AHK_INPUT_END:
 			{
 				ExprTokenType param = input_hook->ScriptObject;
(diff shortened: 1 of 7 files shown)


Lexikos:
I had already removed the Label functionality but haven't renamed things yet. I've generally been deferring maintenance that doesn't affect functionality. I'll see whether this merges when I'm finished writing documentation.

PR #190, opened 2020-07-05
v1: StrCount()

jeeswg:
`Count := StrCount(Haystack, Needle [, CaseSensitive := false, StartingPos := 1])`

Parameters based on InStr.
Code based on StrReplace's count functionality.
StartingPos is handled as in AHK v2. (I.e. AHK v2: -1 is the last character, AHK v1: 0 is the last character.)

## Error Handling
In AHK v2, InStr (currently) does not throw when given a non-numeric or 0 for StartingPos.
This function does not throw either. Although I'd be happy if throw support were added.

### DIFF
--- a/source/script.cpp
+++ b/source/script.cpp
@@ -8661,6 +8661,12 @@ Func *Script::FindFunc(LPCTSTR aFuncName, size_t aFuncNameLength, int *apInsertP
 		min_params = 0;
 		max_params = 3;
 	}
+	else if (!_tcsicmp(func_name, _T("StrCount")))
+	{
+		bif = BIF_StrCount;
+		min_params = 2;
+		max_params = 4;
+	}
 	else
 		return NULL; // Maint: There may be other lines above that also return NULL.
 
--- a/source/script.h
+++ b/source/script.h
@@ -3356,6 +3356,7 @@ BIF_DECL(BIF_Trim); // L31: Also handles LTrim and RTrim.
 
 BIF_DECL(BIF_Hotstring);
 BIF_DECL(BIF_InputHook);
+BIF_DECL(BIF_StrCount);
 
 
 BIF_DECL(BIF_IsObject);
--- a/source/script2.cpp
+++ b/source/script2.cpp
@@ -18790,6 +18790,57 @@ BIF_DECL(BIF_Exception)
 
 
 
+BIF_DECL(BIF_StrCount)
+{
+	// Load-time validation has already ensured that at least two actual parameters are present.
+	TCHAR needle_buf[MAX_NUMBER_SIZE];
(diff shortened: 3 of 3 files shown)


jeeswg:
I'm fine with this PR being rejected.
For maintenance/documentation reasons, I'd prefer that StrCount use the internal StrReplace functionality directly.
And have either 3 or 4 parameters, like so:
`count := StrCount(haystack, needle, casesen:=0)`
`count := StrCount(haystack, needle, casesen:=0, startpos:=1)`

I collected some reasons for a separate StrCount function:

clean code:
it's expressive/readable, it 'says what it does'
you can easily count how often you use it

clean code+:
using StrReplace to count can be obscure if the variable name lacks the word 'count':
```
StrReplace(vListRGB, ",", "", vImgW), vImgW += 1
v.
vImgW := StrCount(vListRGB, ",") + 1
```

newbies:
'does StrReplace with no return value modify the string or not'
'if I replace it with itself, it won't change anything' [you could end up doing a case insensitive replacement, e.g. 'Abc' to 'abc']
'to count, do I replace with itself, or with nothing, or what, which is faster?'
'if you copy and paste 'var := StrReplace(var, ..., count)', but forget to remove the output variable ...'
'don't have to remember: StrReplace, then what? ...' [StrCount is simple enough to type from memory]

avoid verbosity:
```
linecount := StrCount(text, "`n") + 1
v.
StrReplace(text, "`n", "", linecount)
linecount++
```

avoid verbosity+:
case sensitivity is awkward in AHK v1:
(string count case sensitive)
(string count guaranteeing case insensitive)
e.g.:
```
scs := A_StringCaseSense
StringCaseSense, On
StrReplace(text, "A", "", count)
StringCaseSense, % scs
v.
count := StrCount(text, "A", 1)
```

avoid verbosity++:
it can be more handy to retrieve a value directly (e.g. combinatorics algorithms):
```
if (StrCount(var, "a") > 2)
|| (StrCount(var, "b") > 2)
|| (StrCount(var, "c") > 2)
|| (StrCount(var, "d") > 2)
	continue
v.
...
```

it gets used a lot:
e.g. as a progress counter (current/total), when parsing lines
e.g. to prepare a string/array capacity
```
linecount := StrCount(text, "`n") + 1
```

it plays nicely when working with multiple languages (more translatable):
other languages may have something similar to StrCount built-in
some languages lack ByRef parameters

I intended StrCount and StrRept to be a pair, like Sign and Abs (anyway, that's what I hoped for)

code is more consistent across forum examples and libraries:
and obscure/slower workarounds are discouraged, e.g.:
```
var := "abc_abc_abc_abc_"
MsgBox, % StrSplit(var, "abc", "_").Length() - 1
```

PR #191, opened 2020-07-05
v1: StrJoin()

jeeswg:
`NewStr := StrJoin(PadStrings, String1 [, String2, ..., StringN])`

Code based on StrSplit.

## Examples
`MsgBox, % StrJoin(",", "a", "b", "c", "d", "e", "f") ;a,b,c,d,e,f`

`oArray := ["a", "b", "c", "d", "e", "f"]`
`MsgBox, % StrJoin([":", ","], oArray*) ;a:b,c:d,e:f`

## Error Handling
This function treats missing array values as blank strings.
Although I'd be fine with treating this as an error.
[EDIT:] It appears that when an array is used for param 1, any keys without a value are treated as though they don't exist. And that if the array is empty, the function silently fails.

### DIFF
--- a/source/script.cpp
+++ b/source/script.cpp
@@ -8661,6 +8661,12 @@ Func *Script::FindFunc(LPCTSTR aFuncName, size_t aFuncNameLength, int *apInsertP
 		min_params = 0;
 		max_params = 3;
 	}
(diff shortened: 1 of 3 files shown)


jeeswg:
Replaced by #215.

PR #192, opened 2020-07-05
v1: StrRept()

jeeswg:
`NewStr := StrRept(String, Count)`

Repeat a string n times.
Code appends text via repeated doubling, then appends the remainder.

## Examples
`MsgBox, % StrJoin("abc_", 4) ;abc_abc_abc_abc_`

## Alternative approach
A slightly less fast, but built-in approach:
The custom FastStrRepeat function, [here](https://autohotkey.com/boards/viewtopic.php?f=5&t=71658), uses wcsncat.
However, the ANSI equivalent, strncat, appears to have a bug.
So I'd suggest wcsncat_s and strncat_s.


jeeswg:
Replaced by #216.

PR #195, opened 2020-07-05
Removing Gosub

HelgeffegleH:
__Reason__, `gosub` has no benefits over using functions. Imo, `gosub` has been completely unnecessary since the addition of nested functions. The last imaginable excuse for its existence might have been to _call_ hotkey labels, which now isn't possible.

Also updated comments and removed `global_struct::CurrentFuncGosub`.

Cheers.

### DIFF
--- a/source/Debugger.h
+++ b/source/Debugger.h
@@ -278,7 +278,7 @@ public:
 	}
 
 	
-	// Stack - keeps track of threads, function calls and gosubs.
+	// Stack - keeps track of threads and function calls.
 	DbgStack mStack;
 
 private:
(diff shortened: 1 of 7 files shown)


iseahound:
I disagree. As an automation based language, labels are central to the first step of modeling a process routine. If there is a long sequence of keystrokes and events, labels chop up the sequences in the routine into manageable bits. 

Therefore if there is confusion as to how long a routine should be, labels and gosub solve the issue. 
```
if WinExist('some_App')
    GoSub 'Step_1'
else
    GoSub 'Step_2'


Step_1:
Send '^c'
Sleep 20
Send '{Down}'
Step_2:
Send '{Tab}'
Sleep 20
Send '^v'
return
```

The fact that labels, gosub, and goto are not in the final product are a testament to our programming ability - we've extracted the core routines and finalized them as functions. 

Just because GoSub is simple is not a good reason to remove it.

Lexikos:
I had already done this locally, but withheld it because it's taking a while to update the documentation.

PR #198, opened 2020-08-01
Disabling __try __except in MainExecuteScript for debug builds.

HelgeffegleH:
__Reason__, when debugging in VS the message box in the `__except` block is less helpful than VS' exception info.

For example, run this in db,

```c++
BIF_DECL(BIF_Abs)
{
	auto x = aParam[111111111];
```
VS will point to the above line if we remove the `__try` block, else you get the access violation message box.


Lexikos:
See 522e83ee.

PR #199, opened 2020-08-02
Changing Array::GetEnumItem to skip SYM_MISSING

HelgeffegleH:
__Reason__ convenience.

Usually you only want the values in the array, with this change you don't have to ask `aArray.has()`. Use `loop aArray.length` if you want to loop over the array's entire range of valid indices.

### DIFF
--- a/source/script_object.cpp
+++ b/source/script_object.cpp
@@ -2000,8 +2000,15 @@ Array::index_t Array::ParamToZeroIndex(ExprTokenType &aParam)
 
 ResultType Array::GetEnumItem(UINT &aIndex, Var *aVal, Var *aReserved)
 {
-	if (aIndex < mLength)
+	while (aIndex < mLength)
 	{
+		auto &item = mItem[aIndex];
+		if (item.symbol == SYM_MISSING)
+		{
+			// Skip missing indices.
+			++aIndex; 
+			continue;
+		}
 		if (aReserved)
 		{
 			// Put the index first, only when there are two parameters.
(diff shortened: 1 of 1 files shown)


Lexikos:
The original object enumerator would have already skipped the missing items, but the Array enumerator specifically and intentionally enumerates exactly *Length* times, over array elements, including elements that have no value. As I haven't ever (as yet) been inconvenienced by the new behaviour, I don't think I'll be reversing that choice.

Why would you have sparse items in an array, only to ignore them?

This seems to take the opposite direction to #200.

HelgeffegleH:
> Why would you have sparse items in an array, only to ignore them?

Consider this,
```
myFunc(p*) {
	for v in p
		; assume v is not an empty string
}
a :=  []
a.length := 5
loop 5 {
	a[a_index] := ...
	if ?
		break
}
myFunc a*
```
I think this code is plausible. So I will need to change `myFunc` to,
```
myFunc(p*) {
	for v in p {
		if v != '' { ; or p.has(a_index)
			; the stuff
		}
	}
}
```
which is a bit annoying.

> This seems to take the opposite direction to #200.

How so? Both branches removes the _SYM_MISSING -> empty string_ behavioiur. For _constency_ I suppose `Array::GetEnumItem` could set the __value__ variable to __unset__ instead. But then we need to use `IsSet` in the loop example above. 

Cheers.

Cheers.

Lexikos:
You specify that there will be 5 items, but then only assign n<5 items, and expect the script to treat the array as having n items despite the length showing otherwise.

If the array is intended to have only the number of items actually assigned, the script should use Push rather than assigning the Length in advance (or it could correct the Length prior to breaking). If it is intended to have 5 items, it makes little sense to enumerate fewer items by default.

Length is not intended to be an *estimate* of the number of items in the array. If people will use it that way due to the bound-checking, perhaps assignment should be permitted to grow the array (as I originally intended, but iirc changed at your suggestion). Alternatively, perhaps Capacity should be used as an indication of what range of indices may be assigned rather than purely for optimization.

> How so?

#200 includes unassigned items in the sequence, while #199 excludes them. I'd call those opposites.

PR #200, opened 2020-08-02
Adding IsUninitializedNormalVar() check in Array::FromEnumerable

HelgeffegleH:
__Reason__, to allow `not_an_array*` to _omit_ the output variable for consistency with `aArray*`.

Example,
```autohotkey
f(a:=1,b:=2)=>msgbox(a b)
i:=1
f ((byref o)=>i==1?i--:!i?(o:=3,--i):0)*
``` 
which currently shows a warning for an unnamed global variable, and passes a blank string for parameter `a`. In this branch it yields the same result as `f [,3]*`

### DIFF
--- a/source/script_object.cpp
+++ b/source/script_object.cpp
@@ -366,7 +366,10 @@ Array *Array::FromEnumerable(ExprTokenType &aEnumerable)
 		if (result != CONDITION_TRUE)
 			break;
 		ExprTokenType value;
-		var.ToTokenSkipAddRef(value);
+		if (var.IsUninitializedNormalVar())
+			value.symbol = SYM_MISSING;
+		else
+			var.ToTokenSkipAddRef(value);
 		vargs->Append(value);
 	}
 	var.Free();


Lexikos:
If one iteration assigns a value and the next iteration does not, the detection will fail, because the variable will not be uninitialized.

I have an unused branch with code similar to this but also:
  - changing GetEnumItem to free aVal if the token is SYM_MISSING, and
  - changing CallEnumerator to Free the variables between iterations (`Free(VAR_NEVER_FREE, true)`, so any allocated memory is reused, but it is null-terminated and marked uninitialized).

I've forgotten why I didn't use it. I might have thought it wasn't useful enough to justify the (very little) added code, documentation, and overhead. (In theory, explicitly freeing and then assigning the variables adds more overhead than simply allowing the assignment to do it.) On the other hand, I didn't delete the branch, so I may have been waiting for a more compelling reason to implement it.

HelgeffegleH:
In #199 I wrote,

>  removes the SYM_MISSING -> empty string behavioiur

I suppose this issue will resolve itself when the use of the unset variable becomes an error rather than a warning.

> _If one iteration assigns a value..._

I'll close this for these reasons.

PR #203, opened 2020-12-17
Enable SimpHeap to create larger memory Blocks and fix FileObject.RawWrite for Buffer object.

HotKeyIt:
Removes limitation for lines and probably others


Lexikos:
I do not think the added complexity is warranted. For large blocks of memory, there is relatively little benefit to using SimpleHeap. I chose a simpler solution (0ae12156).

Also, as a general rule, do not combine changes in one PR unless they are related or interdependent. Although in this case, the second change would not merge anyway, since you have based it on an obsolete commit. aa99b990 fixes the RawWrite issue.

PR #206, opened 2021-04-23
v2: Added StrTitle(). Removed 'T' param from StrLower/StrUpper.

jeeswg:


tigerlily-dev:
I was just thinking about this lately - wouldn't it make more sense to remove `StrUpper` and `StrLower` and just have something like `StrCase()` function? It would function just like `StrUpper/StrLower`, Except you would have options of `U`, `L`, and `T` for upper case, lower case, title case. Other options could be added in the future like camel case, etc.

example:
`str := "HaLlO pEoPlE!"
StrCase(str, "L)`

output:
`hallo people!`

jeeswg:
Good points.
My line of thinking had been: StrXXX for the 3 commonly-used cases, and StrCase(str, mode) for any rarely-used cases.
The names StrUpper/StrLower are clear, and you avoid typing double quotes, or squinting at the 2nd parameter.

Lexikos:
@tigerlily-dev 
There are already *two* functions which can format a string as lower, upper or title case depending on a parameter:

    MsgBox Format("{:L}", "HaLlO pEoPlE!")
    MsgBox RegExReplace("HaLlO pEoPlE!", ".*", "$T0")

Both of these functions are much more versatile than StrCase, but less readable than StrLower/Upper/Title when used for this simple task. StrCase would sit somewhere in the middle, not being the best in any respect.

PR #207, opened 2021-04-23
v2: Added StrCount().

jeeswg:
Replaces #190. This PR is for v2, whereas the previous PR was for v1.

`Count := StrCount(Haystack, Needle [, CaseSensitive := false, StartingPos := 1])`

Parameters match InStr.

These should be equivalent:
`Count := StrCount(Haystack, Needle)`
`StrReplace(Haystack, Needle,,, &Count)`

As should these:
`Count := StrCount(Haystack, Needle, CaseSense)`
`StrReplace(Haystack, Needle,, CaseSense, &Count)`

### DIFF
--- a/source/script.cpp
+++ b/source/script.cpp
@@ -313,6 +313,7 @@ FuncEntry g_BIF[] =
 	BIF1(StatusBarGetText, 0, 5),
 	BIF1(StatusBarWait, 0, 8),
 	BIF1(StrCompare, 2, 3),
+	BIF1(StrCount, 2, 4),
 	BIFn(StrGet, 1, 3, BIF_StrGetPut),
 	BIF1(StrLen, 1, 1),
 	BIFn(StrLower, 1, 2, BIF_StrCase),
--- a/source/script.h
+++ b/source/script.h
@@ -3266,6 +3266,7 @@ BIF_DECL(BIF_String);
 BIF_DECL(BIF_StrLen);
 BIF_DECL(BIF_SubStr);
 BIF_DECL(BIF_InStr);
+BIF_DECL(BIF_StrCount);
 BIF_DECL(BIF_StrCase);
 BIF_DECL(BIF_StrSplit);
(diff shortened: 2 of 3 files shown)


jeeswg:
Any feedback would be most welcome.

Is it simply that you view `StrReplace` as sufficient?

To me this seemed like a classic uncontroversial workhorse function, and one of my most commonly used functions e.g. 45th in a collection of key scripts and libs:
```
193	PostMessage
189	Trim
189	ControlGetText
186	Round
181	Floor
172	Map
166	JEE_StrCount
151	WinGetTitle
148	ToolTip
147	JEE_StrStarts
137	JEE_SplashText
132	ControlGetHwnd
```
`StrCount` is used in 18 of my key function libraries.

In short, this seemed overcomplicated, and difficult to remember, every time I wanted to do a string count:
```
StrReplace(Haystack, Needle,,, &Count)
StrReplace(Haystack, Needle,, CaseSense, &Count)
```

Of course not every common function I use will be a built-in function in AHK, but it slows you down sharing scripts, if multiple times when you do something basic, you have to convert the code before you share it, or share a custom function to avoid rewriting your code.
... I tried to find the optimum selection of missing functions and propose those as PRs.

And then: 'write the code, write the shareable version, edit the shareable version, edit the original'. That adds unnecessary complication and duplication.

I had thought about 'just using `StrReplace`', every time, but, using `StrCount` is more readable/maintainable, easier to write from memory, and more portable to other programming languages.

Some simple uses beyond those mentioned in #190:
- count LFs [line count]
- count LFs/CRs/CRLFs count [nonstandard EOL check]
- count items in a comma-separated list
- count dots in a floating-point number (as string)
- count colons in timestamps
- count backslashes [dir depth]
- count slashes [URL depth]
- one-line if-statement cf. using `StrReplace`

Lexikos:
I don't see any value in having it, or debating it with you.

PR #208, opened 2021-04-23
v2: Changed InStr Occurrence parameter. Improved InStr error handling.

jeeswg:
If Occurrence is positive, the search is left-to-right. StartPos does not determine the direction.
If Occurrence is negative, the search is right-to-left. StartPos does not determine the direction.
If Occurrence is omitted, the search direction is based on the sign of StartPos. This is unchanged.

If Occurrence is negative, and StartPos is omitted, StartPos is set to -1.

You can easily do repeated right-to-left searching. You can now provide a positive StartPos, and search right-to-left.
You can easily start at the nth-to-last character, and search left-to-right.

### DIFF
--- a/source/script2.cpp
+++ b/source/script2.cpp
@@ -11610,36 +11610,46 @@ BIF_DECL(BIF_InStr)
 	LPTSTR haystack = ParamIndexToString(0, _f_number_buf, (size_t *)&haystack_length);
 	size_t needle_length;
 	LPTSTR needle = ParamIndexToString(1, needle_buf, &needle_length);
-	
+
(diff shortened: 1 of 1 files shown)


Lexikos:
There are some issues with this code:
  - There are redundant optional parameter checks and default value assignments, which make it harder to follow.
  - The offset calculations have unnecessary redundancy and (for RTL mode) complication.
  - `InStr(a, b,, StrLen(a) + n)` starts the search outside (to the right of) haystack when `n > 0`.
  - `InStr(a, b,, -(StrLen(a) + 1), 1)` starts the search outside (to the left of) haystack.

But the idea is reasonable, so I have implemented it in f6f6574.

PR #210, opened 2021-04-23
v1: Added StrSplitLen().

jeeswg:
`Array := StrSplitLen(String, Length)`
Separates a string into an array of substrings based on the Length parameter.
The final substring may be shorter than Length.

The function name, StrSplitLen, can be changed if desired.
The handling for blank strings, can be changed if desired. As with StrSplit, if String is blank, an empty array is created.


Lexikos:
I don't think this warrants a built-in function.

PR #213, opened 2021-04-23
v1: Added StrEquals/StrContains/StrStarts/StrEnds functions.

jeeswg:
## Introduction

```
ItemNum := StrEquals(Haystack, NeedleStringOrArray [, CaseSense := false])
ItemNum := StrContains(Haystack, NeedleStringOrArray [, CaseSense := false])
ItemNum := StrStarts(Haystack, NeedleStringOrArray [, CaseSense := false])
ItemNum := StrEnds(Haystack, NeedleStringOrArray [, CaseSense := false])
```
If a needle array is passed, the function returns the 1-based index of the first needle that matches the haystack, else 0.
If a needle string is passed, the function returns 1 if the needle matches the haystack, else 0.

## Some benefits

Some benefits versus existing functionality:
- `StrEquals` is more intuitive than `StrCompare`, and can handle multiple needles.
- `StrContains` can handle multiple needles, unlike `InStr`.
- `StrContains` can accept a blank needle, unlike `InStr` (in AHK v2).
- A minor point. When checking if one string contains another, `StrContains` returns 1 or 0, whereas `InStr` would need `!!` to achieve the same effect.
- `StrStarts`/`StrEnds` are more expressive and readable, and less bug-prone, than `SubStr` comparisons.
-  `StrStarts`/`StrEnds` are more reusable than `SubStr` comparisons: simpler to write and copy-and-paste, as string length values aren't needed.
- Versus `RegExMatch`, no character escaping is necessary.
- The `in`/`contains` AHK v1 pseudo-operators use special handling for commas, which is unusual, breaking the principle of least astonishment, and could cause bugs.
- The `in`/`contains` AHK v1 pseudo-operators don't let you specify case sensitivity.
- (It's easy to recreate functions in other programming languages, not so much with operators.)

## Numeric comparison

Potentially, in future, `CaseSense` parameters could accept 'N' to enable a numeric comparison for `StrEquals`/`StrCompare`.

## Comma-separated lists

I also use 4 additional functions, with a 'CSL' suffix, that treat `NeedleString` as a comma-separated list. The 'CSL' and non-'CSL' functions both handle `NeedleArray` identically.

The 4 'CSL' functions use the following logic:
```
StrContainsCSL(ByRef vText, ByRef oNeedles, vCaseSen:=0)
{
	if IsObject(oNeedles)
		return StrContains(vText, oNeedles, vCaseSen)
	return StrContains(vText, StrSplit(oNeedles, ","), vCaseSen)
}
```

They are not implemented in this PR, but are sufficiently useful that they may be worth considering as built-in functions, possibly with a different suffix than 'CSL'.

(The top 11 functions that I use, are in order: `'RetX', SubStr, InStr, 'StrEqualsCSL', DllCall, StrReplace, StrLen, SendInput, Chr, MsgBox, 'GetSelectedText'`. I use `RetX` instead of `return` to record hotkey usage data, then exit the thread. My `OnHotkey` proposal would allow me to use `return` again.)

If desired, special handling for double commas could be used as with the `in`/`contains` pseudo-operators in AHK v1, or some other handling. I am neutral re. special handling.

## tcscmpn2 macro

This PR contains a `tcscmpn2` macro, to compare n characters of a string, for use with `StrStarts`.
It complements the `tcsstr2` and `tcscmp2` macros.

The order of the macro parameters can be changed if desired.

## What about 'if array in/contains string'?

My current view is to be against this.

Logically, the 7 key scenarios are:
```
'equals':
if str1 equals str2

'contains':
if str_long contains/starts/ends str_short
e.g. does 'abc' contain 'a'/'b'/'c'
e.g. does 'abc' start with 'a'/'b'/'c'

'within':
if str_short within/at-the-start-of/at-the-end-of str_long
e.g. is 'a' within 'abc'/'def'/'ghi'
e.g. is 'a' at the start of 'abc'/'def'/'ghi'
```

The 4 functions handle this:
`if haystack contains/starts/contains needle_list`
I suppose:
`if haystack_list contains/starts/contains needle`
Would be equivalent to this (the 'within' functionality mentioned above):
`if needle within/at-the-start-of/at-the-end-of haystack_list`

Personally, I have only wanted 'if array contains/starts/contains string' functionality about 3 times, using a loop with 'if string operator string' was fine.
Also, I would like to be able to easily find such instances in code, retrospectively, hence I would want separate function names. I implemented a custom function `StrWithin`, but only used it once.

## What about 'if array in/contains array'?

Again, my current view is to be against this.

Re. 'if arrayA contains arrayB', I suppose this would tell you that at least one item in arrayB is a substring of at least one item in arrayA. This is somewhat odd. And what return value to use?

If people wanted something like 'if arrayA in arrayB', some kind of `Intersection`/`Union` functions/methods might be better.

## What about in/contains operators?

I believe that `StrEquals` and `StrContains` are sufficient, but would not oppose `in`/`contains` operators.

A complicating factor is: would `in`, when checking 2 values for equality, treat all values as strings, or take type into account.
Perhaps `contains` would compare everything as strings.
Re. type, I proposed that `CaseSense` parameters accept 'N', for numeric comparisons.

Also, would the operators be case sensitive? Yes, I suppose, as with `switch`.

## Implementation

I would hope that these functions or similar be added at some stage. And am happy for a debate to take place on the details.
After a decade of toing and froing, this is what I myself use.

The functions aim to handle the `CaseSense` parameter similarly to AHK v2.
Although it may be preferable to replace this handling with a backport of AHK v2's `ParamIndexToCaseSense` macro.
That macro would need some modifying to work in AHK v1.

## Test code

```
;test code: StrEquals/StrContains/StrStarts/StrEnds (AHK v1)

MsgBox, % StrEquals("abc", "abc") ;1
MsgBox, % StrContains("abc", "b") ;1
MsgBox, % StrStarts("abc", "a") ;1
MsgBox, % StrEnds("abc", "c") ;1

MsgBox, % StrEquals("abc", "xxx") ;0
MsgBox, % StrContains("abc", "x") ;0
MsgBox, % StrStarts("abc", "x") ;0
MsgBox, % StrEnds("abc", "x") ;0

MsgBox, % StrEquals("abc", ["xxx", "abc"]) ;2
MsgBox, % StrContains("abc", ["x", "b"]) ;2
MsgBox, % StrStarts("abc", ["x", "a"]) ;2
MsgBox, % StrEnds("abc", ["x", "c"]) ;2

MsgBox, % StrEquals("abc", ["xxx", "xxx"]) ;0
MsgBox, % StrContains("abc", ["x", "x"]) ;0
MsgBox, % StrStarts("abc", ["x", "x"]) ;0
MsgBox, % StrEnds("abc", ["x", "x"]) ;0

/*
;custom functionality not in PR:

MsgBox, % StrEqualsCSL("abc", "xxx,abc") ;2
MsgBox, % StrContainsCSL("abc", "x,b") ;2
MsgBox, % StrStartsCSL("abc", "x,a") ;2
MsgBox, % StrEndsCSL("abc", "x,c") ;2

MsgBox, % StrEqualsCSL("abc", "xxx,xxx") ;0
MsgBox, % StrContainsCSL("abc", "x,x") ;0
MsgBox, % StrStartsCSL("abc", "x,x") ;0
MsgBox, % StrEndsCSL("abc", "x,x") ;0
*/
```

### DIFF
--- a/source/script.cpp
+++ b/source/script.cpp
@@ -8661,6 +8661,15 @@ Func *Script::FindFunc(LPCTSTR aFuncName, size_t aFuncNameLength, int *apInsertP
 		min_params = 0;
 		max_params = 3;
 	}
+	else if (!_tcsicmp(func_name, _T("StrEquals"))
+		|| !_tcsicmp(func_name, _T("StrContains"))
+		|| !_tcsicmp(func_name, _T("StrStarts"))
+		|| !_tcsicmp(func_name, _T("StrEnds")))
+	{
+		bif = BIF_StrMatch;
+		min_params = 2;
+		max_params = 3;
+	}
 	else
 		return NULL; // Maint: There may be other lines above that also return NULL.
 
--- a/source/script.h
+++ b/source/script.h
@@ -3356,6 +3356,7 @@ BIF_DECL(BIF_Trim); // L31: Also handles LTrim and RTrim.
 
 BIF_DECL(BIF_Hotstring);
 BIF_DECL(BIF_InputHook);
+BIF_DECL(BIF_StrMatch);
 
 
 BIF_DECL(BIF_IsObject);
--- a/source/script2.cpp
+++ b/source/script2.cpp
@@ -18791,6 +18791,77 @@ BIF_DECL(BIF_Exception)
 
 
 
+BIF_DECL(BIF_StrMatch)
+{
+		TCHAR mode = ctoupper(aResultToken.marker[4]); // Get 5th char of function name.
+		aResultToken.symbol = SYM_STRING;
+		aResultToken.marker = _T("");
+
+		StringCaseSenseType string_case_sense = SCS_INSENSITIVE;
+		if (!ParamIndexIsOmitted(2))
+		{
+			TCHAR scs_buf[MAX_NUMBER_SIZE];
+			LPTSTR scs = ParamIndexToString(2, scs_buf);
+			string_case_sense = Line::ConvertStringCaseSense(scs);
+			if (string_case_sense == SCS_INVALID)
+				_f_throw(ERR_PARAM3_INVALID);
+		}
+
+		int aNeedleCount = 0;
+		LPTSTR *aNeedleList = NULL;
+		if (Object *obj = dynamic_cast<Object *>(TokenToObject(*aParam[1]))) // Needle object.
+		{
+			aNeedleCount = obj->GetNumericItemCount();
+			aNeedleList = (LPTSTR *)_alloca(aNeedleCount * sizeof(LPTSTR *));
+			if (!obj->ArrayToStrings(aNeedleList, aNeedleCount, aNeedleCount))
+				// Array contains something other than a string.
+				_f_throw(ERR_PARAM2_INVALID);
+		}
+		else
+		{
+			aNeedleList = (LPTSTR *)_alloca(sizeof(LPTSTR *));
+			*aNeedleList = TokenToString(*aParam[1]);
+			aNeedleCount = 1;
+		}
+
+		LPTSTR haystack = ParamIndexToString(0, aResultToken.buf);
+		int len_haystack = _tcslen(haystack);
+		int result = 0;
+		if (mode != 'Q') // If not StrEquals, disallow blank needles.
+		{
+			for (int i = 0; i < aNeedleCount; ++i)
+			{
+				if (!*aNeedleList[i])
+					_f_throw(ERR_PARAM2_INVALID);
+			}
+		}
+		for (int i = 0; i < aNeedleCount; ++i)
+		{
+			LPTSTR needle = aNeedleList[i];
+			int len_needle = _tcslen(needle);
+			if (len_haystack < len_needle)
+				continue;
+			else if (mode == 'Q') // StrEquals.
+				result = !tcscmp2(haystack, needle, string_case_sense);
+			else if (mode == 'O') // StrContains.
+				result = !!tcsstr2(haystack, needle, string_case_sense);
+			else if (mode == 'T') // StrStarts.
+				result = !tcscmpn2(haystack, needle, string_case_sense, len_needle);
+			else if (mode == 'N') // StrEnds.
+				result = !tcscmp2(haystack+len_haystack-len_needle, needle, string_case_sense);
+
+			if (result)
+			{
+				result = i + 1;
+				break;
+			}
+		}
+		aResultToken.symbol = SYM_INTEGER;
+		aResultToken.value_int64 = result;
+}
+
+
+
 ////////////////////////////////////////////////////////
 // HELPER FUNCTIONS FOR TOKENS AND BUILT-IN FUNCTIONS //
 ////////////////////////////////////////////////////////
--- a/source/util.h
+++ b/source/util.h
@@ -621,6 +621,8 @@ inline LPTSTR HwndToString(HWND aHwnd, LPTSTR aBuf)
 // returns 0 on failure, but failure occurs only when parameter/flag is invalid, which should never happen in
 // this case.
 #define lstrcmpni(str1, len1, str2, len2) (CompareString(LOCALE_USER_DEFAULT, NORM_IGNORECASE, str1, (int)(len1), str2, (int)(len2)) - 2) // -2 for maintainability
+#define tcscmpn2(str1, str2, string_case_sense, len) ((string_case_sense) == SCS_INSENSITIVE ? _tcsnicmp(str1, str2, len) \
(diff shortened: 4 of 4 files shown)


Lexikos:
There appears to be no need to duplicate the implementation of `ConvertStringCaseSense` in this function, but you have done that.

I suppose that some more code duplication can be eliminated by having the second (non-object) path produce a single-element array.

jeeswg:
I recently ported `StrSplit` (and strings/maths/dates/sort functions) to JavaScript, and noticed you did something similar.
I thought it might slow down the code, but perhaps it's negligible.

I can modify this PR, and force push an update, but my hunch is it would be better if you merged the PR and modified/optimised it to your liking!

Regarding direct replacements for `in` and `contains`, `contains` is easily done via `RegExMatch`, but not `in`, *maybe* a function form of `in` (working name: `StrEqualsCSL`), is worthwhile.

Lexikos:
> StrContains can accept a blank needle, unlike InStr (in AHK v2).

This is not a benefit. It's not that InStr *can't* accept a blank needle, but that it *doesn't* accept a blank needle. It's more likely that the caller passed a blank value by mistake than with the intent of having the starting position returned to them.

> A minor point. When checking if one string contains another, StrContains returns 1 or 0, whereas InStr would need !! to achieve the same effect.

!! is only needed to coerce a value to 1 or 0, which is unnecessary in the majority of cases, including "when checking if one string contains another".

jeeswg:
> StrContains can accept a blank needle, unlike InStr (in AHK v2).
> This is not a benefit.

Over 10 years, I'm used to the functions accepting a blank string as a valid needle.
You are welcome to change this. (For `StrContains`/`StrEnds`/`StrStarts` and potentially even `StrEquals`.)
What I had in mind, was the ambiguity that arises if `InStr`/`StrContains` were allowed to accept blank needles.
`InStr` returns a character *position* index or 0.
`StrContains` returns an array *item* index or 0.
```
InStr("abc", "",, startpos:=10) ;high ambiguity (position returned) [perhaps return 1, perhaps return 10, both have drawbacks]
InStr("abc", "") ;some ambiguity (position returned) [perhaps return 1]
StrContains("abc", "") ;no ambiguity (1 or 0 returned) [return 1]
StrContains("abc", ["", "a", "b"]) ;no ambiguity (1-based array index or 0 returned) [return 1]
```

> !! is only needed to coerce a value to 1 or 0, which is unnecessary in the majority of cases

I often write functions  to parse an `Options` function parameter, e.g. `vDoRecurse := !!InStr(vOpt, "R")`.
That said, `StrContains` is longer than `!!InStr`.
And admittedly, `vDoRecurse` need only be 'truthy', not necessarily 1, but then again, I sometimes port code to other languages, debugging is fractionally simpler (1 versus a random index), and clarity is gained since it's made a clear a 'bool' is wanted, not a position index.

Lexikos:
> InStr returns a character position index or 0. Possible ambiguity.

InStr gives a definite result that has only one meaning. When you specify an empty needle, in v1 the result is always 1. The non-zero result means that the string was found. An empty string can always be found at every position. Extracting a substring of the same length as the needle (0) at that position will always yield a value equal to the needle.

> StrContains returns an array item index or 0. No ambiguity.

Does it return the index of the empty string because the empty string is always found, or does it return 0 (or ignore the empty string) because no one really wants to search for the empty string?

There is nothing more or less ambiguous about either one. In either case, there is no ambiguity once the behaviour is defined. In either case, returning any value is questionable. The only safe behaviour is to throw an exception, as InStr does in v2.

jeeswg:
## An update

It now throws if a blank needle is found, for 3 of the 4 functions (not `StrEquals`).
`StrContains`/`StrStarts`/`StrEnds` throw if `NeedleStringOrArray` contains a blank needle.

Implementation:
- An 8-line if-condition was added. No other code was changed.
- It checks each needle *before* the for loop.
- If you checked *during* the for loop, you might get a matching needle before you saw a blank needle., thus the blank needle wouldn't be spotted (and the function wouldn't throw).

## `InStr` versus `StrContains`

Here's a simpler explanation for my rationale:

`InStr`:
- Does it contain a blank string? [Yes. Arguably all strings contain a blank string.]
- At what position is that blank string? [In particular, what should `InStr("", "",, 10)` return? In AHK v1 it returns 1.]
- How do we handle a `StartPos` beyond the string?

`StrContains`:
- Does it contain a blank string?  [Yes.]
- Position isn't checked.
- `StartPos` isn't specified.

E.g. Edge cases (AHK v1):
```
MsgBox(InStr("abc", "",, 3)) ;3
MsgBox(InStr("abc", "",, 4)) ;4 [4 is outside 'abc']
MsgBox(InStr("abc", "",, 5)) ;0
```

For reference, JavaScript's  `includes`/`startsWith`/`endsWith` all allow blank needle strings.

## Notifications

Clipboard append bug in AHK v1, not in v2. It has existed since AHK Basic. #235.
`ControlExist` shouldn't throw rationale. #244.

PR #218, opened 2021-04-25
v2: Fixed IsDigit/IsXDigit misidentifying digits/hex digits.

jeeswg:
A cast to UCHAR causes IsDigit/IsXDigit to misidentify digits/hex digits.
This same problem also exists in AHK v1, and is fixed by #219.

Some code demonstrating the problem:
```
;test IsDigit and IsXDigit:

;test IsDigit:
vOutput := ""
vCount1 := vCount2 := vCount3 := 0
Loop 65535
{
	vChar := Chr(A_Index)
	vCount1 += IsDigit(vChar)
	vCount2 += Test_IsDigit(vChar)
	vCount3 += Test_IsDigit(vChar, 1)
	if Test_IsDigit(vChar)
		vOutput .= vChar
}
A_Clipboard .= "`r`n" vOutput
MsgBox(vCount1 "`r`n" vCount2 "`r`n" vCount3) ;3328 313 3328 [313 is correct]

;test IsXDigit:
vOutput := ""
vCount1 := vCount2 := vCount3 := 0
Loop 65535
{
	vChar := Chr(A_Index)
	vCount1 += IsXDigit(vChar)
	vCount2 += Test_IsXDigit(vChar)
	vCount3 += Test_IsXDigit(vChar, 1)
	if Test_IsXDigit(vChar)
		vOutput .= vChar
}
A_Clipboard .= "`r`n" vOutput
MsgBox(vCount1 "`r`n" vCount2 "`r`n" vCount3) ;5632 44 5632 [44 is correct]
return

Test_IsDigit(vText, vDoBug:=0)
{
	static A_IsUnicode := 1
	static vFunc := A_IsUnicode ? "msvcrt\iswdigit" : "msvcrt\isdigit"
	static vType := A_IsUnicode ? "UShort" : "Int"
	Loop Parse, vText
	{
		vChar := A_LoopField
		vOrd := Ord(vChar)
		if vDoBug
			vOrd := vOrd & 0xFF ;this recreates the UCHAR cast bug: if (!_istdigit((UCHAR)*cp))

		if !DllCall(vFunc, vType,vOrd, "Cdecl")
			return 0
	}
	return 1
}

Test_IsXDigit(vText, vDoBug:=0)
{
	static A_IsUnicode := 1
	static vFunc := A_IsUnicode ? "msvcrt\iswxdigit" : "msvcrt\isxdigit"
	static vType := A_IsUnicode ? "UShort" : "Int"
	if (SubStr(vText, 1, 2) = "0x")
[...]

PR #219, opened 2021-04-25
v1: Fixed 'var is digit/xdigit' misidentifying digits/hex digits.

jeeswg:
A cast to UCHAR causes 'var is digit/xdigit' to misidentify digits/hex digits.
This same problem also exists in AHK v2, and is fixed (and further explained) by #218.

### DIFF
--- a/source/script.cpp
+++ b/source/script.cpp
@@ -13093,7 +13093,7 @@ ResultType Line::EvaluateCondition() // __forceinline on this reduces benchmarks
 		case VAR_TYPE_DIGIT:
 			if_condition = true;
 			for (cp = ARG1; *cp; ++cp)
-				if (!_istdigit((UCHAR)*cp))
+				if (!_istdigit((USHORT)*cp))
 				{
 					if_condition = false;
 					break;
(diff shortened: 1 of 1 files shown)


Lexikos:
I assume the cast to UCHAR is intended to avoid undefined behaviour:

> The behavior of isdigit and _isdigit_l is undefined if c is not EOF or in the range 0 through 0xFF, inclusive.

`(USHORT)*cp` will produce values in the range `0xFF80`..`0xFFFF` if `*cp` is a non-ASCII character on the ANSI build, because `char` is signed. `(TBYTE)*cp` would produce `0x80`..`0xFF` for non-ASCII characters on the ANSI build and `0x0080`..`0xFFFF` on the Unicode build.

However, this doesn't fix the misidentification of digits; it just misidentifies more sensibly.

The documentation says:

> True if Var is empty or contains only digits, which consist of the characters 0 through 9.

But iswdigit returns true for non-ASCII "digits", including:
  -  ² (U+00B2), ³ (U+00B3), ¹ (U+00B9)
  - Digits from various scripts, such as Kayah Li (U+A900 .. U+A909).

but not
  - ⁰ (U+2070) .. ⁹ (U+2079). 
  - Digits from a few scripts, such as Javanese (U+A9D0 .. U+A9D9).

If we want to identify Unicode digits, iswdigit isn't useful because it isn't consistent about what it considers a digit. It can't even be called for code points greater than 0xFFFF, since that would require a type larger than `wint_t`. On the other hand, if we want to only identify ASCII digits 0 - 9 (as documented), iswdigit is just the wrong function.

The superscript numbers are not in the "Number, Decimal Digit" Unicode category, so the regex `\p{Nd}` correctly does not match any of them. It does match more scripts, but not all, possibly due to the outdated version of PCRE AutoHotkey is using (8.20).

PR #220, opened 2021-04-25
v2: Changed IsTime to ensure digit count is even and 14 maximum.

jeeswg:
This would change IsTime to only report true for dates of length: 4/6/8/10/12/14.

At present, IsTime allows dates of an odd length, and that are longer than 14 digits.
Also, the underlying algorithm treats '2006x' as '20060x', '200605x' as '2006050x', ..., '200605040302x' as '2006050403020x'.
E.g. '20002' is considered valid, even though it's ambiguous as either the 2nd or '20th' month of the year.

I had thought that the proposed behaviour was already the default, when I started using AutoHotkey, until I systematically tested 'var is date'/IsTime.

So, on balance, I would support this PR or one like it, although I don't feel particularly strongly.

A test script:
```
;test IsTime:
vDate := ""
vOutput := ""
Loop 20
{
	if (A_Index = 1)
		vDate .= "2"
	else
		vDate .= "1"
	vOutput .= IsTime(vDate) "`t" vDate "`r`n"
}
A_Clipboard := vOutput
MsgBox(vOutput)
```

### DIFF
--- a/source/util.cpp
+++ b/source/util.cpp
@@ -211,7 +211,8 @@ ResultType YYYYMMDDToSystemTime(LPTSTR aYYYYMMDD, SYSTEMTIME &aSystemTime, bool 
 		// explicit zero for the day of the month.  It also reports failure if st.wYear is
 		// less than 1601, which for simplicity is enforced globally throughout the program
 		// since none of the Windows API calls seem to support earlier years.
-		return SystemTimeToFileTime(&aSystemTime, &ft) ? OK : FAIL;
+		// Length checks, and 1601 being the minimum year value, ensure a digit count of 4/6/8/10/12/14.
+		return (SystemTimeToFileTime(&aSystemTime, &ft) && !(length % 2) && (length <= 14)) ? OK : FAIL;
 		// Above: The st.wDayOfWeek member is ignored by the above (but might be used by our caller), but
 		// that's okay because it shouldn't need validation.
 	}


Lexikos:
This PR currently makes IsTime inconsistent with what some of the functions accept. For instance, `IsTime("20216")` returns false, but `DateAdd("20216", 1, "D")` and `FormatTime("20216", "yyyy/MM")` interpret it as the input as 2021-06-01 (which is the only sensible interpretation other than to throw an error).

Not mentioned in the PR, it also (appropriately) affects what the Gui options and properties consider to be valid. However, currently only MonthCal.Value throws an exception for invalid values. DateTime.Value and the option strings ignore invalid values, so the additional validation could be more of a hindrance than a help.

> '20002' is considered valid, even though it's ambiguous as either the 2nd or '20th' month of the year.

There is nothing in the documentation indicating that a missing **digit** would default to 0, or any other reason that 2 would be interpreted as 20. The documentation indicates default values are supplied per **element**, such as MM, which has a default value of 01.

jeeswg:
Thanks! Your input is greatly appreciated.
My intention was that `IsTime` would 'only report true for dates of length: 4/6/8/10/12/14'.
And that any other details would be at your discretion.

My preferences would be:
- If a `DateTime` parameter is specified, but it doesn't meet the new stricter conditions, the function throws. E.g. `DateAdd`/`DateDiff`/`FormatTime`/`FileSetTime`/GUI setters are affected.
- `IsTime` would use stricter conditions. It would continue to return 0, and not throw.
- 'Get' functionality would be unchanged. E.g. `FileGetTime`/`A_Now`/GUI getters.

I may have failed to anticipate some other side effect.

> so the additional validation could be more of a hindrance than a help.

I'm not sure I understand.

I suppose the best way of expressing the current behaviour is this:
'You can specify a 1-digit value for a non-year element, but only if it's the rightmost element that you specify.'
That's a pretty unusual and long-winded rule.
AFAICS, in the documentation, a datestamp of length 5/7/9/11/13/15+ is currently undefined behaviour.
Anyone who knows that 'rightmost' rule would have had to infer it by mathematical experimentation.

My idea is to promote safety, e.g. to prevent someone accidentally missing/adding a digit.
In a similar way, CryptStringToBinary, with CRYPT_STRING_HEX, fails if the hex string length is odd.

Lexikos:
Don't take it out of context.

> DateTime.Value and the option strings **ignore invalid values**,

Changing the GUI options to *ignore* values which have an odd number of digits is not helpful.

jeeswg:
Great commit, thanks!

1 source code recommendation:
- `DateAdd`: accept a blank string (equivalent to `A_Now`, like `DateDiff/FormatTime/FileSetTime`).

3 documentation recommendations:
- `IsTime`: a blank string is regarded as an invalid date.
- `IsTime`: remove the obsolete reference to 'DATE'. `The word DATE may be used as a substitute for the word TIME, with the same result.`
- `DateAdd/DateDiff`: DateTime: A blank string is equivalent to `A_Now`, omitting a string throws an error. (See `FormatTime/FileSetTime` for good example descriptions.)

PR #224, opened 2021-04-26
v2: Added VerCompare().

jeeswg:
Compares two strings as version numbers.
```
Result := VerCompare(Version1, Version2)
```

Version numbers must not contain any characters other than digits and dots.
Version numbers must start and end with a digit.
Version numbers must not contain any consecutive dots.

Some examples:
```
MsgBox(VerCompare("1.2.3", "1.2.4")) ;-1
MsgBox(VerCompare("1.2", "1.2")) ;0
MsgBox(VerCompare("1.2", "1.2.0")) ;0
MsgBox(VerCompare("1.2", "1.2.0.0.0.0.0.0.1")) ;-1
MsgBox(VerCompare("1.2", "1.2.0.0.0.0.0.0.0")) ;0
```

Two uses are to compare version numbers and to compare IP addresses.

This function or similar, could be added as a stand-alone function, or by some other means such as a 'V' comparison mode in StrCompare or a 'NumCompare' function.
The function name can be changed if desired.
The code can be radically altered if desired.
I don't feel too strongly regarding this function's inclusion.

### DIFF
--- a/source/script.cpp
+++ b/source/script.cpp
@@ -332,6 +332,7 @@ FuncEntry g_BIF[] =
 	BIFn(Trim, 1, 2, BIF_Trim),
 	BIF1(Type, 1, 1),
 	BIF1(VarSetStrCapacity, 1, 2, {1}),
+	BIF1(VerCompare, 2, 2),
 	BIFn(WinActivate, 0, 4, BIF_WinActivate),
 	BIFn(WinActivateBottom, 0, 4, BIF_WinActivate),
 	BIFn(WinActive, 0, 4, BIF_WinExistActive),
--- a/source/script.h
+++ b/source/script.h
@@ -3431,6 +3431,7 @@ BIF_DECL(BIF_Process);
 BIF_DECL(BIF_ProcessSetPriority);
 BIF_DECL(BIF_MonitorGet);
 BIF_DECL(BIF_Wait);
+BIF_DECL(BIF_VerCompare);
 
 BIF_DECL(BIF_PerformAction);
 BIF_DECL(BIF_SetBIV);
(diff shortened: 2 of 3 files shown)


Lexikos:
This fails a basic usability test.

    MsgBox VerCompare(A_AhkVersion, "2.0")  ; Error: Parameter #1 of VerCompare is invalid.

See 8a015f7.

jeeswg:
(Thanks for the commit link.)

> basic usability test

That was by design. It works on version numbers that contain dots/digits only. (That excludes any alpha/beta versions.)

I tend to write all of my core functions for multiple programming languages. I wouldn't alter the default functionality just to suit AutoHotkey alpha/beta version numbers, that would be too niche. Although I might add it as an Options parameter.

If you've found any links re. de facto or de jure standards for alpha/beta version number formats, I might add a mode based on those standards.

The [`#Requires` criteria](https://www.autohotkey.com/docs/commands/_Requires.htm#Remarks) are sufficiently quirky/odd/extensive, that it could take users a few hours to fully get to grips with them, and yet still miss one or two logical implications; and they'd need to perform systematic tests to confirm their understanding.

The `#Requires` criteria also permit all sorts of unusual strings as legitimate version numbers, I'd much rather be warned about such comparisons. I.e. using the `#Requires` criteria by default effectively makes the function unsafe/unusable. That said, you could use `RegExMatch` to ensure that version numbers only contain dots/digits.

jeeswg:
Documentation for lexikos's implementation: [VerCompare](https://lexikos.github.io/v2/docs/commands/VerCompare.htm). Which mentions the SemVer standard. See also: [Software versioning](https://en.wikipedia.org/wiki/Software_versioning).

Great, I think I can work with this as is, without adding any modes. I'd just apply a validation check when dealing with version numbers other than `A_AhkVersion`, such as those obtained via `FileGetVersion`.

I'll wait a few weeks, in case people have any queries about the function, and then post a PR around the 10th of July, to add lexikos's version to AHK v1.1. @ lexikos: or if you're going to add it yourself, I'll skip the PR. Cheers.

PR #225, opened 2021-04-26
v2: Changed // to do floor division. Added ~/ to do integer division.

jeeswg:
## Changes:
- Changed `//` and `//=` to do floor division.
- Added `~/` and `~/=` to do integer division.
- The code for `//` and `//=` is optimised to use integer division when the numbers are both non-negative (positive or zero).

## Notes:
For reference: three types of division:
```
-3 / 2 = -1.5 [true division]
-3 ~/ 2 = -1 [integer division: divide and round towards 0]
-3 // 2 = -2 [floor division: divide and round down (towards -infinity)]
```

- The operator, `//`, is most commonly seen in Python, and its role is to do floor division. This PR aims to replicate that behaviour.
- Previously, in AHK v2, `//` was being used to do integer division. This broke the principle of least astonishment. For comparison, imagine switching the meaning of `*` and `/`, that would be very confusing.
- Previously, in AHK v1, `//` was performing two roles, floor and integer division, depending on whether the input values were integers or floats. This also broke the principle of least astonishment.
- What we have here is an attempt to neatly solve the problems by having two separate operators, one for floor division, and one for integer division.

## Choosing a symbol:
- The symbol, `~/`, comes from Dart. It is quite intuitive: `~` for approximate, and `/` for divide.
- A sometimes-used symbol for integer division, that I avoided using, because people often find it confusing when reading code, is `\`, a backslash. Imagine distinguishing between `\` and `/` at speed.
- Many languages don't have a symbol for integer division, they just use `/`, which performs integer division on integers, and true division on floats.
- Many languages use `//` for comments, and so can't use it for floor division.

## Further notes:
- Floor division and integer division are equivalent, when the input values share the same sign, e.g. if both numbers are positive, or both numbers are negative, you get the same result.
- Having all 3 division operators is a real advantage because you can copy complex formulas and code from other languages, without having to modify the structure. You don't have to replace operators with functions, which can be awkward and introduce bugs.
- A good summary:
[[rust-dev] Division and modulo for signed numbers](https://mail.mozilla.org/pipermail/rust-dev/2013-April/003680.html)

## Example code:
```

;true / integer / floor division: positive results:

MsgBox(3 / 2) ;1.5
MsgBox(3 ~/ 2) ;1 [divide and round towards 0]
MsgBox(3 // 2) ;1 [divide and round down (towards -infinity)]

var := 3
var /= 2
MsgBox(var) ;1.5

var := 3
var ~/= 2
MsgBox(var) ;1

var := 3
var //= 2
MsgBox(var) ;1

;true / integer / floor division: negative results:

MsgBox(-3 / 2) ;-1.5
MsgBox(-3 ~/ 2) ;-1 [divide and round towards 0]
MsgBox(-3 // 2) ;-2 [divide and round down (towards -infinity)]

var := -3
var /= 2
MsgBox(var) ;-1.5

var := -3
var ~/= 2
MsgBox(var) ;-1

var := -3
var //= 2
MsgBox(var) ;-2

```

## FloorDiv via IntDiv
I've been considering code possibilities regarding speed and maximum precision.
This would only affect one or two lines of code.
One possibility is code that in AutoHotkey would look like this:
```
;note: where vNum1 and vNum2 are integers
;note: code for how // should behave (not a proposed FloorDiv function)
;note: ~/ is equivalent to / for integers in C++
FloorDiv(vNum1, vNum2)
{
	if !vNum2
		throw Exception("Divide by zero.", -1) ;to match AHK v2

	if ((vNum1 >= 0) && (vNum2 > 0))
	|| ((vNum1 <= 0) && (vNum2 < 0))
		return vNum1 ~/ vNum2
	else
		return (vNum1 ~/ vNum2) - (Mod(vNum1, vNum2) ? 1 : 0)
}

;or:

FloorDiv(vNum1, vNum2)
{
	if !vNum2
		throw Exception("Divide by zero.", -1) ;to match AHK v2

	;separate ifs, or an if/else-if/else ladder:
	vRet := vNum1 ~/ vNum2
	if ((vNum1 >= 0) && (vNum2 > 0))
	|| ((vNum1 <= 0) && (vNum2 < 0))
		return vRet
	if Mod(vNum1, vNum2)
		--vRet
	return vRet
}
```


jeeswg:
I force-pushed an update to the intdiv branch and hence this PR.
This is because I noticed a problem relating to qmathFloor.
It was causing the following to be calculated incorrectly:
```
14 / -13 = -1.076923
1277 / -20 = -63.85
14 // -13 = -2 [gave the incorrect value -1]
1277 // -20 = -64 [gave the incorrect value -63]
```
But it's clear already in the AHK source code, that qmathFloor appears to be unreliable, and odd fixes have been used to get around this.

I experimented with the following 6 code snippets, and replaced number 3 with number 6 in the new version of the PR.
I found the speeds for all 6 methods virtually identical.
Since snippets 2 and 3 used qmathFloor, and had very rare errors, that left only 4, 5 and 6 possible candidates.

Floor division (by casting integers to doubles, float division, flooring, and casting a double to an integer) appeared to be marginally faster than integer division by itself, so I'm unclear why this comment was present in the AHK v1 source code:
> performance is much higher for integer vs. float division

I know in some contexts, like in assembly language, that that can be true, but not so in this context, it appears.

By using integer division as the basis for the calculations, and avoiding casting to and from doubles, the highest precision is maintained.
```
==================================================

1 original (note: SYM_INTEGERDIVIDE, the rest are SYM_FLOORDIVIDE):
				case SYM_INTEGERDIVIDE:
					if (right_int64 == 0)
						goto divide_by_zero;
					this_token.value_int64 = left_int64 / right_int64;
					break;

==================================================

2 floor division (no optimisation):
				case SYM_FLOORDIVIDE:
					if (right_int64 == 0)
						goto divide_by_zero;
					this_token.value_int64 = (__int64)qmathFloor((double)left_int64 / (double)right_int64);
					break;

==================================================

3 1st PR attempt (optimised):
				case SYM_FLOORDIVIDE:
					if (right_int64 == 0)
						goto divide_by_zero;
					if (left_int64 < 0 || right_int64 < 0)
						this_token.value_int64 = (__int64)qmathFloor((double)left_int64 / (double)right_int64);
					else
						this_token.value_int64 = left_int64 / right_int64; // Performance is much higher for integer vs. float division.
					break;

==================================================

4 one-liners:
				case SYM_FLOORDIVIDE:
					if (right_int64 == 0)
						goto divide_by_zero;
					if ((left_int64 >= 0 && right_int64 > 0)
						|| (left_int64 <= 0 && right_int64 < 0))
						this_token.value_int64 = left_int64 / right_int64;
					else
						this_token.value_int64 = (left_int64 / right_int64) - (left_int64 % right_int64 ? 1 : 0);
					break;

==================================================

5 preset, and 2 ifs:
				case SYM_FLOORDIVIDE:
					if (right_int64 == 0)
						goto divide_by_zero;
					this_token.value_int64 = left_int64 / right_int64;
					if ((left_int64 >= 0 && right_int64 > 0)
						|| (left_int64 <= 0 && right_int64 < 0))
						break;
					if (left_int64 % right_int64)
						--this_token.value_int64;
					break;

==================================================

6 preset, and if/else-if (2nd PR attempt):
				case SYM_FLOORDIVIDE:
					if (right_int64 == 0)
						goto divide_by_zero;
					this_token.value_int64 = left_int64 / right_int64;
					if ((left_int64 >= 0 && right_int64 > 0)
						|| (left_int64 <= 0 && right_int64 < 0))
						break;
					else if (left_int64 % right_int64)
						--this_token.value_int64;
					break;

==================================================
```
Test code:
```

MsgBox(-3 // 2) ;-2
MsgBox(-3 ~/ 2) ;-1

vNum := -3
vNum //= 2
MsgBox(vNum) ;-2

vNum := -3
vNum ~/= 2
MsgBox(vNum) ;-1

vSeed := 1234
RandomSeed(vSeed)
Loop 10000000
{
	;vNum1 := Random(-20000, 20000)
	;vNum2 := Random(-20, 20)
	vNum1 := Random(-2000000, 2000000)
	vNum2 := Random(-2000, 2000)
	if !vNum2
		continue
	vRet1 := vNum1 // vNum2
	vRet2 := Floor(vNum1 / vNum2)
	vRet3 := vNum1 ~/ vNum2
	vRet4 := Integer(vNum1 / vNum2)
	if !(vRet1 = vRet2)
		MsgBox("//" "`r`n" vNum1 "`r`n" vNum2 "`r`n" vRet1 "`r`n" vRet2)
	if !(vRet3 = vRet4)
		MsgBox("~/" "`r`n" vNum1 "`r`n" vNum2 "`r`n" vRet3 "`r`n" vRet4)
}
MsgBox("done")

```

Lexikos:
> The operator, //, is most commonly seen in Python, 

By users of Python, perhaps. AutoHotkey isn't for them.

> Previously, in AHK v2, // was being used to do integer division. This broke the principle of least astonishment. 

`//` has always been used for integer division when used with integers, which is how I've always used it, and what I've always expected it to do.

> Previously, in AHK v1, // was performing two roles, floor and integer division, depending on whether the input values were integers or floats. This also broke the principle of least astonishment.

It did indeed, and this was solved in v2.0-a116 by changing `//` and other integer operators to throw if given a floating-point number. It therefore behaves as a user of v1 should expect when used with integers, and avoids any nasty surprises by reporting floating-point inputs as an error rather than performing a different operation or losing precision.

You may think your solution is better. I do not.

jeeswg:
### Quote 1
> You may think your solution is better.

This PR is better than the present solution in at least one respect:
It makes available *both* operators.
Integer divide (towards 0), and floor divide (towards negative infinity).

We make it easier to copy and paste code from other languages (a) structurally unchanged (i.e. you don't have to swap any operators with functions), (b) completely unchanged.

An additional advantage of my proposal is: 2 well-known symbols for each operator.
If we would confusingly use `//` for integer division (towards 0), and then later decide we also want a floor divide operator, what symbol are we to use?

### Quote 2
> // has always been used for integer division when used with integers, which is how I've always used it, and what I've always expected it to do.

This is one valid way of interpreting it.
I always viewed the operator in this valid way:

'The floor divide operator in AHK v1 is broken.
It uses the floor divide symbol from Python, `//`, and is called floor divide, but doesn't always work like Python's floor divide.
If the input values are integers, you should avoid using it if either value could be negative.
Else you should cast to floats before performing the operation.
Hopefully this will be fixed in AHK v2.'

We could introduce `~/` to both AHK v1 and v2, if you wanted guaranteed integer division (towards 0) for both.

Probably your uses used non-negative input values anyway, so wouldn't have to be changed.
This is because integer and floor divide are equivalent for non-negative (positive/zero) input values.

### Quote 3
> By users of Python, perhaps. AutoHotkey isn't for them.

It is a logical statement to say that: just because one language uses one operator symbol, another language doesn't have to use the same symbol.
But there are benefits to being in sync with Python on this one particular issue. Mentioned lower down.

Python differs from C++ re. `%`. C++ uses integer mod (towards 0), and Python uses floor mod (towards -infinity).
This alone is the source of a lot of bugs. It would have been much better if Python had gone with `%`.
And chosen a different symbol for floor mod. E.g. R uses `%%`.
We'd be creating similar problems if we redefined `//`.

`//` is sufficiently well-known as floor divide, that it's listed on Wikipedia:
[// - Wikipedia](https://en.wikipedia.org/wiki///)

I advocate for originality and novelty in some situations, but for operators (something that can't be easily replaced, unlike functions), I think being conservative is the best approach.

It's not just Python that uses `//` for floor divide, at least Crystal and Smalltalk do too:
[Arithmetic/Integer - Rosetta Code](https://rosettacode.org/wiki/Arithmetic/Integer)

Personally I am not pro- or anti- Python, but it is becoming effectively the number one programming language, and the number one programming language for AI.
AI involves maths and hence floor divide.
It would be helpful if we could use mathematical formulas from Python, without bugs relating to integer/floor divide.

### A small revolution

A language that neatly supports both integer and floor functionality is a small revolution, and highly convenient.
That alone would get AutoHotkey some much deserved recognition, and help solve a long-standing problem in the programming community.
A lot of forward-looking programming languages are still grappling with this issue, and you get debates about which one to choose, when you could choose both.

I.e. one potential scheme:
```
int. mod: Mod() and % or mod
floor mod: FloorMod() and %%

int. div: ~/
floor div: //
```
I.e. we move towards a scenario where AHK is 'a great language for mathematicians' and for code that uses complex formulas.

If we use `//` for floor division, then we can potentially add `%%` (floor mod) from R.
And so we have a nice symmetry of double characters for floor functionality: `//` and `%%`.

If we confusingly used `//` for integer division (towards 0), that would mean that other major languages could not copy our integer division operator symbol. (Because it's already used for comments in many key languages.)
If we use `~/` then this can be more easily adopted by other languages, and could also provide a useful popular alternative to the confusing `\` for integer divide.

AHK also has the benefit of `/` as true division, with no truncation for integer input values.

We would then have all 3 classic division operators:
```
true div: /
int. div: ~/
floor div: //
```

So in short:
We popularise the idea of having an true divide operator.
We popularise the idea of having an integer divide operator.
We popularise the idea of having an integer divide symbol that isn't `\`.
We popularise the idea of having an integer divide symbol that can be used in multiple languages. `//` can't be used in most major languages, it's used for comments.
We popularise the idea of having an floor divide operator.
We popularise the idea of a language having all three operators: true (`/`), integer (`~/`) and floor (`//`).

### No man is an island

When programming in different languages, every similarity is a bonus, every incompatibility a hindrance.

'To use our language you must use our philosophy, therefore no ternary operator.'
Issues like this are a real pain. In particular the ternary operator, which is such a useful tool in programming.

### Bugs

We will avoid bugs, if we make our `//` consistent with Python's.

### Borrow the symbol but change the meaning

We adopted the `//` symbol from Python, it would be perverse to borrow the symbol but change the meaning.

### Start from scratch

If we started from scratch, with the task: choose an int. div. operator and a floor div. operator.
Floor div. operator, most famous existing operator would be Python's `//`.
Int div. operator, most famous existing operator would be `\`, which is confusing, searching around in major languages, Dart's `~/` is a great option.

PR #229, opened 2021-04-28
Add virtual keys for Japanese keyboards

gsittyz:
Fixes #55

This fix includes key names for Japanese-specific keys, which simplifies AHK settings for Japanese keyboard users.
In my opinion, it is better to keep the key names consistent with the VK names, not the names based on actual behaviors.   

Due to historical reasons, there are some unique keys and VKs for Japanese inputting systems, and users are dependent on these keys. Some of the keys behave like two or three state radio keys; keydown for enabling the function of a pressed key, and keyup for disabling that of a key previously pressed. This behavior might imitate physical switches for backward compatibility. 

Typical Japanese IMEs correspond to keydown events and ignore keyups of IME keys. In the default setting of the Microsoft IME, pressing Ime_off/on, Hiragana and Katakana keys guarantee to change an input mode to the designated one. The other keys toggle the input mode in various ways, sometimes ignoring the (virtual) key names.

When the hardware keyboard is set to "Japanese keyboard (106/109 key)," some of the key events are suppressed by OS and the behaviors became so complicated that the current AHK cannot handle them. However, when the keyboard setting is changed to "English keyboard (101/102 key)," the behaviors became much simpler and AHK can handle them.

I found the more detailed explanation in the following URL:
https://metacpan.org/pod/UI::KeyboardLayout#Far-Eastern-keyboards-on-Windows 

We can see how a Japanese keyboard works in the following code:
https://github.com/microsoft/Windows-driver-samples/blob/master/input/layout/fe_kbds/jpn/106/kbd106.c
https://github.com/microsoft/Windows-driver-samples/blob/master/input/layout/fe_kbds/jpn/101/kbd101.c

### DIFF
--- a/source/globaldata.cpp
+++ b/source/globaldata.cpp
@@ -730,6 +730,27 @@ modifier is specified along with it:
 , {_T("Launch_App1"), VK_LAUNCH_APP1}
 , {_T("Launch_App2"), VK_LAUNCH_APP2}
 
+// Japanese Virtual Key Support
+, { _T("Kana"), VK_KANA}
+, { _T("Ime_On"), VK_IME_ON}
+, { _T("Kanji"), VK_KANJI}
+, { _T("Henkan"), VK_CONVERT}
+, { _T("Ime_Off"), VK_IME_OFF}
+, { _T("Muhenkan"), VK_NONCONVERT}
+
+// These keys behave like 3-state radio keys.
+, { _T("Eisu"), VK_OEM_ATTN} // VK_DBE_ALPHANUMERIC
+, { _T("Katakana"), VK_OEM_FINISH} // VK_DBE_KATAKANA
+, { _T("Hiragana"), VK_OEM_COPY} // VK_DBE_HIRAGANA
+
+// These keys behave like toggle radio keys.
+, { _T("Hankaku"), VK_OEM_AUTO} // VK_DBE_SBCSCHAR
+, { _T("Zenkaku"), VK_OEM_ENLW} // VK_DBE_DBCSCHAR
+
+// These keys behave like toggle radio keys.
+, { _T("Romaji"), VK_OEM_BACKTAB} // VK_DBE_ROMAN
+, { _T("NoRomaji"), VK_ATTN} // VK_DBE_NOROMAN
+
 // Probably safest to terminate it this way, with a flag value.  (plus this makes it a little easier
 // to code some loops, maybe).  Can also calculate how many elements are in the array using sizeof(array)
 // divided by sizeof(element).  UPDATE: Decided not to do this in case ever decide to sort this array; don't


gsittyz:
# About language input keys

As far as I know, only Japanese and Korean keyboards have unique additional (physical) keys for IMEs. This is maybe because these two countries adapted computers earlier than other east asian countries and hardware instructions were needed to input complicated characters at that time. The VK codes inherit this history, so I think there are no overlap problems in the other languages. For example, Chinese keyboards don't have unique physical keys, and a Shift key can enable/disable an IME. 

In [https://en.wikipedia.org/wiki/Language_input_keys](https://en.wikipedia.org/wiki/Language_input_keys), only Japanese and Korean Keyboards are considered.

We can reference these materials about key mappings:
1. [https://docs.microsoft.com/en-us/windows/win32/inputdev/virtual-key-codes](https://docs.microsoft.com/en-us/windows/win32/inputdev/virtual-key-codes)
2. [https://docs.microsoft.com/en-us/previous-versions/windows/embedded/ee504832(v=winembedded.80)](https://docs.microsoft.com/en-us/previous-versions/windows/embedded/ee504832(v=winembedded.80))
3. [https://docs.microsoft.com/en-us/previous-versions/windows/embedded/ee505653(v=winembedded.80)](https://docs.microsoft.com/en-us/previous-versions/windows/embedded/ee505653(v=winembedded.80))

## no overlap
There are no overlaps in
- VK_IME_ON
- VK_CONVERT
- VK_IME_OFF
- VK_NONCONVERT.

## important overlaps
The material 1 reveals overlaps with Korean keyboards:
- VK_KANA(0x15): it overlaps with VK_HANGEUL and VK_HANGUL.
- VK_KANJI(0x16): it overlaps with VK_HANJA.

I think it is better to use these keys both in Korean and Japanese names. Is it allowed to assign multiple names on the same virtual key value in AHK?

## (trivial?) overlaps
By comparing 2 and 3, we can see overlaps. 
- VK_OEM_ATTN = VK_DBE_ALPHANUMERIC
- VK_OEM_FINISH = VK_DBE_KATAKANA
- VK_OEM_COPY = VK_DBE_HIRAGANA
- VK_OEM_AUTO = VK_DBE_SBCSCHAR
- VK_OEM_ENLW = VK_DBE_DBCSCHAR
- VK_OEM_BACKTAB = VK_DBE_ROMAN

Also by referencing 1 and 2 we get
- VK_ATTN = VK_DBE_NOROMAN.

I suspect the overlaps of VK_DBE_ keys are trivial because these oem keys are not used in standardized modern keyboards in the other languages. In my opinion, it is better to use names based on Japanese IME methods, since standardized Japanese keyboards have these language input keys.

gsittyz:
I suppose I described and answered the overlap problem....
Are there any obstacles to merge this PR?

Animan8000:
> I suppose I described and answered the overlap problem....
> Are there any obstacles to merge this PR?

Correct me if I am wrong but perhaps Lexikos is currently focusing on the beta of AHK v2.0 (current one is still in the alpha, so it will be a bigger update) which is supposed to be released in June or July, which could potentially explain the bit of inactivity for v1.1 which is fine.

Source: https://www.autohotkey.com/v2/v2-thoughts.htm

Lexikos:
Why the discrepancies between #55 and this #229?

> Is it allowed to assign multiple names on the same virtual key value in AHK?

Yes, but only one name will be returned by GetKeyName.

PR #231, opened 2021-05-03
v2: Added StartingPos parameter to StrReplace.

jeeswg:
Added a `StartingPos` parameter to `StrReplace()`, similar to `SubStr()`.

`ReplacedStr := StrReplace(Haystack, Needle [, ReplaceText, CaseSense, OutputVarCount, Limit, StartingPos])`

Some test code:
```
;StrReplace (StartingPos) test code:

;ReplacedStr := StrReplace(Haystack, Needle, ReplaceText, CaseSense, OutputVarCount, Limit, StartingPos)

CaseSense := 1
Limit := -1
;Limit := 3

Haystack := "aaaaaaaaaa", Needle := "a", Start := -12
;Haystack := "xy_xy_xy_xy_xy_xy_xy_xy_xy_xy_", Needle := "xy_", Start := -33

StartPos := Start
End := -Start
Loop
{
	vOutput := StartPos "`r`n`r`n"
	if StartPos
	{
		vOutput .= StrReplace(Haystack, Needle, "b", CaseSense, &Count, Limit, StartPos) "`r`n" Count "`r`n`r`n"
		vOutput .= StrReplace(Haystack, Needle, "", CaseSense, &Count, Limit, StartPos) "`r`n" Count "`r`n`r`n"
		vOutput .= StrReplace(Haystack, Needle, "cc_", CaseSense, &Count, Limit, StartPos) "`r`n" Count
	}
	MsgBox(vOutput)

	if (StartPos = End)
		break
	StartPos++
}
```
At one point, in `util.cpp`, I was torn between various options: increasing code size but adding code outside 2 loops, adding one line of code inside 2 loops, etc. In this PR, I have opted for the simplest solution, but there may be a more optimal solution.

I do not wish for functions to have too many parameters, however, I do not feel that 7 parameters are too many for a function, and that this parameter is worthwhile. Although not ideal, in AHK v1, `InputBox` had 11 parameters, and I found that quite manageable. I also feel that the `StartingPos` parameter is useful, and consistent with other functions such as `SubStr`, `InStr`, `RegExMatch`, `RegExReplace`, and a possible `StrCount` function.

Although some of this functionality could be achieved by using `SubStr` and concatenation etc, there are certain distracting and code-convoluting low-level tasks that I'd rather be handled directly by the built-in functions.

Some uses:
- Use `RegExMatch` to identify any leading whitespace, and then after that position, replace any tabs with spaces, and apply `RTrim`.
- File paths: replace a string in a file's name, but not in its directory.
- Use `InStr` to find the position, then replace the nth occurrence of a string.
- Compared to `RegExReplace`, you do not have to escape any characters.
- For very large strings, this should reduce some of the copying and moving about of data that would otherwise be required.

### DIFF
--- a/source/script.cpp
+++ b/source/script.cpp
@@ -317,7 +317,7 @@ FuncEntry g_BIF[] =
 	BIFn(StrLower, 1, 2, BIF_StrCase),
 	BIF1(StrPtr, 1, 1),
 	BIFn(StrPut, 1, 4, BIF_StrGetPut),
-	BIF1(StrReplace, 2, 6, {5}),
+	BIF1(StrReplace, 2, 7, {5}),
 	BIF1(StrSplit, 1, 4),
 	BIFn(StrUpper, 1, 2, BIF_StrCase),
 	BIF1(SubStr, 2, 3),
--- a/source/script2.cpp
+++ b/source/script2.cpp
@@ -5925,6 +5925,20 @@ BIF_DECL(BIF_StrReplace)
 		|| string_case_sense == SCS_INSENSITIVE_LOGICAL) // Not supported, seems more useful to throw rather than using SCS_INSENSITIVE.
 		_f_throw_param(3);
 
+	INT_PTR starting_offset = (INT_PTR)ParamIndexToOptionalInt(6, 1); // The one-based starting position in haystack (if any).
+	if (starting_offset > (INT_PTR)length)
+		starting_offset = length;
+	else if (starting_offset < 0) // Same convention as RegExMatch/Replace(): Treat negative StartingPos as a position relative to the end of the string.
+	{
+		starting_offset += length;
+		if (starting_offset < 0)
+			starting_offset = 0;
+	}
+	else if (!starting_offset)
+		_f_throw_param(6);
+	else
+		--starting_offset; // Convert to zero-based.
+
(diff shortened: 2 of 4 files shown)


Lexikos:
I already rejected this idea when you brought it up in 2018. As I said in 2018,

> If you want to operate on a substring, get one (with SubStr), then do that. 

The performance concerns, which I think are generally more theoretical than practical, can be mitigated by optimizing string handling. For example, ChakraCore has [several internal representations of strings](https://github.com/microsoft/ChakraCore-wiki/blob/master/String-Internals.md), including one to avoid copying a substring until actually necessary.

For very large strings, using StrReplace with a StartingPos parameter might be less efficient. StrReplace might reallocate and copy the string multiple times, including the leading part which wasn't included in the search.

> consistent with other functions such as SubStr, InStr, RegExMatch, RegExReplace

StrReplace does not need to be consistent with functions of different purposes, and still wouldn't be consistent with RegExReplace.

SubStr obviously needs StartingPos because it's half the function's purpose, and what gives StrReplace and every other function the indirect ability to operate on a substring.

For InStr and RegExMatch, a StartingPos parameter is used to find multiple matches. This wouldn't apply to StrReplace since it doesn't tell you where the replacement was made.

For RegExReplace, passing a substring might change the results due to an assertion at the start of the pattern (such as the common `^` or a look-behind assertion).

PR #232, opened 2021-05-03
v2: Changed OnMessage callback function parameter order.

jeeswg:
## Rationale

```
;OnMessage callback function (old order):
MyWndProc(wParam, lParam, uMsg, hWnd)
{
}

;OnMessage callback function (new order):
;the new order is consistent with a WndProc specified by SetWindowLong and GWL_WNDPROC:
MyWndProc(hWnd, uMsg, wParam, lParam)
{
}
```
Someone (not me) suggested that the OnMessage callback function parameter order be changed to be more standard, consistent with window procedures.

I found while working on some GUI-based code that this would be beneficial, e.g. to allow the same callback function to work with both AutoHotkey's OnMessage and the Winapi's SetWindowLong.

The AHK v1 parameter order is unusual, I didn't find the arguments made for it particularly convincing, and as a newbie I didn't find specifying all 4 arguments to be any great disadvantage, whereas if you want a callback function to work with both OnMessage and SetWindowLong, or if you're translating existing code from other languages, the more standard order would be better.

For reference:
[WindowProc callback function (Windows) | Microsoft Docs](https://docs.microsoft.com/en-us/previous-versions/windows/desktop/legacy/ms633573(v=vs.85))

## Note

AFAICS you only need to change 4 lines of code. Apologies if I have overlooked anything.

## Test code

```

;test OnMessage (AHK v2.0-a133)

;OnMessage parameter order receive.ahk
Persistent()
OnMessage(111, MyMessageMonitor)
MyMessageMonitor(hWnd, uMsg, wParam, lParam)
{
	MsgBox(A_ScriptHwnd "`r`n`r`n" hWnd "`r`n" uMsg "`r`n" wParam "`r`n" lParam)
}

;OnMessage parameter order send.ahk
DetectHiddenWindows(1)
hWnd := WinGetID("OnMessage parameter order receive.ahk")
PostMessage(111, 222, 333,, "ahk_id " hWnd)

```

### DIFF
--- a/source/application.cpp
+++ b/source/application.cpp
@@ -1759,10 +1759,10 @@ bool MsgMonitor(MsgMonitorInstance &aInstance, HWND aWnd, UINT aMsg, WPARAM awPa
 	// Set up the array of parameters for func->Invoke().
 	ExprTokenType param[] =
 	{
-		(__int64)awParam,
-		(__int64)(DWORD_PTR)alParam, // Additional type-cast prevents sign-extension on 32-bit, since LPARAM is signed.
+		(__int64)(size_t)aWnd,
 		(__int64)aMsg,
-		(__int64)(size_t)aWnd
+		(__int64)awParam,
+		(__int64)(DWORD_PTR)alParam // Additional type-cast prevents sign-extension on 32-bit, since LPARAM is signed.
 	};
 
 	ResultType result;


jeeswg:
Any feedback would be interesting. Thanks.

You had mentioned this on 'Thoughts for v2.0', so I thought you might be considering a change.
Not that I mind either way.

I see 4 options (all fine):
- Do nothing. (The user can redirect the parameters to be in WndProc order.) Hopefully the overhead is not an issue.
- Make the order like WndProc by default. Like this PR.
- Change the `OnMessage` `MaxThreads` parameter into an `Options` parameter, where you can specify a letter to use the standard WndProc order.
- Add an `Order` parameter to `OnMessage`.

Lexikos:
I figured closing the PR should send the message that I don't want this change and don't care to discuss it with you. I have limited time and personally don't want to spend any of it debating with you (jeeswg) in particular, based on past experience.

> You had mentioned this on 'Thoughts for v2.0'

I made a decision and removed it from the list without making any change.

> I didn't find the arguments made for it particularly convincing

I'd say the same about changing the order.

>  (The user can redirect the parameters to be in WndProc order.) Hopefully the overhead is not an issue.

No one should be writing one function to handle both OnMessage and WndProc directly. WndProc is called for all messages and generally requires calling CallWindowProc or DefWindowProc. OnMessage functions usually only handle a specific message or some related messages.

If the "overhead" of a single function call is an issue for you, you're using the wrong language.

Putting aside whether there is any value in matching the order of WindowProc, having two potential parameter orders for OnMessage seems like a terrible idea to me.

PR #235, opened 2021-05-13
v1: Backported A_Clipboard from alpha branch.

jeeswg:
A built-in variable, `A_Clipboard`, equivalent to the existing built-in variable, `Clipboard`.
This makes AHK v1 consistent with AHK v2, so that both versions have `A_Clipboard`.

Test code:
```

;test A_Clipboard:

;manually: copy image data in MS Paint
vClipSaved := ClipboardAll

A_Clipboard := "abc"
MsgBox, % A_Clipboard "`r`n" Clipboard ;abc
A_Clipboard .= "def"
MsgBox, % A_Clipboard "`r`n" Clipboard ;abcdef

Clipboard := "ghi"
MsgBox, % A_Clipboard "`r`n" Clipboard ;ghi
Clipboard .= "jkl"
MsgBox, % A_Clipboard "`r`n" Clipboard ;ghijkl

A_Clipboard := vClipSaved
;manually: paste image data in MS Paint

```

### DIFF
--- a/source/script.cpp
+++ b/source/script.cpp
@@ -64,6 +64,7 @@ VarEntry g_BIV_A[] =
 	A_x(BatchLines, BIV_BatchLines),
 	A_x(CaretX, BIV_Caret),
 	A_x(CaretY, BIV_Caret),
+	A_x(Clipboard, (BuiltInVarType)VAR_CLIPBOARD),
 	A_x(ComputerName, BIV_UserName_ComputerName),
 	A_(ComSpec),
 	A_x(ControlDelay, BIV_xDelay),


jeeswg:
### A long-standing Clipboard append bug.

I discovered a bug in AHK v1 (since at least AHK 1.0.48.05).
If you select files in Explorer, and press Ctrl+C to copy, and then append text to the clipboard, e.g. `Clipboard .= "abc"`, the resulting string is `<<>>abc`.

I discovered that the problem was fixed in AHK v2:
AutoHotkey v2.0-a110 had the problem.
AutoHotkey v2.0-a111 didn't.

`clipboard.h` (currently identical in AHK v1 and v2) contains this line:
`return IsClipboardFormatAvailable(CF_HDROP) ? _T("<<>>") : _T("");`

I've tried stepping through AHK v1. I don't expect to find a solution for weeks.

PR #236, opened 2021-05-14
v1: Backported Object-with-HWND support from alpha branch for WinActive/WinExist.

jeeswg:
## Rationale

The main use I had in mind was in backporting AHK v2 GUIs.

`WinActive`/`WinExist` are the only `WinXXX`/`ControlXXX` functions that exist in both AHK v1 and AHK v2.
(The others I have created in AHK v1 as custom backports, each with 'Object with HWND' code.)

## Implementation

The error handling can be changed as desired.

Btw I couldn't do this as a one-liner:
```
error: '=': left operand must be l-value

	if ((aParamCount)
	&& (Object *obj = dynamic_cast<Object *>(TokenToObject(*aParam[0]))))
	{
	}
```
I believe this is a warning to prevent people using `=` when they mean `==`, but as of yet, I haven't found a workaround.

## Test code
```

;test Window object with WinExist and WinActive (AHK v1):

;the target window (e.g. WordPad) should be open before running the script

hWnd := WinExist("ahk_class WordPadClass")
oWin := {hWnd: hWnd}
MsgBox, % oWin.HWND

;try with the target window active/inactive/hidden,
;the results should be the same for string and object,
;except for hidden: the object should work, but not the string

q::
MsgBox, % WinActive("ahk_id " hWnd) "`r`n" WinExist("ahk_id " hWnd)
. "`r`n`r`n" WinActive(oWin) "`r`n" WinExist(oWin)
return

```

### DIFF
--- a/source/script2.cpp
+++ b/source/script2.cpp
@@ -16214,10 +16214,35 @@ BIF_DECL(BIF_WinExistActive)
 	for (int j = 0; j < 4; ++j) // For each formal parameter, including optional ones.
 		param[j] = ParamIndexToOptionalString(j, param_buf[j]);
 
+	bool need_restore = false;
+	if (!ParamIndexIsOmitted(0)) 
+	{
+		if (Object *obj = dynamic_cast<Object *>(TokenToObject(*aParam[0])))
+		{
+			ExprTokenType t;
+			TCHAR wintitle[30] = _T("ahk_id ");
+			if (obj->GetItem(t, _T("HWND")))
+			{
+				param[0] = TokenToString(t, param_buf[0]);
+				if (_tcslen(param[0]) <= 22)
+					_tcscat(wintitle+7, param[0]);
+				param[0] = wintitle;
+				if (!g->DetectHiddenWindows)
+				{
+					need_restore = true;
+					g->DetectHiddenWindows = true;
+				}
+			}
+		}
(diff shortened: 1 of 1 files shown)


Lexikos:
This doesn't seem useful to me.

`wintitle`  goes out of scope before it is used by WinExist/WinActive (indirectly, via `param[0]`).

It seems likely that retrieving the HWND as an integer and returning it directly if valid/active would be cleaner and more efficient than what you're doing.

jeeswg:
Good points. Thanks!

I realised that I could write custom backport functions that could achieve this.
Variations of these:
```
WinActive(WinTitle:="", WinText:="", ExcludeTitle:="", ExcludeText:="")
{
	local hWnd := 0
	IfWinActive, % WinTitle, % WinText, % ExcludeTitle, % ExcludeText
		WinGet, hWnd, ID
	return hWnd ? hWnd + 0 : 0
}

WinExist(WinTitle:="", WinText:="", ExcludeTitle:="", ExcludeText:="")
{
	local hWnd
	WinGet, hWnd, ID, % WinTitle, % WinText, % ExcludeTitle, % ExcludeText
	return hWnd ? hWnd + 0 : 0
}
```
That said, it could be worth adding such functionality as built-in at some stage.

PR #242, opened 2021-05-29
v1: Added VarGetName().

jeeswg:
## Description

`VarName := VarGetName(Var)`

Retrieve the original variable name for a variable.
E.g. the original variable name may differ from the local variable name for a function.

## Implementation

A 2nd parameter, `Depth` could be added, to only go up n levels, rather than to the very top.
Perhaps -1 could indicate maximum depth, and be the default.
That said, I haven't wanted that functionality for myself.

## Rationale

This can be used in debugging e.g.:
```
myvar := 123
PrintVar(myvar)

PrintVar(ByRef param)
{
	MsgBox, % VarGetName(param) ": " param ;myvar: 123
}
```

Something I wanted when I first started using AutoHotkey around 10 years ago, although you get used to things not being available, and somehow get by without them.

Anyhow, I feel that this function or similar, is worth considering.

## Test code

```
;test VarGetName:

myvar := 123

MsgBox, % VarGetName(myvar) ;myvar
MyFunc(myvar)
MyFunc2(myvar)
PrintVar(myvar)

MyFunc(param)
{
	MsgBox, % VarGetName(param) ;param
}

MyFunc2(ByRef param)
{
	MsgBox, % VarGetName(param) ;myvar
}

PrintVar(ByRef param)
{
	MsgBox, % VarGetName(param) ": " param ;myvar: 123
}
```

### DIFF
--- a/source/script.cpp
+++ b/source/script.cpp
@@ -8674,6 +8674,8 @@ Func *Script::FindFunc(LPCTSTR aFuncName, size_t aFuncNameLength, int *apInsertP
 		min_params = 0;
 		max_params = 3;
 	}
+	else if (!_tcsicmp(func_name, _T("VarGetName")))
+		bif = BIF_VarGetName;
 	else
 		return NULL; // Maint: There may be other lines above that also return NULL.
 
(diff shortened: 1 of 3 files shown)


Lexikos:
I do not intend to add anything new to v1 that is not a good fit for v2. I would do this differently in v2.

PR #244, opened 2021-05-29
v2: Added ControlExist().

jeeswg:
## Description
```
Hwnd := ControlExist(Control [, WinTitle, WinText, ExcludeTitle, ExcludeText])
```
Similar to `ControlGetHwnd`, except it does not throw, it instead returns 0 if no matching control is found.

The `ControlExist`/`ControlGetHwnd` pairing would be similar to the `WinExist`/`WinGetID` pairing.

The function is useful for succinctly checking whether a control exists, and retrieving its window handle. It avoids needing the try/catch code that `ControlGetHwnd` would need.

## Test code

```
;test ControlExist:

q::
{
	hCtl := ControlGetHwnd("Edit1", "A")
	MsgBox(hCtl)
}

w::
{
	hCtl := ControlExist("Edit1", "A")
	MsgBox(hCtl)
}

e::
{
	hCtl := ControlGetHwnd("Edit2", "A") ;throws on Notepad main window
	MsgBox(hCtl)
}

r::
{
	hCtl := ControlExist("Edit2", "A") ;returns 0 on Notepad main window
	MsgBox(hCtl)
}
```

### DIFF
--- a/source/script.cpp
+++ b/source/script.cpp
@@ -65,6 +65,7 @@ FuncEntry g_BIF[] =
 	BIFn(ControlChooseString, 2, 6, BIF_Control),
 	BIF1(ControlClick, 0, 8),
 	BIFn(ControlDeleteItem, 2, 6, BIF_Control),
+	BIFn(ControlExist, 1, 5, BIF_ControlGet),
 	BIFn(ControlFindItem, 2, 6, BIF_ControlGet),
 	BIF1(ControlFocus, 1, 5),
 	BIFn(ControlGetChecked, 1, 5, BIF_ControlGet),
--- a/source/script.h
+++ b/source/script.h
@@ -665,7 +665,7 @@ enum BuiltInFunctionID {
 	FID_ProcessExist = 0, FID_ProcessClose, FID_ProcessWait, FID_ProcessWaitClose, 
 	FID_MonitorGet = 0, FID_MonitorGetWorkArea, FID_MonitorGetCount, FID_MonitorGetPrimary, FID_MonitorGetName, 
 	FID_OnExit = 0, FID_OnClipboardChange, FID_OnError,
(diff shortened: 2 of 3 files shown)


ExoVamp:
Noobs now can delve straight into infrastructure security handling.

Lexikos:
DetectTargetControl returns FAIL in several cases:
  - If *Control* is an object and...
    - it has no HWND property.
    - it has a HWND property with an invalid value.
    - the attempt to retrieve the HWND property throws an error.
  - The same again for *WinTitle*.
  - If the target window cannot be found.

The return value should not be set when the function is not returning successfully.

It is more robust to throw when the target window doesn't exist than not throw, and not overly inconvenient if the script relies on the last found window (as in `WinExist(WinTitle) && ControlExist(Control)`), but the behaviour still seems questionable. This is why ControlExist hasn't been implemented before now, despite obviously being trivial to implement.

The FID mechanism is in place for function implementations which share code, to reduce code size and repetition. In the case of your implementation, there is literally *nothing* being shared between ControlExist and the other functions at the source code level, except the enclosing BIF. (It may still result in smaller code due to shared function prologue/epilogue or unpredictable effects on the compiler's optimizer.)

ControlExist is *exactly the same* as ControlGetHwnd, except for a single parameter value for DetermineTargetControl. Capitalizing on that would reduce repetition, which would generally improve maintainability and reduce code size.

Compare to commit 4cfdff7a.

jeeswg:
Excellent. Thanks for the feedback!

jeeswg:
After consideration:
If the *window* cannot be found, `ControlExist` should return 0:
- The benefit of `WinGetID` and `ControlGetHwnd` is that they *do* throw. Protecting the user.
- The benefit of `WinExist` and `ControlExist` is that they *don't* throw. Thus you avoid convoluted code.

Having it not throw, gives more flexibility:
- If `ControlExist` didn't throw for a non-existent window. You could make it throw by checking `WinExist` / `WinGetID` first.
- If `ControlExist` did throw for a non-existent window. You would have to use `try` with `ControlExist`, which is no better than using `try` with `ControlGetHwnd`.
(You could use `WinExist` first, before `ControlExist`, but `ControlExist` could still throw, because the window might have ceased to exist in that small gap.)

In general, 'Exist' functions should only throw if passed invalid parameters.

(I'm not 100% sure what your position is, whether it aligns with mine or not, I'm simply stating my current view.)

PR #222, opened 2021-04-25
v2: Added WinIsEnabled/WinIsVisible.

jeeswg:
```
IsEnabled := WinIsEnabled(WinTitle, WinText, ExcludeTitle, ExcludeText)
IsVisible := WinIsVisible(WinTitle, WinText, ExcludeTitle, ExcludeText)
```

Together with #221, the PRs give us this situation:
```
Before:
ControlGetChecked
ControlGetEnabled/___, ControlSetEnabled, WinSetEnabled
ControlGetVisible/___, ControlShow/WinShow, ControlHide/WinHide

After: [where REN = renamed, ADD = added]
[REN]ControlIsChecked
[REN]ControlIsEnabled/[ADD]WinIsEnabled, ControlSetEnabled, WinSetEnabled
[REN]ControlIsVisible/[ADD]WinIsVisible, ControlShow/WinShow, ControlHide/WinHide
```

For functions that return a bool, 'Is', is more natural than 'Get'.
And it can be somewhat/a little distracting when certain functions lack an inverse.

Like WinShow, WinIsVisible ignores A_DetectHiddenWindows.

One use for WinIsVisible is to list hidden windows.
You use DetectHiddenWindows(1) and WinGetList to list visible *and* hidden windows, and then WinIsVisible to filter out the visible windows.

Since writing this description, I've noticed that the 'L' and 'i' in ControlIsXXX, clash a little, although they look great together in Lucida Console font amongst others.

In the [AHK v2 function index](https://lexikos.github.io/v2/docs/commands/index.htm), this would mean that all the functions that state 'Returns a non-zero [number/value]' would now contain 'Is' or 'Has' in the name:
```
ControlGetChecked -> ControlIsChecked
ControlGetEnabled -> ControlIsEnabled
ControlGetVisible -> ControlIsVisible
HasBase
HasMethod
HasProp
IsLabel
IsObject
IsSet / IsSetRef
WinIsEnabled [NEW]
WinIsVisible [NEW]
```

I'm not super concerned with whether these changes are made or not, but on the whole, find them worthwhile.

Here are some demo scripts:
```
q:: ;test WinIsVisible:
{
	DetectHiddenWindows(0)
	hWnd := WinGetID("ahk_class Notepad")

	WinHide("ahk_id " hWnd)
	Sleep(500)
	ToolTip(WinIsVisible("ahk_id " hWnd))
	Sleep(3000)

	WinShow("ahk_id " hWnd)
	Sleep(500)
[...]

PR #246, opened 2021-06-02
v2: Added WinIsAlwaysOnTop/WinIsEnabled/WinIsVisible.

jeeswg:
## Description

Replaces #222. See the link for more rationale info.

```
IsAlwaysOnTop := WinIsAlwaysOnTop([WinTitle, WinText, ExcludeTitle, ExcludeText])
IsEnabled := WinIsEnabled([WinTitle, WinText, ExcludeTitle, ExcludeText])
IsVisible := WinIsVisible([WinTitle, WinText, ExcludeTitle, ExcludeText])
```

`WinIsAlwaysOnTop`: Returns a non-zero value if the specified window is always-on-top.
`WinIsEnabled`: Returns a non-zero value if the specified window is enabled.
`WinIsVisible`: Returns a non-zero value if the specified window is visible.

## Differences versus previous PR

Added `WinIsAlwaysOnTop`.
- It completes the pair with `WinSetAlwaysOnTop`.
- It's something simple that newbies might want to check, that, without this function, would involve using a bitwise operator and a system constant.

Added an `IsWindowCloaked` check to `WinIsVisible`.
- This keeps the definition of a visible window consistent throughout AHK.
- This makes the function more useful than simply checking `IsWindowVisible` (which can be done via `DllCall`).

Comments clearly indicate code related to `WinIsVisible`.
- Users had wanted something like `ahk_dhw/ahk_mode/ahk_opt`. If introduced, code might be refactored, and you might be able to remove the `need_restore` variables for `WinIsVisible` and `WinShow`.

All 3 functions use `BIF_WinGet`, none use `BIF_WinShow`.
- `BIF_WinShow` is untouched, which keeps everything cleaner.

## ahk_XXX modes (re. WinShow/WinIsVisible refactoring)

I had in mind something like:
`XXX ahk_opt VHS`: a visible or hidden window whose title starts with XXX.
`XXX ahk_opt HC`: a hidden window whose title contains XXX.
`ahk_opt VE ahk_xtitle XXX`: a visible window, who's title doesn't end with XXX.

Modes before and after:
```
1/2/3/4/RegEx -> S/C/X-or-Q/E/R [Starts/Contains/eXact-or-eQuals/Ends/RegEx] [new: 4 (Ends)]
0/1/2 -> V/VH/H [Visible/Visible-Or-Hidden/Hidden] [new: 2 (Hidden)]

Equals makes more sense than Exact, if any case sensitivity options are added.
```

I don't use `ExcludeTitle` or `ExcludeText`, and would be fine with `ahk_xtitle` and `ahk_xtext` instead.

## Test code

```

;test WinIsVisible/WinIsEnabled/WinIsAlwaysOnTop:

DetectHiddenWindows(0)

;win criteria 1:
vWinCriteria := "Untitled - Notepad ahk_class Notepad"

;win criteria 2:
hWnd := WinGetID(vWinCriteria)
;vWinCriteria := "ahk_id " hWnd

q:: ;test WinIsVisible (hide/show):
{
	WinHide(vWinCriteria)
	Sleep(500)
	ToolTip(WinIsVisible(vWinCriteria))
	Sleep(3000)

	WinShow(vWinCriteria)
	Sleep(500)
	ToolTip(WinIsVisible(vWinCriteria))
}

w:: ;test WinIsEnabled (disable/enable):
{
	WinSetEnabled(0, vWinCriteria)
	Sleep(500)
	ToolTip(WinIsEnabled(vWinCriteria))
	Sleep(3000)

	WinSetEnabled(1, vWinCriteria)
	Sleep(500)
	ToolTip(WinIsEnabled(vWinCriteria))
}

e:: ;test WinIsAlwaysOnTop (on-top/not on-top):
{
	WinSetAlwaysOnTop(1, vWinCriteria)
	Sleep(500)
	ToolTip(WinIsAlwaysOnTop(vWinCriteria))
	Sleep(3000)

	WinSetAlwaysOnTop(0, vWinCriteria)
	Sleep(500)
	ToolTip(WinIsAlwaysOnTop(vWinCriteria))
}

```


Lexikos:
After further thought, I realized that I had already considered "Is" vs "Get" for the existing functions when I implemented them, and settled on "Get" along with renaming some of the former sub-commands to give them obvious counterparts and show that they belong to a single category. With a few exceptions like WinGetList, they correspond to what would be properties if Window was an object, operating on a single target window.

I had also considered a function for testing window visibility, and ultimately decided it wasn't a problem that needed a solution. Also, there isn't a single solution that fits all scenarios.

With DetectHiddenWindows off, WinExist tells you whether *any* visible window matches the parameters. By contrast, WinIsVisible only tests the first matching window, which could potentially be a hidden window even if there is a matching visible window.

In the rare event that one needs to test a specific top-level window's enabled status, it can actually be done with `ControlGetEnabled(WinExist(...))`, which is a little easier than DllCall or WinGetStyle.

`ControlGetVisible(WinExist(...))` also works, but typically requires `DetectHiddenWindows true` and doesn't exclude cloaked windows.

Note that IsWindowVisible will return false if any parent window is hidden, even if the target window has the `WS_VISIBLE` style. Whether this is desirable depends on the reason for testing visibility.

PR #247, opened 2021-06-18
Arm64

fobrs:
I've tested this on an Asus NOVAGO and MacBook Air M1 with Parallels. Both running the latest preview version of Windows 10.

The script which I used, uses the WinEnum function from here: https://github.com/cocobelgica/AutoHotkey-Util/blob/master/WinEnum.ahk

The DllCall() and RegisterCallback() functions run fine with this port.


joedf:
Interesting, never thought I'd see ARM64 support so soon!

PR #248, opened 2021-07-28
v1: Backported VerCompare() from alpha branch.

jeeswg:
## Introduction

A backport attempt for `VerCompare`.

Original commit: https://github.com/Lexikos/AutoHotkey_L/commit/8a015f75017e1fad58575b959a7aa6cc49c58f3c.

## Test code

```

;test code: VerCompare (AHK v1)

if (VerCompare(A_AhkVersion, "2.0") < 0)
    MsgBox("This version < 2.0; possibly a pre-release version.")
else
    MsgBox("This version is 2.0 or later.")

MsgBox(VerCompare("1.20.0", "1.3"))  ; Returns 1
MsgBox(StrCompare("1.20.0", "1.3"))  ; Returns -1

MsgBox(VerCompare("2.0-a137", "2.0-a136"))  ; Returns 1
MsgBox(VerCompare("2.0-a137", "2.0"))  ; Returns -1
MsgBox(VerCompare("10.2-beta.3", "10.2.0"))  ; Returns -1

MsgBox(Text)
{
	MsgBox, % Text
}

StrCompare(ByRef String1, ByRef String2, CaseSensitive:=0) ;ByRef for performance
{
	local Output, SCS
	SCS := A_StringCaseSense
	if (CaseSensitive = 1)
		StringCaseSense, % "On"
	else if (SCS = "On")
		StringCaseSense, % "Locale"
	Output := ("" String1 > "" String2) ? 1 : ("" String1 < "" String2) ? -1 : 0
	StringCaseSense, % SCS
	return Output
}

```

## Btw

Btw.
Congrats on AHK v2.0-beta.1. Excellent Stuff.
Sometimes in everyday life, when something good happens, I say Fantastikos.

### DIFF
--- a/source/script.cpp
+++ b/source/script.cpp
@@ -8674,6 +8674,12 @@ Func *Script::FindFunc(LPCTSTR aFuncName, size_t aFuncNameLength, int *apInsertP
 		min_params = 0;
 		max_params = 3;
 	}
+	else if (!_tcsicmp(func_name, _T("VerCompare")))
+	{
+		bif = BIF_VerCompare;
+		min_params = 2;
+		max_params = 2;
+	}
 	else
 		return NULL; // Maint: There may be other lines above that also return NULL.
 
--- a/source/script.h
+++ b/source/script.h
@@ -3358,6 +3358,7 @@ BIF_DECL(BIF_Trim); // L31: Also handles LTrim and RTrim.
 
 BIF_DECL(BIF_Hotstring);
 BIF_DECL(BIF_InputHook);
+BIF_DECL(BIF_VerCompare);
 
 
 BIF_DECL(BIF_IsObject);
--- a/source/script2.cpp
+++ b/source/script2.cpp
@@ -18790,6 +18790,16 @@ BIF_DECL(BIF_Exception)
 
 
 
+BIF_DECL(BIF_VerCompare)
+{
+	TCHAR buf[MAX_NUMBER_SIZE];
+	LPTSTR a = ParamIndexToString(0, aResultToken.buf);
+	LPTSTR b = ParamIndexToString(1, buf);
+	aResultToken.value_int64 = CompareVersion(a, b);
+}
+
+
+
 ////////////////////////////////////////////////////////
 // HELPER FUNCTIONS FOR TOKENS AND BUILT-IN FUNCTIONS //
 ////////////////////////////////////////////////////////


Lexikos:
> // Set default return value in case of early return.
> aResultToken.symbol = SYM_STRING;
> aResultToken.marker = _T("");

If you remove this clearly unnecessary code, `aResultToken.symbol = SYM_INTEGER;` will also be unnecessary, because it is the default return type for BIFs.

> LPTSTR a = ParamIndexToString(0, aResultToken.buf);
> LPTSTR b = ParamIndexToString(1, aResultToken.buf);

If both parameters are pure numbers, they will both be written into the same buf, and the result will be incorrect (or correct for the wrong reason, if the numbers are equal).

jeeswg:
I see now, `ParamIndexToString`, calls `TokenToString`, which sometimes *returns* `aBuf`.
I had thought the buffer was temporary and discarded. I'll review my other PRs in light of this.

I would have expected `ParamIndexToString` to sometimes throw, for an unusual parameter.
I'll assume that the check is unnecessary.

I'll force push an update in the next 20 minutes.
You're welcome to submit an alternative.

Lexikos:
> I had thought the buffer was temporary and discarded.

I can't imagine what purpose that would have. When the token contains a pure number, the converted string must go somewhere. If the function was able to allocate dynamic memory (and have it be freed automatically at some point), or if it had a shared buffer available for this purpose, it would just use that and have no need for a buffer to be passed.

> I would have expected ParamIndexToString to sometimes throw,

You will not find any C++ exception handling in the AutoHotkey source code. For what purpose would it throw an exception? The expected behaviour is for an object to be interpreted as an empty string, and a pure number to be converted to a string as needed. TokenToString must always succeed, and produce a string.

>  for an unusual parameter.

Unusual in what way? Each call to the macro was correct, if considering only the requirements of the macro. There is nothing wrong with passing the same buffer multiple times, as long as you aren't intending to use the previous content of the buffer.

> I'll assume that the check is unnecessary.

I've no idea what you're referring to.

jeeswg:
Many thanks for the detailed response.

> I'll assume that the check is unnecessary.
I meant: I'll assume that setting the return value is unnecessary.

Possible issues with various kinds of parameter validation in AHK v1/v2:
- unset value
- object without 'ToString'
- object 'ToString' crashing
- VarRef
- VarRef pointing to an unset variable
- NaN
- ClipboardAll
- A_ variable

To know that `ParamIndexToString` or any other macro/function doesn't throw or can't fail, requires checking everything it calls, all the way down the chain, which isn't always straightforward.

PR #250, opened 2021-07-28
v1: Backported Object.Has() from alpha branch. Added ObjHas().

jeeswg:
## Introduction

A backport attempt for `Object.Has()`.

## Test code

```
;test code: Object.Has() and ObjHas() (AHK v1)

oArray := [1, 2, 3]
MsgBox, % oArray.Has(2) ;1
MsgBox, % oArray.Has(4) ;0
MsgBox, % ObjHas(oArray, 2) ;1
MsgBox, % ObjHas(oArray, 4) ;0

oMap := Object("a",1, "b",2, "c",3)
MsgBox, % oMap.Has("b") ;1
MsgBox, % oMap.Has("d") ;0
MsgBox, % ObjHas(oMap, "b") ;1
MsgBox, % ObjHas(oMap, "d") ;0

```

### DIFF
--- a/source/script.cpp
+++ b/source/script.cpp
@@ -8551,7 +8551,6 @@ Func *Script::FindFunc(LPCTSTR aFuncName, size_t aFuncNameLength, int *apInsertP
 		BIF_OBJ_CASE(Length, 		0, 0)
 		BIF_OBJ_CASE(MinIndex, 		0, 0)
 		BIF_OBJ_CASE(MaxIndex, 		0, 0)
-		BIF_OBJ_CASE(HasKey,		1, 1) // key
 		BIF_OBJ_CASE(GetCapacity,	0, 1) // [key]
 		BIF_OBJ_CASE(SetCapacity,	1, 2) // [key,] new_capacity
 		BIF_OBJ_CASE(GetAddress,	1, 1) // key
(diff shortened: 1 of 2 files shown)


Lexikos:
I'm not in favour of adding more v1-only "Obj" functions. I have reservations about back-porting anything that will ultimately create more ways to write scripts that aren't fully compatible with v2 (as I did with `Loop Files/Reg`, for instance).

That aside, `BIF_OBJ_CASE` is just a macro for an if-statement which sets some variables based on a string match, and `BIF_METHOD` is a macro for a trivial function definition. There is no need for `BIF_ObjHas` to exist as a function distinct from `BIF_ObjHasKey`. `BIF_OBJ_CASE(HasKey,		1, 1)` can be expanded out to include *suffix = "has"*.

jeeswg:
I personally would not miss an `ObjHas` function.
However, I did not see a good reason why the `Has` method alone should lack a function equivalent.

jeeswg:
## `ObjXXX` functions

To clarify my point:
In AHK v1, if `MyObj.XXX(...)` exists, then `ObjXXX(MyObj, ...)` must exist.
Since consistency is absolutely essential to any programming endeavour.

A further point is to keep this safe for all built-in methods, otherwise special case code would be required:
```
;AHK v1: test oMap[vMethod](...) v. Obj%vMethod%(oMap, ...)
oMap := Object("k1","v1")
vMethod := "HasKey"
MsgBox, % oMap[vMethod]("k1")
MsgBox, % oMap[vMethod]("k2")
MsgBox, % Obj%vMethod%(oMap, "k1")
MsgBox, % Obj%vMethod%(oMap, "k2")
return
```

## Further consistency

This article will be revamped soon, but it covers concerns about consistency/compatibility: [My AHK v2 Priorities](https://bazzle.co.uk/ahk/v2-priorities.htm). In particular `Loop XXX` and `NumPut`.

PR #252, opened 2021-08-07
v2: Added RegCreateKey().

jeeswg:
## Description

`RegCreateKey(KeyName)`

Creates a registry key.

## Rationale

Currently, `RegWrite` is available, but this creates a key *and* a value.
This would only create a key.

Modelled on `RegDeleteKey` and `DirCreate`.

## Implementation

All regarding the `case FID_RegCreateKey` code block:

Implementation possibilities for `RegCreateKeyEx` (perhaps ignore):
`samDesired`: 3 possibilities: `KEY_WRITE` or `KEY_CREATE_SUB_KEY | STANDARD_RIGHTS_WRITE` or `KEY_CREATE_SUB_KEY`
`lpClass`: `_T("")` or `NULL` [elsewhere the source code uses `_T("")`]

Implementation possibilities for `RegCreateKeyEx` (probably implement):
`dwRes`: `&dwRes` or `NULL` [i.e. change to `NULL`, and omit definition line: `DWORD dwRes`]

Implementation correction:
This should be added under `RegCreateKeyEx`, just before `break`:
`RegCloseKey(hRegKey)`

Note:
`g->RegView` returns `KEY_WOW64_32KEY` or `KEY_WOW64_64KEY` (not 32 or 64).

## Test code

```
;test code: RegCreateKey (AHK v2)

RegCreateKey("HKEY_CURRENT_USER\Software\Microsoft\Notepad\MyKey0")

Loop 4
	RegCreateKey("HKEY_CURRENT_USER\Software\Microsoft\Notepad\MyKey" A_Index)

MsgBox("done")
```

## The Final PR

Most likely, this is the final PR, re. my view of a 'complete AutoHotkey'.
But I will at least explore these (names can be changed):

RemoteBuffer class. [Similar to this](https://autohotkey.com/board/topic/11173-module-remotebuffer-20/page-2#entry501339).

DTPGetDate/DTPSetDate. Get/set external `SysDateTimePick32` control dates.

BinCompare/BinCopy.
InBuf. Find nth/nth-to-last occurrence.
BinDiff/StrDiff. 2 buffers/strings. Find nth/nth-to-last difference (byte/char).

Struct class. (AHK code prototype, not C++.)

### DIFF
--- a/source/script.cpp
+++ b/source/script.cpp
@@ -260,6 +260,7 @@ FuncEntry g_BIF[] =
 	BIFn(ProcessWait, 1, 2, BIF_Process),
 	BIFn(ProcessWaitClose, 1, 2, BIF_Process),
 	BIF1(Random, 0, 2),
+	BIFn(RegCreateKey, 1, 1, BIF_Reg),
 	BIFn(RegDelete, 0, 2, BIF_Reg),
 	BIFn(RegDeleteKey, 0, 1, BIF_Reg),
 	BIFn(RegExMatch, 2, 4, BIF_RegEx, {3}),
--- a/source/script.h
+++ b/source/script.h
@@ -666,7 +666,7 @@ enum BuiltInFunctionID {
 	FID_DriveGetList = 0, FID_DriveGetFilesystem, FID_DriveGetLabel, FID_DriveGetSerial, FID_DriveGetType, FID_DriveGetStatus, FID_DriveGetStatusCD, FID_DriveGetCapacity, FID_DriveGetSpaceFree,
 	FID_EnvGet = 0, FID_EnvSet,
 	FID_PostMessage = 0, FID_SendMessage,
-	FID_RegRead = 0, FID_RegWrite, FID_RegDelete, FID_RegDeleteKey,
+	FID_RegRead = 0, FID_RegWrite, FID_RegDelete, FID_RegDeleteKey, FID_RegCreateKey,
 	FID_SoundGetVolume = 0, FID_SoundGetMute, FID_SoundGetName, FID_SoundGetInterface, FID_SoundSetVolume, FID_SoundSetMute,
 	FID_RunWait = 0, FID_ClipWait, FID_KeyWait, FID_WinWait, FID_WinWaitClose, FID_WinWaitActive, FID_WinWaitNotActive,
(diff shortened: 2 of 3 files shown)


iseahound:
Surprisingly useful (for edge cases). 

I went though my collection of .reg files and found 2 files that only created keys without values:

* #### Adds the "Previous Versions" button to the to context menu

    Windows Registry Editor Version 5.00
    [HKEY_CLASSES_ROOT\AllFilesystemObjects\shellex\ContextMenuHandlers\{596AB062-B4D2-4215-9F74-E9109B0A8153}]

* #### Part of a script to restore magnet link functionality to all web browsers.

    Windows Registry Editor Version 5.00
    [HKEY_CURRENT_USER\Software\Classes\magnet\shell]
    [HKEY_CURRENT_USER\Software\Classes\magnet\shell\open]

The above are excepts of full scripts. The full working .reg files can be provided on demand.

Lexikos:
> Rationale

You say that, but I see nothing to explain what gives this addition any purpose or value (and at this point, I don't want you to tell me).

> This should be added under RegCreateKeyEx, just before break:
> RegCloseKey(hRegKey)

As general advice, and not a request: If you find an error in your pull request, do not merely comment on it - **fix it**! You only need to make a commit to the same branch and push it. You can optionally amend the commit and `push -f` to replace the previous commit.

If your function is able to open a root key on a remote system and is unable to create the key, it will fail to close the root key.

It may be meaningful to omit KeyName, as in either of these cases it would create an empty key:
  - After `RegDeleteKey()` has deleted the key of the current registry item, and all of its content.
  - After some user action or external code has deleted the key of the current registry item (unlikely).

*samDesired* is the access requested for the returned key handle, not the access required in order to create the key (which you either have or don't have). If you are creating a key and not writing to it, `KEY_WRITE` is not necessary. If the key already exists, passing `KEY_WRITE` would cause the function to fail if you don't have write access to it, while omitting `KEY_WRITE` would allow the function to succeed. It is hard to say which behaviour is more useful.

Commit e5df2365 provides an implementation with marginally reduced impact on code size and (I think) fewer issues.

@iseahound 
Use fenced code blocks when posting code. Because you didn't, some of the slashes were misinterpreted as escape characters, and became invisible.

PR #254, opened 2021-09-13
Create SECURITY.md

zidingz:
A simple instruction for security researchers.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,5 @@
+# Security Policy
+
+## Reporting a Vulnerability
+
+Please report security issues to <email>


OmgImAlexis:
Since issues aren't on this repo I hope this is okay to put this here. The SSL cert on the website needs updating since it's signed with the old Let's Encrypt root.

```
This server could not prove that it is autohotkey.com; its security certificate expired 1,492 days ago. This may be caused by a misconfiguration or an attacker intercepting your connection. Your computer's clock is currently set to Friday, October 1, 2021. Does that look right? If not, you should correct your system's clock and then refresh this page.
```

joedf:
The ssl cert issue is being discussed here:
- https://www.autohotkey.com/boards/viewtopic.php?f=3&t=95148
- https://www.autohotkey.com/boards/viewtopic.php?p=422901#p422901

Lexikos:
Given the (poor) quality of the few "issues" that have been reported thus far, no thanks.

PR #255, opened 2021-10-07
Fix FlatVector::Remove

HotKeyIt:
FreeRange is designed to remove a range from + to rather than a count of items. count = 1 makes only sense if i=0 since otherwise the for loop is never executed and Data not freed: for (; i < count; ++i)

### DIFF
--- a/source/script_object.h
+++ b/source/script_object.h
@@ -170,7 +170,7 @@ public:
 	{
 		auto v = Value();
 		ASSERT(i >= 0 && i + count <= data->length);
-		FreeRange(i, count);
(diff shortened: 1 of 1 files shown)


Lexikos:
I see this was preventing dynamic properties from being freed if the property is not at index 0. (It does not affect object or string values because of ReturnMove.)

> FreeRange is designed to remove a range from + to 

That's clearly not intended:
```C++
void FreeRange(index_t i, index_t count)
```

To keep FreeRange consistent with the other methods, commit ca58fa8 fixes it to behave as implied by the parameter names.

PR #256, opened 2021-11-09
mark some static arrays as const so they are not reinitialized

mikeblas:
This saves code size and time, particularly in frequently called
functions and loops

### DIFF
--- a/source/script.cpp
+++ b/source/script.cpp
@@ -3548,7 +3548,7 @@ inline ResultType Script::IsDirective(LPTSTR aBuf)
 
 		int i;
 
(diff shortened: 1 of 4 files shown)


Lexikos:
Static variables are never *reinitialized*; they are initialized only once.

Further to that, because the entire content of each of these static arrays is known at compile-time, the compiler is already able to pre-initialize the entire array; i.e. it exists as data in the PE file, and no code is needed for constructing the array.

Whether these static arrays are const or not only affects which PE section they end up in (.rdata or .data).

Changing szHexData **from non-static to static** avoids the need to reinitialize it each time the function is called. This also reduces the size of the function, but increases the size of the .data or .rdata section (by a smaller amount).

PR #258, opened 2021-11-10
fix build to correctly reference PDBs from regex library

mikeblas:
The build would generate a dozen or so errors about PDB files not
being found. This was because the build didn't reference the correct
directory for the PDB files generated by the library build. This patch
fixes that issue, and now the builds have correct symbols, as expected,
with no warnings.


Lexikos:
For v1,
> The project is configured to build with the Visual C++ 2010 toolset if available

... and for that toolset, there were no warnings or errors.

PR #259, opened 2021-11-10
fix issues in GetEnvVarReliable

mikeblas:
GetEnvVarReliable() does some casting between size_t and DWORD, which is
problematic because size_t is eight bytes on Win32.

Further, the call to GetEnvVariable() may return non-zero but still not
copy any data to the output buffer. This case can be encountered when the
buffer is precisely the length of the required output, but can happen in
other circumstances. An extra check isn't expensive and makes this
code safer by correctly catching this error condition.

### DIFF
--- a/source/util.cpp
+++ b/source/util.cpp
@@ -1661,7 +1661,7 @@ void GetVirtualDesktopRect(RECT &aRect)
 
 
 DWORD GetEnvVarReliable(LPCTSTR aEnvVarName, LPTSTR aBuf)
-// Returns the length of what has been copied into aBuf.
+// Returns the length of what has been copied into aBuf, not including the null terminator.
 // Caller has ensured that aBuf is large enough (though anything >=32767 is always large enough).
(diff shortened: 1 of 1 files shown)


Lexikos:
> GetEnvVarReliable() does some casting between size_t and DWORD, which is
problematic because size_t is eight bytes on Win32.

I don't see what you're getting at. There was no casting, only implicit conversion from DWORD to size_t, which isn't inherently problematic.

Some static analysis tools will tell us that `length + 1` might overflow before being assigned to a larger type, but the possibility of an environment variable of 0xfffffffe characters (producing a return value of 0xffffffff) is not a realistic concern.

> This case can be encountered when the buffer is precisely the length of the required output

The buffer size is always 32767, which Microsoft states is the maximum size limit:

> An environment variable has a maximum size limit of 32,767 characters, including the null-terminating character.

... but testing (on Windows 11) shows that environment variables can be larger than this.

GetEnvVarReliable is only used for `A_ComSpec` and the deprecated auto-env retrieval mechanism which is disabled by `#NoEnv`.

PR #262, opened 2021-11-29
fix linkage redefinition on DoKeyDelay()

mikeblas:
This fix addresses the redefinition of DoKeyDelay() from extern to static;
it worked out, but was pretty dangerous. The change also removes the dynamic
default value from DoKeyDelay() and replaces it with an alternate
DoDefaultKeyDelay() function so the default of the duration is explicit at
calling sites.

### DIFF
--- a/source/keyboard_mouse.cpp
+++ b/source/keyboard_mouse.cpp
@@ -22,6 +22,7 @@ GNU General Public License for more details.
 #include "window.h" // for IsWindowHung()
 
 
+
 // Added for v1.0.25.  Search on sPrevEventType for more comments:
 static KeyEventTypes sPrevEventType;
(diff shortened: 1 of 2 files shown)


Lexikos:
> This fix addresses the redefinition of DoKeyDelay() from extern to static; 
it worked out, but was pretty dangerous.

Please elaborate. What redefinition? What danger?

> ... so the default of the duration is explicit at calling sites.

I don't see what value that adds. It is already obvious that the default duration will be used when no duration is specified. What sets this apart from the use of optional parameters for any other function?

Lexikos:
I accidentally pushed a commit which merged this, but it is not actually in the current `master`.

PR #264, opened 2022-04-04
Changing the way CheckScriptTimers() resloves next_timer

HelgeffegleH:
**Reason**, `next_timer` might be deleted via `timer.mCallback`'s `__delete` routine.
**About**, `DeleteTimer` outputs `next_timer` on request if it is still present in the linked list after `delete timer` is done. 

**Examples**, explicitly deleted via `SetTimer`
```autohotkey
class x {
	__delete()=>settimer(y, 0)
	call()=>0
}
f() => ''
y := f.bind()
settimer x(), -1
settimer y, -50
``` 
And indirectly via interruption,
```autohotkey
class x {	
	__delete()=>sleep(100)
	call()=>0
}
settimer x(), -1
settimer x(), -50
```

### DIFF
--- a/source/application.cpp
+++ b/source/application.cpp
@@ -1625,12 +1625,23 @@ bool CheckScriptTimers()
 
 		// Resolve the next timer only now, in case other timers were created or deleted while
 		// this timer was executing.  Must be done before the timer is potentially deleted below.
-		next_timer = timer.mNextTimer;
+
+		// Cannot unconditionally resolve the next timer as:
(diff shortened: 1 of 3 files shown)


Lexikos:
Thanks for your work. I've come up with a simpler solution, which I also found easier to apply to v1. I couldn't reproduce any crashes in release mode, but your test scripts did produce access violations in debug mode without the fix, and exited cleanly with the fix.

PR #266, opened 2022-05-02
Changing VK_ESCAPE to reset hotstring detection

HelgeffegleH:
New, `esc` causes the hotstring recognition process to reset.

Before, pressing `esc` while collecting input for hotstrings, `chr(27)` was stored in the hotstring buffer.

Reason, `esc` is more commonly used to _abort_ rather than producing _text_.

_edit_, fixed comment, to avoid alerting the word repetition police

### DIFF
--- a/source/hook.cpp
+++ b/source/hook.cpp
@@ -2536,6 +2536,9 @@ bool CollectInput(KBDLLHOOKSTRUCT &aEvent, const vk_type aVK, const sc_type aSC,
 	{
 		switch (aVK)
 		{
+		case VK_ESCAPE:		// Reset hotstring detection if user presses esc, 
+							// as this commonly indicates a wish to "abort" rather than producing a character.
+			char_count = 0; // To avoid storing VK_ESCAPE in g_HSBuf below.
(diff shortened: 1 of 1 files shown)


Lexikos:
What benefit does this provide? As far as I can tell, it only prevents hotstrings which contain `Chr(27)` from working (a niche feature, but worth keeping if there are no drawbacks).

With a hotstring `::btw::by the way`,
  - if you type `btw{esc}{space}`, the escape character will prevent it from triggering, the same as if a reset was performed.
  - if you type `something{esc}btw{space}`, it will trigger, the same as if a reset was performed, and the same as if you typed `something{space}btw{space}`.

HelgeffegleH:
Yeah, it seems you are right. I think I was just focused on removing, what I found to be an unintuitive feature, i.e., _Esc_ producing a _character_, I forgot to think about if it had any real use.

Anyways, if some one comes up with a good reason for _Esc_ to reset, I guess a hotkey like, `~esc::hotstring 'reset'`,  would do it.

I will close this for this reason.

PR #267, opened 2022-05-02
Converting ACT_TOOLTIP to BIF_ToolTip

HelgeffegleH:
**New**, `ToolTip` now returns `tip_hwnd` when showing a tooltip. Validates numeric parameters.

**Reason**, BIF implies slightly improved param verification/error type and returning `tip_hwnd` might be useful for some purposes.

**About**, throws win32 error if `CreateWindowEx` fails, this is probably very unlikely but guarantees that the return value is a valid hwnd. When destroying a tooltip, the function returns blank. Added `malloc` check in `ToolTipTextEquals`.

### DIFF
--- a/source/defines.h
+++ b/source/defines.h
@@ -615,7 +615,7 @@ enum enum_act {
 // All others are not included in g_act, and are only used with BIF_PerformAction:
 // ================================================================================
 , ACT_EXITAPP
-, ACT_TOOLTIP, ACT_TRAYTIP
+, ACT_TRAYTIP
 , ACT_RUNAS, ACT_DOWNLOAD
 , ACT_SEND, ACT_SENDTEXT, ACT_SENDINPUT, ACT_SENDPLAY, ACT_SENDEVENT
 , ACT_SENDMODE, ACT_SENDLEVEL, ACT_COORDMODE, ACT_SETDEFAULTMOUSESPEED
--- a/source/script.cpp
+++ b/source/script.cpp
@@ -325,7 +325,7 @@ FuncEntry g_BIF[] =
 	BIF1(SysGetIPAddresses, 0, 0),
 	BIF1(Tan, 1, 1),
 	BIFA(Thread, 1, 3, ACT_THREAD),
-	BIFA(ToolTip, 0, 4, ACT_TOOLTIP),
+	BIF1(ToolTip, 0, 4),
 	BIF1(TraySetIcon, 0, 3),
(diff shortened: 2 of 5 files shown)


Lexikos:
I generally don't like to add a return value which is 1) easily obtained by some other means, and 2) rarely of much use; because it precludes the future use of the return value for something more useful. I think the last example of this was a suggested return value for the `Hotkey` function (I forget what), where I imagine in future it could return an object that provides more information and control. After reading this PR, I considered how something similar might apply to ToolTip.

However, I've concluded that the current ToolTip isn't suited to returning an object for more control, because:
  - It shows the tooltip immediately (or `ToolTip()` removes it). Requiring the caller to work around this (if needed) makes any built-in functionality less appealing.
  - It replaces a global tooltip at the given index.

If tooltip functionality is ever extended, it could be done via a new class with a new name, or a module that must be imported.

---

If we're adding the HWND return value to make it slightly easier for scripts to customize the tooltip, it might be worthwhile adding `ti.hwnd = g_hWnd;` as well to enable the script to handle the tooltip's notifications. I considered doing this recently, but chose not to because:
  - Creating your own tooltip window manually offers more flexibility, and really isn't difficult. Making full use of a tooltip's capabilities requires utilizing structs regardless of whether you use ToolTip or CreateWindowEx (or even Gui.AddCustom and SetParent) to create the window.
  - Using ToolTip implies certain limitations or required workarounds, since it shows the window automatically.
  - For scripts that also create their own tooltip windows, receiving unexpected tooltip notifications for ToolTip might cause issues.

My notes about `ti.hwnd` are:
> Despite what the doc says, notifications are not sent to "the owner window"; they are sent to TOOLINFO::hwnd, which is "Handle to the window that contains the tool".  This means we could set the tooltip's owner to some other window to keep it on top, while still having notifications sent to us.

---

A general note about PR style: I prefer a verb in present tense (Convert) for commit messages that describe a change being made that doesn't affect users, or past tense (Fixed, Changed, Added) for a change that affects users and therefore will be copied into the changelog (or forum); i.e. not the present participle (Converting).
  - One reason is that (in general) omitting a few characters makes it less likely to be truncated when viewed in the commit history on GitHub or in other tools.
  - Another reason is that the change isn't in progress; it is completed by the commit.
  - I think the present tense is more common; e.g. "Merge pull request ..."

HelgeffegleH:
Thanks for the feedback. I will take this into account in the future.

PR #275, opened 2022-05-05
v2: Added Count property to Array.

jeeswg:
## Introduction

A `Count` property for arrays, similar to AHK v1's `Count` method.
Consistent with AHK v2's `Length` property, and AHK v1's `Length` method.
It counts the number of array keys that have a value.

## Rationale

For validating parameters in variadic functions.
To check if an array is dense or not.
A relatively common task, which requires a few code lines, and is slow, if done as AHK code.
Facilitates conversion: any use of `Count` in AHK v1, has a functional equivalent in AHK v2.
`Array`, `Map` and a possible `Range` would all have a `Count` property.

## Test code

```

;test code: Array.Count (AHK v2)

oArray := [1,,,, 5]
oArray.Length := 10
MsgBox(oArray.Count " " oArray.Length)

oArray.Length := 4
MsgBox(oArray.Count " " oArray.Length)

vNum := 1000000
oArray := []
oArray.Capacity := vNum
oArray.Length := vNum
Loop vNum
{
	if Random(0, 1)
		oArray[A_Index] := 1
}

vTickCount2 := A_TickCount
vCount2 := oArray.Count
vTickCount2 := A_TickCount - vTickCount2

vTickCount1 := A_TickCount
vCount := 0
for vKey, vValue in oArray
	vCount += IsSet(vValue)
vTickCount1 := A_TickCount - vTickCount1

MsgBox(vTickCount1 "`r`n" vTickCount2)
MsgBox(vCount "`r`n" vCount2 "`r`n" oArray.Length)

return

/*
;oMap := Map()
;MsgBox(oMap.Count)
;oMap.Count := 3

;---------------------------
;test array count.ahk
;---------------------------
;Error:  Property is read-only.

;Specifically: Count

;	Line#
;	001: oMap := Map()
;	002: MsgBox(oMap.Count)
;--->	003: oMap.Count := 3
;	005: oArray := [1,,, 5]
;	006: MsgBox(oArray.Length)
;	007: MsgBox(oArray.Count)
;	008: Exit

;Try to continue anyway?
;---------------------------
;Yes   No
;---------------------------

;check for method error message:
MsgBox(oArray.Length())
MsgBox(oArray.Count())

oArray := [1,,,, 5]
MsgBox(oArray.Length)
MsgBox(oArray.Count)

;check for property set error message:
oArray.Length := 10
oArray.Count := 10
(oMap := Map()).Count := 3

MsgBox("done")
return
*/

```

### DIFF
--- a/source/script_object.cpp
+++ b/source/script_object.cpp
@@ -1910,6 +1910,7 @@ ObjectMember Array::sMembers[] =
 	Object_Property_get_set(__Item, 1, 1),
 	Object_Property_get_set(Capacity),
 	Object_Property_get_set(Length),
+	Object_Property_get(Count),
 	Object_Member(__New, Invoke, M_Push, IT_CALL, 0, MAXP_VARIADIC),
 	Object_Method(__Enum, 0, 1),
 	Object_Method(Clone, 0, 0),
@@ -1953,6 +1954,14 @@ void Array::Invoke(ResultToken &aResultToken, int aID, int aFlags, ExprTokenType
 			return;
 		}
 		_o_return(aID == P_Capacity ? Capacity() : Length());
+
+	case P_Count:
+	{
+		__int64 count = 0;
+		for (__int64 i = 0; i < mLength; ++i)
+			count += (mItem[i].symbol != SYM_MISSING) ? 1 : 0;
+		_o_return(count);
+	}
 
 	case M_InsertAt:
 	case M_Push:
--- a/source/script_object.h
+++ b/source/script_object.h
@@ -564,6 +564,7 @@ public:
 		P___Item,
 		P_Length,
 		P_Capacity,
+		P_Count,
 		M_InsertAt,
 		M_Push,
 		M_RemoveAt,


Lexikos:
I think you missed the point of the `Object_Property_get_set` macro. Read-only properties should not have a *setter*. Without a setter, the read-only error is thrown automatically.

> A relatively common task

Can you point to real-world examples?

jeeswg:
Thank you, I see now, there is also an `Object_Property_get` macro.

Examples for checking the count:
- Error checking/validation. In functions that do something to every array member, you might want to bail out if the array isn't dense.
- Loops are slow. If the array is dense, you can take a fast approach, if it isn't, you use the slower loop approach.
- Buffer capacity. For a sparse array, `Count` lets you prepare the exact capacity in advance.
- String capacity. For a sparse array, `Count` lets you prepare `VarSetStrCapacity` reasonably accurately in advance.
- When checking for omitted parameters, knowing the array is dense, or that every parameter was passed, may let you bypass various checks.

Other points:
- Sometimes you write code, and you intended the meaning `Count`, not `Length`. For correctness/clarity.
- In some quick code, I could rely on an Array/Map/RegExMatchInfo/Scripting.Dictionary/'Range'/custom Array object, having a `Count` property. E.g. when copying code to reuse, or deciding partway through you're going to change the object type you're using.
- Other languages or existing AHK code that uses a `Count` property. So, easy conversion/maintenance across scripts/languages.
- Force of habit while typing, you don't have to backspace and type in Length.

jeeswg:
Here's a fixed version of the branch:
https://github.com/jeeswg/AutoHotkey_L/tree/arraycount_fix

RaptorX:
I think the divide between `Map`._count_ and `Array`._length_ is great and should be kept like that.

jeeswg:
@RaptorX: `Array.Count` is `Array.Length` minus the number of unset keys.
In AHK v1 `Array.Count` and `Map.Count` both exist (under the combined object type).
If in AHK v2, `Map.Count` (and `MyCustomClass.Count`) existed, but `Array.Count` didn't, that would break a lot of code, and require workarounds to fix.
Also, `Array.Count` beats using a for loop every time you want to check the array has no gaps, or want to count the keys/gaps.

iseahound:
`array.count(x?)` is normally a method to find all instances of an element `x`, where if no parameter is passed, returns all the values that are not unset/null.

jeeswg:
@iseahound: Please state any languages where `.Count`/`.Count()` does/doesn't filter in the way as you describe.
Please state any languages with equivalent filter functionality, that use a different name.

Your point would be equally applicable to `Map.Count`.
If you wanted to replace `Map.Count` with `Map.Count()`, or have both, you should probably write about it on the forum.
If such filter functionality were implemented, I could think of many acceptable possibly more expressive names, including 'Filter'.

Lexikos:
> Examples for checking the count:

These are all hypothetical. You haven't posted any real-world examples.

> Error checking/validation. In functions that do something to every array member, you might need to throw if the array isn't dense.

Functions that do something with every array member would generally need to enumerate over the array to achieve their main purpose, and can react to one element having no value, without having to check in advance.

> Buffer capacity. For dense and sparse arrays, Count lets you prepare the exact capacity in advance.

In the specific case where unset elements should be completely omitted from whatever the buffer represents. Seems unusual.

> String capacity. For dense and sparse arrays, Count lets you prepare VarSetStrCapacity reasonably accurately in advance.

In the specific case where elements are approximately the same size, and again, unset elements should be completely omitted. Seems unusual.

> Sometimes you write code, and you intended the meaning Count, not Length. For correctness/clarity.

No, I don't. If there are unset elements, they must be handled during enumeration, and the Count is not usable for calculating or validating an array index. It does not even reflect the number of iterations that the enumerator will give.

> In AHK v1, Array.Count and Map.Count both exist (under the combined object type).

No, Object.Count() exists.

> If in AHK v2, Map.Count (and MyCustomClass.Count) existed, but Array.Count didn't, that would break a lot of code, and require workarounds to fix.

No.

> Also, Array.Count beats using a for loop  ...

As a user, doing anything with a ready-made, available-as-standard function beats having to add a function to the script, at any given moment. This is not compelling for me.

---

This property is not something I want to be a standard part of Array at this time.

If this functionality is added at any point, I would prefer to make it a method to help discourage users from unnecessarily favouring `.Count` over `.Length`. As you said, loops are slow; `.Count` is implemented with a loop, with cost proportionate to the size of the array, while `.Length` just returns a variable. It could be optimized for common cases (i.e. arrays keeping track of whether they have unset elements), but that would come at the cost of complexity and code size.

jeeswg:
### Uses

Check an array is dense before proceeding:
- array fill: fill unset keys with a value
- StrJoin
- ChrUnused(oParams*) (find a character not in oParams*, to use as a temporary delimiter)
- Array.Reverse, Array.Sort, Array.Swap
- Max/Min, StrMax/StrMin
- ArrToVars(oArray, oByRefOutVars*) (check if array or out variables array is dense)
- get every nth array item
- rearrange array based on a pattern (e.g. '132' makes [11,22,33] into [11,33,22])
- join/fold a range of array items (replacing multiple keys with one key)
- (check if an array is empty)
- (delete items from a dense/sparse array, loop until array is empty)
- (sometimes an array must be dense, other times a dense array can use a built-in function whereas a sparse array must use a custom loop)

Other points:
- Print object: `.Count` would work on arrays *and* maps (you wouldn't need extra code to check for and handle an array). (Almost all the common types I use have a `.Count` property e.g. Array (AHK v1)/custom 0-based array/Map/RegExMatch/Scripting.Dictionary.)
- It simplifies functions, and adds safety, insisting that all keys have values. (A user could provide any manner of arrays with unusual contents.)

### General points

>  It could be optimized for common cases (i.e. arrays keeping track of whether they have unset elements), but that would come at the cost of complexity and code size.

I considered this.
I felt that keeping track of the count was not worthwhile.
And that the documentation could state that the count was generated on-the-fly, not stored.

>  loops are slow; `.Count` is implemented with a loop, with cost proportionate to the size of the array

I found the loop very slow in AutoHotkey, but fine in C++.

(I had wondered if AHK v2.1 could somehow speed up AHK loops, for all tasks, or else a subset of tasks, e.g. a subset of AHK/C++ code converted to machine code. E.g. reverse UInts in a binary buffer.)

(Also possible: avoid for loops / Loop Parse by using: StrSplit/StrJoin, and variants of Arr.Map/Arr.Filter/Arr.Partition/Arr.FoldLeft/Arr.FoldRight.)

> I would prefer to make it a method to help discourage users from unnecessarily favouring `.Count` over `.Length`.

I'd recommend, in the documentation for `Array.Count`, to encourage users to use `Array.Length`, and clarify the difference.

Making `.Count` a method, would:
- make it more verbose (by requiring parentheses)
- add mental load (since other types have it as a property)
- necessitate 'IsArray' checks (for code/functions checking the count, that can handle multiple types)

`File.__Handle` later became `File.Handle`.
Many types already have `.Count`, AHK v1 'arrays' already have it.
I err towards informing the user, in documentation and error messages, and not 'patronising' them or forcing their hand.

### Verbosity and Speed

With `.Count`, there are several advantages:
- The code is much faster.
- The code is shorter, you even avoid a pair of curly braces.
- Where possible, it's clearer/tidier to do error checks at the start.

```
MsgBox(SimpleStrJoin1(",", ["a","b","c"]*))
MsgBox(SimpleStrJoin2(",", ["a","b","c"]*))

;simpler:
SimpleStrJoin1(vPad, vItem, oArray*)
{
	if (oArray.Length != oArray.Count)
		throw Error("", -1)

	for _, vValue in oArray
		vItem .= vPad vValue
	return vItem
}

;more complex:
SimpleStrJoin2(vPad, vItem, oArray*)
{
	for _, vValue in oArray
	{
		if !IsSet(vValue)
			throw Error("", -1)
		vItem .= vPad vValue
	}
	return vItem
}
```

In general it could be prudent to add this to every custom function that uses arrays, it's much better if it's short:
```
;short:
	if (oArray.Length != oArray.Count)
		throw Error("", -1)

;long:
	for _, vValue in oArray
	{
		if !IsSet(vValue)
			throw Error("", -1)
	}
```

And for counting, you get significantly shorter code:
The advantage compounds if you need to check the count in multiple places.
```
;short:
	vCount := oArray.Count

;long:
	vCount := 0
	for _, vValue in oArray
		vCount += IsSet(vValue)
```

PR #271, opened 2022-05-05
v2: Added ShortcutKey parameter to FileGetShortcut.

jeeswg:
## Introduction

Adds a ShortcutKey parameter to FileGetShortcut:
```
FileGetShortcut(LinkFile [, &OutTarget, &OutDir, &OutArgs, &OutDescription, &OutIcon, >>>&ShortcutKey<<<, &OutIconNum, &OutRunState])
FileCreateShortcut(Target, LinkFile [, WorkingDir, Args, Description, IconFile, ShortcutKey, IconNumber, RunState])
```
Note: `FileGetShortcut` and `FileCreateShortcut` would thus have the same number of parameters, although the `LinkFile` and `Target` parameters would still have a different order.
Note: `FileCreateShortcut` can still only create `Ctrl+Alt+XXX` shortcut keys. I only noticed that recently, and will look at changing that in a new PR.

## Documentation
```
FileGetShortcut(LinkFile [, &OutTarget, &OutDir, &OutArgs, &OutDescription, &OutIcon, &ShortcutKey, &OutIconNum, &OutRunState])
```
For example, the hotkey `CTRL + ALT + Q` would be returned as `^!q` in a similar syntax to hotkeys.

## Implementation

I couldn't find any definitive information on the meaning of `HOTKEYF_EXT`.

Here's the link for `GetHotkey` method:
[IShellLinkA::GetHotkey (shobjidl_core.h) - Win32 apps | Microsoft Docs](https://docs.microsoft.com/en-us/windows/win32/api/shobjidl_core/nf-shobjidl_core-ishelllinka-gethotkey)

Although this link: [TIP:Windows Search deskbar shortcut / hotkey (XP/Srv2003)](https://social.msdn.microsoft.com/Forums/en-US/e85fd2e1-a095-414a-837e-afe7ac7f0a97/tipwindows-search-deskbar-shortcut-hotkey-xpsrv2003), suggests it means the Windows key, and that's how I've implemented it.

The documentation could simply state: the `ShortcutKey` string will contain a '#' (as one of the leading characters indicating modifier keys) if the extended key flag is set.

[EDIT:]
[[MS-SHLLINK]: Shell Link (.LNK) Binary File Format | Microsoft Docs](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-shllink/16cb4ca1-9339-4d0c-a68d-bf1d6cc0f943)
Under HighByte, if references Shift/Control/Alt, but states that the value should never include 0x8.
[...]

PR #277, opened 2022-05-05
v2: Changed FileCreateShortcut's ShortcutKey parameter to be more permissive.

jeeswg:
## Documentation

The `ShortcutKey` parameter now supports strings that are similar to hotkeys: one or more modifiers ('^#!+') then a key name.
E.g. specify `^!a` for the hotkey `CTRL + ALT + A`.
E.g. specify `^+!a` for the hotkey `CTRL + SHIFT + ALT + A`.
The modifier characters can be in any order.
There must be at least one modifier and exactly one key name.

Note: using `#` will set the extended key flag in the shortcut file.

If converting scripts from AHK v1 to AHK v2, the leading `^!` (for `CTRL + ALT`) must now be specified explicitly.
E.g. in AHK v1: `a` would specify `CTRL + ALT + A`, in AHK v2 it specifies `A` with no modifiers, which would raise an error.
E.g. the equivalent to `a` in AHK v1, is `^!a` in AHK v2.

See also: #271.

## Test code

```
;test code: FileCreateShortcut(): improved ShortcutKey parameter (AHK v2)

Target := A_ScriptDir "\MyTxt.txt"
LinkFile := A_ScriptDir "\MyLink " A_Now ".lnk"
;WorkingDir := A_ScriptDir
WorkingDir := ""
;Args := "arg1 arg2 arg3"
Args := ""
Description := "my description"
IconFile := A_AhkPath
;ShortcutKey := "a" ;Error:  Invalid hotkey.
ShortcutKey := "^!a"
;ShortcutKey := "^#!+a"
IconNumber := 3
RunState := 7
if !FileExist(Target)
	FileAppend("", Target, "UTF-8")
if !FileExist(LinkFile)
	FileCreateShortcut(Target, LinkFile, WorkingDir, Args, Description, IconFile, ShortcutKey, IconNumber, RunState)

;note: Properties dialog closes when script closes:
Run("properties " Chr(34) LinkFile Chr(34))
MsgBox("Press OK to close Properties dialog.")
return
```

### DIFF
--- a/source/script_autoit.cpp
+++ b/source/script_autoit.cpp
@@ -1265,13 +1265,36 @@ ResultType Line::FileCreateShortcut(LPTSTR aTargetFile, LPTSTR aShortcutFile, LP
 			psl->SetIconLocation(aIconFile,  icon_index - (icon_index > 0 ? 1 : 0)); // Convert 1-based index to 0-based, but leave negative resource IDs as-is.
 		if (*aHotkey)
 		{
-			// If badly formatted, it's not a critical error, just continue.
-			// Currently, only shortcuts with a CTRL+ALT are supported.
-			// AutoIt3 note: Make sure that CTRL+ALT is selected (otherwise invalid)
-			vk_type vk = TextToVK(aHotkey);
-			if (vk)
+			WORD mods = 0;
+			LPTSTR cp = aHotkey;
+			for (;; ++cp)
+			{
+				// Note: the last char is always literal, not a modifier, even if it's one of '^#!+'.
+				if (!cp[1])
+					break;
+				else if (*cp == '^')
+					mods |= HOTKEYF_CONTROL;
+				else if (*cp == '#')
+					mods |= HOTKEYF_EXT; // Windows key?
+				else if (*cp == '!')
+					mods |= HOTKEYF_ALT;
+				else if (*cp == '+')
+					mods |= HOTKEYF_SHIFT;
+				else
+					break;
+			}
+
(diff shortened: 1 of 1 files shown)


Lexikos:
> the leading ^! (for CTRL + ALT) must now be specified explicitly.

This appears to be a breaking change.

> AutoHotkey v2 has been in beta since July 2021. Changes prior to the final v2.0.0 release should be of a minor nature (such as improvements to the error and warning dialogs), and should not affect the functionality of scripts. In other words, future releases are expected to be backward-compatible. 
> https://www.autohotkey.com/v2/

jeeswg:
Indeed, this would be a breaking change.
However, any function call using the old syntax, would generate an error message.

The following recent PRs break changes. Although I think it is essential to fix the AHK v1 syntax problems:
- This PR.
- #271. But adding `ShortcutKey` as the last parameter wouldn't be a breaking change.
- #293. In my view, the current AHK v2 `ControlClick` is suboptimal.
- #286. If carried through to AHK v2, this PR would restore the classic `NumPut` parameter order. I suggested a separate `NumPutPairs` for the current AHK v2 functionality.

There have already been some breaking changes since 2.0-beta.1, and I think all of the AHK programmers happily appreciate that more will occur.

Lexikos:
Again,

> Changes prior to the final v2.0.0 release should be of a minor nature (such as improvements to the error and warning dialogs), and should not affect the functionality of scripts. 

Overt changes to documented syntax will not be considered until v3.

PR #282, opened 2022-06-03
Adding ability to specify arbitrary expressions for parameter defaults.

HelgeffegleH:
**New**
```autohotkey
f(param := expr) { ... }
```
where _expr_ is evaluated and its result is assigned to _param_ each time _param_ is omitted when calling _f_. _expr_ is evaluated in an unnamed nested function in _f_.  All nested functions are added first in _f's_ function body (after the opening brace). This implies the behaviour and scope of the expression evaluation. A notable difference being that if _f_ is a method/prop, the keyword _super_ is supported.

**About**
`FuncParamDefaults::PARAM_DEFAULT_EXPR` is added to indicate that a parameter uses an expression for its value if omitted. This is detected in `DefineFunc` and there the expression string is temporarily stored in `FuncParam::default_str`. `AddLine` then calls `DefineParamExpr` for each function which has any parameters with default expressions. `DefineParamExpr` inserts the expressions as nested functions.

Previous code for handling default values are mostly unchanged, these are kept as optimisations. The debugger have not been considered.

**Pro**

* Convenience, flexibility and brevity (if not misused).
* Minor performance in some cases (guess).
* You can reference **previously assigned** parameters, eg `shape(width, height := width) => ...`
* Simple implementation (a weak argument I suppose, but the only reason I even considered implementing it)

**Con**

* You can reference static variables, but they risk not being assigned a value. Typical trap,
```autothotkey
f(x := y) { 
    static y := 0
}
f    ; y not assigned
f 1  ; no issue
f    ; now it works
```
* Although it makes sense that you cannot reference a parameter (without a run time error) to the right of _expr_, eg, in `f(x := y, y := '')` `y` is not assigned a value when `x` is being assigned, this might be surprising if `y` was passed a value as in `f(, 0)`. 
* Any reference to `A_ThisFunc` in _expr_ results in a blank string, which is counter-intuitive. I guess this is easy to fix.
* For the same reason `A_ThisFunc` is blank, error reporting is affected when _expr_ fails. 
* Although the implementation is simple, it _piles on_ the code base, without improving it. Also could imply incompatibility and/or inconsitency with a theoretical future `#directive expr` (although that is probably natural).
* __edit__, Added handling of `f(x := "y" z)` which isn't _missing comma_ but concatenation. I suppose that is a bit unfortunate. For example, if you do miss a comma in `f(x := "y" z := "w")` you will not be notified, unless you attempt to call _f_, passing a second parameter.

**Neutral**

* _expr_ is evaluated each time the parameter is omitted, so these aren't _default values_, which might be unexpected in case of something like `f(x:=[])`, where `[]` will be a new array each time _x_ is omitted.

Likely, I forgot something.

### DIFF
--- a/source/script.cpp
+++ b/source/script.cpp
@@ -5042,7 +5042,49 @@ inline ActionTypeType Script::ConvertActionType(LPCTSTR aActionTypeString)
 	return ACT_INVALID;  // On failure to find a match.
 }
 
-
+ResultType Script::DefineParamExpr(UserFunc* aFunc)
+{
+	// Called for each function which defines at least one "param := expr".
+	// For a script function 
+	// f(x:=expr) {
+	// }
+	// this function adds,
+	// f(x:=expr) {
+	//	<unnamed>() {
+	//		return expr
+	//	}
+	// }
+	auto count = aFunc->mParamCount;
+	auto current_func = g->CurrentFunc;
+	auto cls = g->CurrentFunc->mClass;
+	
+	for (int i = 0; i < count; ++i)
+	{
+		auto &param = aFunc->mParam[i];
+		if (param.default_type == FuncParamDefaults::PARAM_DEFAULT_EXPR)
+		{
+			auto param_func = g->CurrentFunc = AddFunc(_T(""), 0, nullptr); // No name to avoid creating a variable.
+			if (cls)
+			{	// To allow using "super" in a method/property parameter expression.
+				// Also see SYM_SUPER in ExpandExpression()
+				cls->AddRef();
+				param_func->mClass = cls;
+			}
+			auto expr_str = param.default_str;
+			
+			mNextLineIsFunctionBody = false;
+			if (!AddLine(ACT_BLOCK_BEGIN)
+				|| !ParseAndAddLine(expr_str, ACT_RETURN)
+				|| !AddLine(ACT_BLOCK_END))
+				return ScriptError(ERR_OUTOFMEM);
+			g->CurrentFunc = current_func;
+			param.default_expr = param_func;
+			free(expr_str); // allocated in DefineFunc with _tcsdup
+		}
+	}
+	mNextLineIsFunctionBody = true;
+	return OK;
+}
 
 ResultType Script::AddLine(ActionTypeType aActionType, LPTSTR aArg[], int aArgc, LPTSTR aArgMap[], bool aAllArgsAreExpressions)
 // aArg must be a collection of pointers to memory areas that are modifiable, and there
@@ -5068,6 +5110,15 @@ ResultType Script::AddLine(ActionTypeType aActionType, LPTSTR aArg[], int aArgc,
 		mLastHotFunc = nullptr;
 	}
 
+	// Check if this is a function which needs to define any parameter default expresssions,
+	// eg f(param := expr)
+	if (mNextLineIsFunctionBody)
+	{
+		ASSERT(g->CurrentFunc);
+		if (g->CurrentFunc->mHasParamExpr
+			&& !DefineParamExpr(g->CurrentFunc))
+			return FAIL;
+	}
 	DerefList deref;  // Will be used to temporarily store the var-deref locations in each arg.
 	ArgStruct *new_arg;  // We will allocate some dynamic memory for this, then hang it onto the new line.
 	LPTSTR this_aArgMap, this_aArg;
@@ -5966,8 +6017,21 @@ ResultType Script::DefineFunc(LPTSTR aBuf, bool aStatic, bool aIsInExpression)
 			}
 			else // A default value other than a quoted/literal string.
 			{
-				if (!(param_end = StrChrAny(param_start, _T(", \t=)")))) // Somewhat debatable but stricter seems better.
-					return ScriptError(ERR_MISSING_COMMA, aBuf); // Reporting aBuf vs. param_start seems more informative since Vicinity isn't shown.
+				
+				auto len_expr = FindExprDelim(param_start, ',');
+				if (param_start[len_expr] == '\0')
+				{
+					len_expr = FindExprDelim(param_start, ')');
+					if (param_start[len_expr] == '\0')
+					{
+						if (!(param_end = StrChrAny(param_start, _T(", \t=)")))) // Somewhat debatable but stricter seems better.
+							return ScriptError(ERR_MISSING_COMMA, aBuf); // Reporting aBuf vs. param_start seems more informative since vicinity isn't shown.
+					}
+					else
+						param_end = param_start + len_expr;
+				}
+				else
+					param_end = param_start + len_expr;
 				value_length = param_end - param_start;
 				if (value_length > MAX_NUMBER_LENGTH) // Too rare to justify elaborate handling or error reporting.
 					value_length = MAX_NUMBER_LENGTH;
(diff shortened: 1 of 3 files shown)


Lexikos:
Thanks!

> Minor performance in some cases (guess).

I would guess that `IsSet(p) && p := expr` would perform better than calling a closure and performing the assignment from C++, but I don't care enough to check.

> You can reference static variables, but they risk not being assigned a value. 

I think this is obvious enough, and in line with any reference to a static variable above its declaration. Static initializers are expected to evaluate only if and when they are reached. It is possible for the same static variable to have multiple initializers.

> you cannot reference a parameter (without a run time error) to the right of *expr*

This is a problem. I don't think it's clear from `f( , b)` / `f(left := right, right := expr) =>` that `right` accepts its value by the equivalent of `f(left := right, right := b)`. One would likely expect the caller's parameter list to be evaluated in full (which it is...), then omitted parameters to be initialized based on default expressions/values, without really considering how the caller's parameter values end up in parameter variables.

> Although the implementation is simple, it *piles* on the code base, without improving it. 

That's always a concern (and thanks for putting a name to it), although that is consistent with some other additions to the language. Things that get put off until they can be implemented more cleanly tend to be put off for a very long time, and can't be used in the meantime. Sometimes a hacked-in implementation helps to clarify the requirements for future revisions. And *interesting* new features help to motivate development...

> For example, if you do miss a comma in `f(x := "y" z := "w")` you will not be notified, unless you attempt to call *f*

If `z` isn't assigned inside `f` itself, would it be local to the closure? In that case, I suppose any read-reference to `z` inside `f` would trigger a load-time warning.

> Any reference to `A_ThisFunc` in *expr* results in a blank string

I haven't looked at your implementation yet (and probably won't for a while for the same reasons that I haven't implemented this myself; priorities and time constraints), but I would suggest deferring evaluation of the initializers until after all of the following are done:
  - All parameter values supplied by the caller have been assigned.
  - The function has been pushed onto the debugger's call stack.
  - `g->CurrentFunc` has been assigned.

HelgeffegleH:
>  would suggest deferring evaluation of the initializers until after all of the following are done:

Thanks for this, I have moved some code around. See latest commit. Note that I have not tested if this impacts _half-baked named paramters_, but _they_ should be assigned before default values/expressions are assigned/evaluated.

> I would guess that IsSet(p) && p := expr would perform better than calling a closure and performing the assignment from C++, but I don't care enough to check.

I made no performance tests. And your guess is probably correct. However, if a certain parameter is rarely omitted, you can avoid many `isset(p)` checks with a default expression/value 👍 .

Cheers.

Lexikos:
I think even with parameter default expressions, there is still a place for optional parameters without a default value. If nothing else, they make more sense where multiple functions accept the same optional parameter - i.e. one function can define a default value and the rest can just forward the parameter to that function.

Commit f12398c8 adds `fn(var?)` to either pass or omit `var` depending on whether it is set (it also adds `unset` for use with ternary/`&&`/`||` to conditionally omit parameters).

Commit b0cc608f adds `var ?? value`. C# has `var ??= value`, which could be used like `param ??= expr` at the top of the function, repeating only the parameter name; but I haven't taken the time to work out how to implement a short-circuiting compound assignment, so for now one can just use `param ?? expr` directly. I suppose that `param := param ?? expr` isn't much better than `IsSet(param) || param := expr`.

I was about to release it as v2.0-beta.6, but now I think I'll hold off until I've had time to review this pull request.

I had in my notes to experiment with implementing expression parameter defaults by simple preprocessing: for `f(param := expr)`, insert `IsSet(param) || param := expr` at the top of the function.

I think if the parser was more flexible, another option would be to parse `expr` as an expression and evaluate it in the context of the function. This is one of many reasons that I intend to focus on cleaning up/reorganizing/restructuring the code before adding more features, although I haven't found myself really doing that yet.

HelgeffegleH:
> I had in my notes to experiment with implementing expression parameter defaults by simple preprocessing: for f(param := expr), insert IsSet(param) || param := expr at the top of the function.

That would probably work well, be even simpler to implement and less likely to introduce any bugs. But assuming one keeps the current code for assigning default values, as optimisations, which I think would be desirable, there would be a (small) inconsistency for cases like `f(a:=b, b:=0)` and `f(a:=b, b:=non_optimised_expr)`, where the former would work if called when both parameters were omitted, but the latter wouldn't. It would not suffer from the issue with `A_ThisFunc` as this branch does, although that can probably be fixed, easily.

> _the other stuff_

Seem great, especially `var?` / `unset` will be very useful 👍

Lexikos:
> Also could imply incompatibility and/or inconsitency with a theoretical future `#directive expr` (although that is probably natural).

What is this hypothetical directive and how does it relate to the proposed change?

What are you thoughts about the consequences of using nested functions for the expressions?

HelgeffegleH:
> What is this hypothetical directive and how does it relate to the proposed change?

I see this was poorly presented, probably in part because it was poorly thought through in the first place. I was thinking about existing _#directives_, such as for example, `#include`, supporting arbitrary expressions. But I do not think that would relate to the proposed change, and I should have removed that comment.

> What are you thoughts about the consequences of using nested functions for the expressions?

My main concern would be if it turns out _quirky_. One would probably not expect _expr_ to be evaluated in a nested function if seeing it in use for the first time, rather than in the documentation. I had previously discarded this idea when (I think) before having one closure caused all other nested functions to also become closures, and perhaps other limitations with byref params being upvars. 

> If z isn't assigned inside f itself, would it be local to the closure? In that case, I suppose any read-reference to z inside f would trigger a load-time warning.

That is correct, and also presents another problem, which I hadn't thought of, namely that different _expr_ cannot share variables, such as `f(x := a:=1, y := a)`, unless `a` is local to `f`. I think it is a problem in terms of being unintuitive, rather than in terms of usefulness.

Another thing which I did not considered, and only came to think of from the idea of adding `!isset(p) && p:=expr`, is the case when having `critical` as the first line of a function (thread). Then adding `p:=expr` could impact behaviour.

Lexikos:
> having `critical` as the first line of a function (thread). 

I considered that as well and don't think it's a sufficient reason to avoid allowing expressions here. It is unavoidable that the expression would be evaluated before the first line of the function. This situation seems similar to `Suspend Permit` - having it as a directive or option of the hotkey is much better than looking ahead to the first line of the subroutine before it is even called. Perhaps it could be an attribute of the function itself, if not an option for whatever is starting the thread.

See 5e27478c (or 682144bd, which includes additional fixes) for an alternative implementation. It seems to be much simpler, with fewer issues. Explaining it in terms of the equivalent script code should be easy and accurate. I think `critical` is the only issue of note, and it can simply be documented.

source/script.cpp
```diff
@@ -5953,21 +6004,43 @@ ResultType Script::DefineFunc(LPTSTR aBuf, bool aStatic, bool aIsInExpression)
...
-			else // A default value other than a quoted/literal string.
-			{
-				if (!(param_end = StrChrAny(param_start, _T(", \t=)")))) // Somewhat debatable but stricter seems better.
-					return ScriptError(ERR_MISSING_COMMA, aBuf); // Reporting aBuf vs. param_start seems more informative since Vicinity isn't shown.
+			if (!is_default_str) // A default value or expression other than a quoted/literal string.
+			{	
+				auto len_expr = FindExprDelim(param_start, ',');
+				if (param_start[len_expr] == '\0')
+				{
+					len_expr = FindExprDelim(param_start, ')');
```
Lexikos, line 6033: Note that a single call to `FindExprDelim(param_start, 0)` would find the first delimiter that can't be part of the expression. It is used by ParseFatArrow, for instance, to find the end of the fat arrow function within a larger expression.

HelgeffegleH:
>  It seems to be much simpler, with fewer issues. Explaining it in terms of the equivalent script code should be easy and accurate. I think critical is the only issue of note, and it can simply be documented.

Very good 👍. Maybe it would be worth putting all parameter initialisers on the same line, eg `p1 ?? p1:= expr1, p2 ?? p2:= expr2`.

PR #284, opened 2022-07-05
v1: Added binary integer support to Format.

jeeswg:
Replaced by #292.

## Documentation

For the `Format` function:
The `diouxX/eEfgGaA/cCp/b` options treat any input string that starts with `0b` as a binary integer.
The `b` option converts a number to a binary integer.

'b' v. 'B':
Whereas 'x' and 'X' convert to hex, only 'b' (and not 'B') converts to bin. This could be changed if desired.
(I chose lower-case only, to be consistent with other lower-case only options, e.g. 'o' but not 'O' for octal, and in case 'B' might ever be useful for something else.)

## Implementation

If `0b` prefix handling were added to AHK v2, the 5-line code in 4 places (`diouxX/eEfgGaA/cCp/b`) could be replaced with a single line again, in AHK v2.

One omission. `{:04X}` allows you to convert dec to hex *and* add leading zeros.
There isn't currently a binary equivalent.
(Such functionality was added in #292.)
Here's a workaround:
```
Format("0x{:04X}", 255) ;'0x00FF'
Format("0b{:04b}", 3) ;'0b{:04b}' (not '0b0011')
Format("0b{:04}", Format("{:b}", 3)) ;'0b0011' (workaround)
```

## Test code

```

;test code: Format (binary integer support) (AHK v1)

Assert(Format("0x{:X}", 15), "0xF")

Assert(Format("{:u}", "0b000"), 0)
Assert(Format("{:u}", "0b111"), 7)
Assert(Format("{:u}", "0b00000111"), 7)
Assert(Format("{:i}", "0b1111111111111111111111111111111111111111111111111111111111111111"), -1)
Assert(Format("{:u}", "0b1111111111111111111111111111111111111111111111111111111111111111"), "18446744073709551615")

Assert(Format("{:b}", 0), 0)
Assert(Format("{:b}", 7), 111)
Assert(Format("0b{:08}", Format("{:b}", 7)), "0b00000111")
Assert(Format("{:b}", -1), "1111111111111111111111111111111111111111111111111111111111111111")
Assert(Format("abc{:b}def", 7), "abc111def")
Assert(Format("abc{:b}{:b}{:b}def", 4, 7, 4), "abc100111100def")

MsgBox("done")
return

Assert(vValue1, vValue2)
{
	if (vValue1 != vValue2)
		throw Error("Assert mismatch.`r`n" "Return value: " vValue1 "`r`n" "Expected value: " vValue2, -1)
}

;simplified version of function:
MsgBox(Text)
{
	MsgBox, % Text
}

Error(Message, Params*) ;Message, What, Extra
{
	local What
	if (Params.Length() > 2)
		throw Exception("Too many parameters passed to function.", -1)
	if !Params.HasKey(1)
		Params[1] := 0
	What := Params[1]
	if What is integer
		Params[1]--
	return Exception(Message, Params*)
}

```

### DIFF
--- a/source/script2.cpp
+++ b/source/script2.cpp
@@ -15539,19 +15539,41 @@ BIF_DECL(BIF_Format)
 					spec[spec_len++] = '6';
 					spec[spec_len++] = '4';
 					// Integer value; apply I64 prefix to avoid truncation.
-					value.value_int64 = ParamIndexToInt64(param);
+					if ((*aParam[param]).symbol == SYM_STRING
+					&& (*aParam[param]).marker[1] == 'b' && (*aParam[param]).marker[0] == '0')
+						value.value_int64 = _tcstoui64((*aParam[param]).marker+2, NULL, 2);
+					else
+						value.value_int64 = ParamIndexToInt64(param);
 					spec[spec_len++] = *cp++;
 				}
 				else if (_tcschr(_T("eEfgGaA"), *cp))
 				{
-					value.value_double = ParamIndexToDouble(param);
+					if ((*aParam[param]).symbol == SYM_STRING
+					&& (*aParam[param]).marker[1] == 'b' && (*aParam[param]).marker[0] == '0')
+						value.value_double = (double)_tcstoui64((*aParam[param]).marker+2, NULL, 2);
+					else
+						value.value_double = ParamIndexToDouble(param);
 					spec[spec_len++] = *cp++;
 				}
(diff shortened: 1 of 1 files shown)


jeeswg:
This fact:
#284, could only handle `{:b}`. E.g. it treated `{:04b}` as though it were `{:b}`.
It meant that this PR was too unintuitive, I have closed it for that reason. See #292, for an alternative.

PR #276, opened 2022-05-05
v2: Changed SplitPath to return an object when no output parameters are passed.

jeeswg:
## Introduction

[Thoughts for v2.0](https://www.autohotkey.com/v2/v2-thoughts.htm)
> SplitPath could return an object when no output parameters are passed. Property names are probably easier to remember than parameter order, and easier to read and write than counting commas. This should have no negative impact on compatibility if added later, presuming scripts do not rely on SplitPath returning "".

This PR is an attempt at implementing this functionality.

The property names match those listed here:
[AutoHotkey v3.0-alpha.1 - Switching to JavaScript (not really) - AutoHotkey Community](https://autohotkey.com/boards/viewtopic.php?f=24&t=88821)
`Name, Dir, Ext, NameNoExt, Drive`.
My PR also includes a `Path` property, everything considered, I thought it worth including.

## An aside

I have submitted all the PRs I intend to, apart from these, to come in the next 7 days:
BinCompare/BinCopy.
InBuf. Find nth/nth-to-last occurrence.
BinDiff/StrDiff. 2 buffers/strings. Find nth/nth-to-last difference (byte/char).

# Test code

```
;==================================================

;test code: SplitPath: return object (AHK v2)

MsgBox()

;==================================================

vName := "abc"

SplitPath([])

SplitPath([], &vName)

MsgBox(vName)

SplitPath(0.123)

SplitPath(0.123, &vName)

MsgBox()

;==================================================

;test SplitPath to object:

;vPath := A_AhkPath
vPath := "C:\Program Files\AutoHotkey\AutoHotkey.exe"
oPath := SplitPath(vPath)

vOutput := ""
for vProp, vValue in oPath.OwnProps()
	vOutput .= vProp ": " vValue "`r`n`r`n"
A_Clipboard := vOutput
MsgBox(vOutput)

;==================================================

;test SplitPath still works with output variables:

SplitPath(vPath, &vName, &vDir, &vExt, &vNameNoExt, &vDrive)

vList := "Path,Name,Dir,Ext,NameNoExt,Drive"

vOutput := ""
for vKey, vValue in StrSplit(vList, ",")
	vOutput .= vValue ": " v%vValue% "`r`n`r`n"
A_Clipboard := vOutput
MsgBox(vOutput)
return

;==================================================

/*
e.g. output:

Dir: C:\Program Files\AutoHotkey

Drive: C:

Ext: exe

Name: AutoHotkey.exe

NameNoExt: AutoHotkey

[...]

PR #289, opened 2022-07-09
v2: Changed SplitPath to return an object when no output parameters are passed. (New.)

jeeswg:
Here's an improved version of #276.
It adds an extra parameter to the internal `SplitPath` function, and if not null, puts the strings into it.

It's possible there is a macro I missed for the string assignments, that could have simplified them.

I've used `new` without explicitly clearing it up.
I did use `delete` but then removed it, because the function return values were being lost.
So perhaps I missed something here.

Some test strings I used:
```
vPath := "C:\Program Files\AutoHotkey\AutoHotkey.exe"
vPath := "C:\Program Files\AutoHotkey\AutoHotkey"
vPath := "AutoHotkey.exe"
vPath := "https://www.youtube.com/watch?v=ABCDEFGHIJK"
vPath := "C:"
```

Hopefully this PR has some merit, otherwise, best of luck implementing an alternative!

### DIFF
--- a/source/script2.cpp
+++ b/source/script2.cpp
@@ -6065,7 +6065,7 @@ throw_invalid_delimiter:
 
 
 
-ResultType SplitPath(LPCTSTR aFileSpec, Var *output_var_name, Var *output_var_dir, Var *output_var_ext, Var *output_var_name_no_ext, Var *output_var_drive)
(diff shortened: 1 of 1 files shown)


Lexikos:
> I've used new without explicitly clearing it up.

Please stop submitting pull requests until you have learned why that is unacceptable, *at the earliest*.

jeeswg:
I updated the branch with a 2nd commit, which isn't showing (possibly because the PR is closed).

The commit works well, and doesn't make use of `Var` objects.
So it's here if it suits you, or no problem if you have a different coding strategy in mind.

PR #291, opened 2022-07-09
v2: Added support for structes that passed by value

thqby:
Currently, the supported invocation methods are as follows.
```
DllCall(func, structsize := 16, structptr, retsize := 32)
DllCall(func, 'struct16', structptr, 'cdecl struct32')
DllCall(func, '16', structptr, 'cdecl 32')
ComCall(index, thisptr, structsize := 16, structptr, retsize := 32)
...
```

### DIFF
--- a/source/script.h
+++ b/source/script.h
@@ -598,6 +598,7 @@ enum DllArgTypes {
 	, DLL_ARG_FLOAT
 	, DLL_ARG_DOUBLE
 	, DLL_ARG_WSTR
+	, DLL_ARG_STRUCT
 	, DLL_ARG_STR  = UorA(DLL_ARG_WSTR, DLL_ARG_ASTR)
 	, DLL_ARG_xSTR = UorA(DLL_ARG_ASTR, DLL_ARG_WSTR) // To simplify some sections.
 };  // Some sections might rely on DLL_ARG_INVALID being 0.
--- a/source/script2.cpp
+++ b/source/script2.cpp
@@ -10640,9 +10640,9 @@ static inline UINT_PTR DynaParamToElement(DYNAPARM& parm)
 
 #ifdef WIN32_PLATFORM
 DYNARESULT DynaCall(int aFlags, void *aFunction, DYNAPARM aParam[], int aParamCount, DWORD &aException
-	, void *aRet, int aRetSize, DWORD aExtraStackSize = 0)
+	, void*& aRet, int aRetSize, DWORD aExtraStackSize = 0)
 #elif defined(_WIN64)
-DYNARESULT DynaCall(void *aFunction, DYNAPARM aParam[], int aParamCount, DWORD &aException, void* aRet = nullptr)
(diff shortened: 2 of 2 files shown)


HelgeffegleH:
You shouldn't have to pass the struct ptr/size, it should suffice if the _struct object_ has ptr/size properties. (I didn't look at your code, just your _documentation_.)

It might be better to defer this until the script can define structs, i.e., something like this,
```autohotkey
struct s { 
    int a
    int b
}
```

Anyways, thanks for your contribution 👍

thqby:
Sometimes the struct may be part of the buffer. In this case, it is more convenient to pass it directly to the ptr of struct. Of course, it is also possible to pass objects with ptr property.

Lexikos:
When passing structs by value to a C function, the *compiler* must know the exact size of the struct. It cannot vary at runtime. Future optimizations to DllCall could take advantage of this, but (I suppose) only if the struct size can be determined by statically analyzing the code.

The implementation of struct classes would supersede the `structXX` syntax, but only as a way of specifying the type and size. The underlying implementation would be very similar.

The inconsistency between `struct64` and `int64` seems odd, yet it wouldn't make sense to specify the number of *bits* for a struct. I more or less decided to also implement `int8`, `int16`, etc. but procrastinated because I was still considering replacing DllCall with something of a "define once, use many times" kind. (I finished considering; now I'm planning.)

Lexikos:
What was the reason for setting the DC_BORLAND flag? Isn't it redundant when aRetSize >= 8? Are you expecting aRetSize to sometimes be 3, 5, 6 or 7? Is it actually correct to pass such a value (supposing such a value is legal) the same way as for >= 8 (because I think that's what will happen with your code)?

thqby:
It's redundant when aRetSize > 8. But it is also passing the pointer when aRetSize is 3, 5, 6 or 7.

Lexikos:
I checked disassembly of some functions returning structs of 3, 5, 6 or 7 bytes, and confirmed that they are returned via the hidden parameter, which answers my previous questions about that. I assume the original DynaCall code merely failed to account for that (it isn't clearly documented for x86, anywhere I could find), and should be checking aRetSize for values 1, 2, 4 and 8, not merely <= 8.

I don't see any reason to add the DC_BORLAND flag when you could instead remove code that isn't needed. This isn't about compatibility with Borland compilers. On the other hand, it would probably be better to have this detail handled within DynaCall, not BIF_DllCall. In other words, remove the size checks from BIF_DllCall and fix the aRetSize check in DynaCall.

If you wish, you can amend the pull request by pushing to your `Lexikos-alpha` branch, either replacing the previous commit with `push -f` or adding a new commit.

I recently "ported" a version of PerformDynaCall to x86, and realized that the existing 32-bit parts of DynaCall are somewhat over-complicated. As I am thinking of replacing DllCall and adding struct support (though not in the short term), I may choose not to merge this.

PR #294, opened 2022-08-09
v2: Fixed hotstring has not been ltrimmed

thqby:
```
x::  msgbox()   ; ok
:x:xx::  msgbox()   ; error
```

### DIFF
--- a/source/script.cpp
+++ b/source/script.cpp
@@ -2119,7 +2119,7 @@ process_completed_line:
 			if (!hotstring_start)
 			{
 				// Mustn't use ltrim(hotkey_flag) because that would cause buf.length to become incorrect:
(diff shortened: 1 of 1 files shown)


Lexikos:
Thanks for making the effort to fix a bug. I think I either forgot or didn't notice this PR, so I fixed it with a41a1e1.

Some notes:
  - There was no need for the existing trimming after OTB `{` because every line is r-trimmed.
  - I renamed the `otb_brace` variable to something more appropriate for its new purpose.
  - I opted to keep the trimming all in one place. I just noticed we both left a stray comment that no longer fully applies (and makes more sense if it's in the vicinity of the trimming code, so I'm moving it): `// else don't trim hotstrings`.

PR #300, opened 2022-09-02
Added forward-slash support for FileCreateDir.

Ragnar-F:
https://www.autohotkey.com/boards/viewtopic.php?f=86&t=107687

### DIFF
--- a/source/script2.cpp
+++ b/source/script2.cpp
@@ -9818,23 +9818,28 @@ bool Line::FileCreateDir(LPTSTR aDirSpec, LPTSTR aCanModifyDirSpec)
 		return (attr & FILE_ATTRIBUTE_DIRECTORY) != 0; // Indicate success if it already exists as a dir.
 	}
 
+	// v1.1.35.00: Added forward-slash support for FileCreateDir:
+	TCHAR dirSpec[T_MAX_PATH];
+	if (!GetFullPathName(aDirSpec, _countof(dirSpec), dirSpec, NULL))
+		return false;
+
 	// If it has a backslash, make sure all its parent directories exist before we attempt
 	// to create this directory:
-	LPTSTR last_backslash = _tcsrchr(aDirSpec, '\\');
-	if (last_backslash > aDirSpec // v1.0.48.04: Changed "last_backslash" to "last_backslash > aDirSpec" so that an aDirSpec with a leading \ (but no other backslashes), such as \dir, is supported.
+	LPTSTR last_backslash = _tcsrchr(dirSpec, '\\');
+	if (last_backslash > dirSpec // v1.0.48.04: Changed "last_backslash" to "last_backslash > aDirSpec" so that an aDirSpec with a leading \ (but no other backslashes), such as \dir, is supported.
 		&& last_backslash[-1] != ':') // v1.1.31.00: Don't attempt FileCreateDir("C:") since that's equivalent to either "C:\" or the working directory (which already exists), or FileCreateDir("\\?\C:") since it always fails.
 	{
 		LPTSTR parent_dir;
 		if (aCanModifyDirSpec)
 		{
-			parent_dir = aDirSpec; // Caller provided a modifiable aDirSpec.
+			parent_dir = dirSpec; // Caller provided a modifiable aDirSpec.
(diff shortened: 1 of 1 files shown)


Lexikos:
Any modifications to aDirSpec must be reverted before returning, as it may point to the caller's variable (`ARG1` ← `sArgDeref[0]` ← `arg_deref[i]` ← `arg_var[i]->Contents()`).  It may instead point to literal text (`arg_deref[i] = this_arg.text`) which can be seen in ListLines and error messages, or the deref buffer, which is transient and probably okay to modify.

My local `alpha` branch has a simplified version of the function that makes a copy of the top-level input before beginning (mainly for correct usage of `const`). In that case modifications are safe. Alternatively, it would be fairly simple to scan for `'/'` or `'\\'` instead of just the latter (and copy the current character prior to temporary termination).

For a path like `A\B\..\C`, the current function would:
  - find `A\B\..\C` doesn't exist, recurse
  - find `A\B\..` (really `A`) doesn't exist, recurse
  - find `A\B` doesn't exist, recurse
  - find `A` doesn't exist, create it, return
  - create `A\B`
  - attempt to create `A\B\..` (really `A`)... fail, because it already exists.
  
That could be avoided by normalizing the path with GetFullPathName, which would also handle forward slashes.

Ragnar-F:
Thanks for the helpful explanation. I think GetFullPathName seems to make the most sense. I'm surprised it hasn't been used before. Probably it didn't exist at the time the command was created.

If something is not correct, feel free to let me know, I'll change it.

HelgeffegleH (changes requested):
Since the content of `dirSpec` is undefined if [`GetFullPathName`](https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getfullpathnamew) fails, it is not safe to proceed. I.e., you should check its return value and abort on failure.

From ms docs,

> If the function fails for any other reason, the return value is zero. To get extended error information, call GetLastError

Lexikos:
> I'm surprised it hasn't been used before. Probably it didn't exist at the time the command was created.

GetFullPathName is much older than AutoHotkey. FileCreateDir doesn't need a full path, so using it doesn't make immediate sense. The problems solved by normalizing the path aren't obvious, and the use of GetFullPathName for that purpose isn't obvious either. Some programmers write their own path normalization code...

> long-path support

Wat? If the requirements are met, long paths already work with FileCreateDir. Using GetFullPathName won't have any affect on this, as far as I am aware.

Ragnar-F:
> Wat? If the requirements are met, long paths already work with FileCreateDir. Using GetFullPathName won't have any affect on this, as far as I am aware.

You're right, I didn't give it enough thought (or testing). I've removed "long-path support".

Lexikos:
> If the lpBuffer buffer is too small to contain the path, the return value is the size, in TCHARs, of the buffer that is required to hold the path and the terminating null character.

I think the content of the buffer is also undefined in that case. It will happen if the path is too long. The appropriate response would be to SetLastError to a meaningful error number (for A_LastError) and return false.

However, this version calls GetFullPathName once for each parent directory, which is redundant since the top-level call has already resolved it to a full path. On the other hand, it does serve to copy the value into `dirSpec`, which makes the `!aCanModifyDirSpec` branch redundant because it's always safe to modify `dirSpec`.

I've already refactored FileCreateDir in v2 for other reasons (the requirement of a `const` parameter), and have added a GetFullPathName call, so I should just copy that into the v1 code. I was originally going to refactor it to use a v2-specific error-return mechanism to eliminate the SetLastError calls, but decided it wasn't worth the effort.

HelgeffegleH:
> I think the content of the buffer is also undefined in that case. It will happen if the path is too long.

I discarded this because,

> In the ANSI version of this function, the name is limited to MAX_PATH characters. To extend this limit to 32,767 wide characters, call the Unicode version of the function (GetFullPathNameW)

Since `dirSpec` has the size `T_MAX_PATH` characters which matches the stated limits above, it would be surprising if the function asked for more space than it supported (i.e., it should fail and return 0 if the limit is exceeded). However, no one should be surprised if they are surprised by MS, checking for it can certainly do no harm 👍.

PR #302, opened 2022-09-27
v2: Fixed WinAct crashing when aWinTitle was omitted.

thqby:


Lexikos:
I do not monitor GitHub notifications. The issue was reported on the forums and has been fixed by 2ec9dda (with an identical change).

PR #304, opened 2022-10-09
Fixed variable's object value being deleted before marking the variable as uninitialised

HelgeffegleH:
__Reason__, the variable might be read, directly or indirectly, under the false impression that the variable is set (with the object as its value), from the objects `__delete`, routine.

Example,
```autohotkey
x := {__delete : t=>msgbox(isset(x))}
x := unset
```

### DIFF
--- a/source/MdFunc.cpp
+++ b/source/MdFunc.cpp
@@ -458,8 +458,7 @@ bool MdFunc::Call(ResultToken &aResultToken, ExprTokenType *aParam[], int aParam
 		{
 			// Although 0 or "" is a fairly conventional default, it might not be safe.
 			// For error-detection and to avoid unexpected behaviour, "unset" the var.
-			var->Assign();
-			var->MarkUninitialized();
+			var->Uninitialize();
 		}
 	}
 
--- a/source/script_object.cpp
+++ b/source/script_object.cpp
@@ -2089,9 +2089,11 @@ ResultType Array::GetEnumItem(UINT &aIndex, Var *aVal, Var *aReserved, int aVarC
 			switch (item.symbol)
 			{
 			default:
-				aVal->AssignString(item.string, item.string.Length());
 				if (item.symbol == SYM_MISSING)
-					aVal->MarkUninitialized();
+					aVal->Uninitialize();
+				else
+					aVal->AssignString(item.string, item.string.Length());
+				
 				break;
 			case SYM_INTEGER:	aVal->Assign(item.n_int64);			break;
 			case SYM_FLOAT:		aVal->Assign(item.n_double);		break;
--- a/source/var.cpp
+++ b/source/var.cpp
@@ -43,8 +43,7 @@ ResultType Var::Assign(Var &aVar)
 
 	if (source_var.mAttrib & VAR_ATTRIB_UNINITIALIZED)
 	{
-		target_var.Assign();
-		target_var.MarkUninitialized();
+		target_var.Uninitialize();
 		return OK;
 	}
 
@@ -84,10 +83,7 @@ ResultType Var::Assign(ExprTokenType &aToken)
 	default:
 		ASSERT(!"Unhandled symbol");
 	case SYM_MISSING:
-		if (!Assign())
-			return FAIL;
-		MarkUninitialized();
-		return OK;
+		return Uninitialize() ? OK : FAIL;
 	}
 	// Since above didn't return, it can only be SYM_STRING.
 	return Assign(aToken.marker, aToken.marker_length);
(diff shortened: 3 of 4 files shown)


Lexikos:
There is a subtle error on this line:
```AutoHotkey
auto obj = (var.mAttrib & VAR_ATTRIB_IS_OBJECT) ? mObject : nullptr;
```
`mObject` should be `var.mObject`.

As an optimization, we can replace the AddRef call with `var.mAttrib &= ~VAR_ATTRIB_IS_OBJECT;`. This avoids the runtime cost of two virtual method calls, and reduces code size (with the difference being effectively multiplied by inlined copies of the code).

There are a couple of minor style issues with the code:
  - `obj -> AddRef();` (spaces) inconsistent with `obj->Release();`.
  - Spacing between functions (it's not entirely consistent across the whole project, but at least try to match the surrounding code).

---

It appears that `var.Free(VAR_NEVER_FREE, true)` would already handle things in the correct manner, applying `VAR_ATTRIB_UNINITIALIZED` and removing `VAR_ATTRIB_IS_OBJECT` before calling Release. Although this implementation of Uninitialize seems redundant, it does produce smaller code overall than delegating to Free, probably due to inlining.

For `var := unset`, we probably want `VAR_ALWAYS_FREE` or `VAR_FREE_IF_LARGE` rather than `VAR_NEVER_FREE`, which is what it was using (because the parameters of Assign were omitted). In v2.0-beta.12, `var := ""` is more likely to free the memory than `var := unset`, which I suppose is counter-intuitive. Other callers of Uninitialize might be better with `VAR_FREE_IF_LARGE` or as-is, favouring speed over memory usage.

`var->Free(..., true)` shouldn't be used directly for `var := unset`, because `var` can be an alias, in which case `true` would cause it to revert to a normal variable instead of "uninitializing" the target variable. Now that there is a need, the second parameter should probably be split into separate options (flags or parameters). Combining it with the first parameter as a set of flags might be best for code size.

> if (result)

What do you expect to happen when `Assign()` fails? Ordinarily, it might retain its previous value. (Actually, it was generally making the variable *blank*, not unset.) In this case, maybe we want it to be marked as uninitialized regardless; but I don't think it's possible for `Assign()` to fail when the parameters are omitted. It can only fail for memory allocation failures.

When it's not possible for a function to fail, I prefer to remove the return value to show that there can only be one outcome. Sometimes this also results in smaller code.

---

There is a similar case that Uninitialize doesn't cover:
```AutoHotkey
x := {__delete : t=>msgbox(x ?? "not set")}
x := "foo"
```
So this is really a more general problem that can't be solved by just adding a new method.

There's another similar case that AssignString didn't handle correctly: if __Delete assigned a new object to the variable, AssignString unconditionally removed `VAR_ATTRIB_IS_OBJECT` and continued with the assignment regardless. Any other type of value assigned by __Delete was also lost, but without leaking anything.

Actually, even `Assign()` followed by `MarkUninitialized()` can backfire, because Assign can cause __Delete to execute, which can assign a new value to the variable.

---

I think 67776fd covers everything.

HelgeffegleH:
> There are a couple of minor style issues with the code:

Sometimes VS makes some whitespace adjustments, for example, [this was not intended]( https://github.com/Lexikos/AutoHotkey_L/pull/304/files#diff-f475089a4408d208755992ee264c5a1f5d9fefd7551809f4cce4ed1a43a99ddaL929-R929). I accidently cut the code instead of copying it, then pasting it back again changed the whitespace, which I noticed after making the PR, intending to fix it later, but forgot. The `x -> y`  inconstiencty is probably just a helgeffeeghel volatile finger error though, sorry.

> What do you expect to happen when Assign() fails?

I just tried to preserve the return value as it would have been before the change, see previous lines [L87-L90](https://github.com/Lexikos/AutoHotkey_L/pull/304/files#diff-3244e8b1da6a43a8428fa75c32aa0013c79462a51f1aea3cc804c91c1d25781aL87-L90) and [L46-L48](https://github.com/Lexikos/AutoHotkey_L/pull/304/files#diff-3244e8b1da6a43a8428fa75c32aa0013c79462a51f1aea3cc804c91c1d25781aL46-L48)

> I don't think it's possible for Assign() to fail when the parameters are omitted. It can only fail for memory allocation failures.
> When it's not possible for a function to fail, I prefer to remove the return value to show that there can only be one outcome. Sometimes this also results in smaller code.

I think it is more maintainable to treat it as if it can fail though, which I'd prefer over hypothetical code size optimisations.  But as indicated above, I favoured not changing the previous ways the return value propagated.

Thank you very much for the feedback and the fixes :+1:

PR #307, opened 2022-11-27
v2: Change arg 3 of RegExReplace to support callback function

thqby:
`RegExReplace(Haystack, NeedleRegEx, m => expression, &OutputVarCount?, Limit := -1, StartingPosition := 1)`

Support the above forms of calls

### DIFF
--- a/source/lib/regex.cpp
+++ b/source/lib/regex.cpp
@@ -686,11 +686,24 @@ void RegExReplace(ResultToken &aResultToken, ExprTokenType *aParam[], int aParam
 	// Get the replacement text (if any) from the incoming parameters.  If it was omitted, treat it as "".
 	TCHAR repl_buf[MAX_NUMBER_SIZE];
 	LPTSTR replacement = _T("");
+	ResultToken result_token;
+	ExprTokenType matchobj_token, *params;
+	IObject *callback_obj = nullptr;
+	int prev_excpt = g->ExcptMode;
+	result_token.mem_to_free = nullptr;
 	if (!ParamIndexIsOmitted(2))
 	{
-		if (ParamIndexToObject(2))
-			_f_throw_param(2, _T("String"));
-		replacement = ParamIndexToOptionalString(2, repl_buf);
+		if (callback_obj = ParamIndexToObject(2))
+		{
+			if (!ValidateFunctor(callback_obj, 1, aResultToken))
+				return;
+			params = &matchobj_token;
+			matchobj_token.symbol = SYM_OBJECT;
+			result_token.InitResult(repl_buf);
+			g->ExcptMode |= EXCPTMODE_CATCH;
+		}
+		else
+			replacement = ParamIndexToOptionalString(2, repl_buf);
 	}
 
 	// In PCRE, lengths and such are confined to ints, so there's little reason for using unsigned for anything.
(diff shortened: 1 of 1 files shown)


thqby:
```ahk
; Test old code
MsgBox 'RegExReplace("abc123123", "123$", "xyz") = "abc123xyz"`n' (RegExReplace("abc123123", "123$", "xyz") = "abc123xyz")
MsgBox 'RegExReplace("abc123", "i)^ABC") = "123"`n' (RegExReplace("abc123", "i)^ABC") = "123")
MsgBox 'RegExReplace("abcXYZ123", "abc(.*)123", "aaa$1zzz") = "aaaXYZzzz"`n' (RegExReplace("abcXYZ123", "abc(.*)123", "aaa$1zzz") = "aaaXYZzzz")
MsgBox 'RegExReplace("abc123abc456", "abc\d+", "", &ReplacementCount) = ""`n' (RegExReplace("abc123abc456", "abc\d+", "", &ReplacementCount) = "") "`nReplacementCount = " ReplacementCount

; Test regular callback
MsgBox 'RegExReplace("123+6567=", "^(\d+)\+(\d+)=", m => m[1] + m[2]) = "6690"`n' (RegExReplace("123+6567=", "^(\d+)\+(\d+)=", m => m[1] + m[2]) = "6690")

esc:= Map("n", "`n", "r", "`r", "b", "`b", "t", "`t", "v", "`v", "a", "`a", "f", "`f")
MsgBox 'RegExReplace("1\n2\n3\g\\", "\\(.)", m => esc.Get(m[1], m[0])) = "1``n2``n3\g\\"`n' (RegExReplace("1\n2\n3\g\\", "\\(.)", m => esc.Get(m[1], m[0])) = '1`n2`n3\g\\')

i := 0
MsgBox 'RegExReplace("1 2 3 4 5 6 7 8 9", "\d", cb) = "1 4 9 4 5 6 7 8 9"`n' (RegExReplace("1 2 3 4 5 6 7 8 9", "\d", cb) = "1 4 9 4 5 6 7 8 9")
cb(m) {
	global i
	if ++i < 4
		return m[0] * m[0]
	Exit	; Abort regular replacement
}
```

Lexikos:
I do not think it is appropriate to overload the meaning of Exit in this way. RegExReplace does not (and should not) start a new thread.

*If* the ability to abort matching is needed, there are other ways that it could be implemented. For instance, throwing a number to indicate how/whether matching should proceed (see callout return values for example) or setting a property of the match object. In the latter case, the function could provide replacement text for the current match while also controlling how matching proceeds.

For indentation and braces, please aim to match the surrounding code.

thqby:
> For instance, throwing a number to indicate how/whether matching should proceed (see callout return values for example)

I chose the first one and pushed it again.

Lexikos:
I decided not to include the mechanism for affecting subsequent matches.

The normal case is to replace the current match and resume searching at the end of the match.

There are several ways that a callback could affect the replacement process:
  - Skip the current match and resume searching at the next character (like normal regex "bump-along").  
    `throw 1` permits this. The alternative would be to use an additional callout.
  - Skip the current match and resume searching at the end of the match.  
    `throw 0` permits this, but so does `return m.0`.
  - Skip the current match and all subsequent matches.  
    `throw -1` permits this, but the same effect can be achieved by returning `m.0` for all subsequent matches.
  - Skip the current match and resume searching at a specific position.
  - Replace the current match and stop searching. 
  - Replace the current match and resume searching at a specific position.

So it seems `throw -1|0|1` offers a small increase in flexibility, but nothing unique, and (as I mentioned before) there are some cases it doesn't cover that could be covered by some other mechanism.

I measure the current code size cost at 1396 bytes for x64 with VC++ 2022. That can be reduced by a mere 16 bytes by replacing the HasBase check with `TokenIsPureNumeric(*g->ThrownToken) == PURE_INTEGER`, which also avoids silently coercing non-Error values to integer.

If the mechanism for affecting subsequent matches is removed, the cost is only 772 bytes. For adding callback functionality to RegExReplace, this is attractive. By contrast, nearly doubling the cost to allow a small amount of redundant flexibility is not attractive.

There are some other minor reasons:
  - The additional mechanism means more documentation.
  - The basic usage is intuitive, but `throw -1|0|1` isn't, and would more often require referring to the documentation. It might be easier for a user to achieve the same effect using another method, from memory.
  - There is no precedent for "overloading" exception handling for use as a secondary return mechanism in a built-in library.
  - It appears that JavaScript, Python, Ruby, PHP and Go all have some equivalent to a RegExReplace callback, and none of them support this additional mechanism.

If anyone needs to control how the search-and-replace loop behaves, it isn't difficult to write a RegExMatch loop instead. Perhaps in future there will be an enumerator or similar to simplify that further.

PR #280, opened 2022-05-15
v2: Added BinCompare/BinCopy/InBin/BinDiff/StrDiff functions.

jeeswg:
## Introduction

4 possible binary/buffer-related functions, now that we have a built-in `Buffer` object.
And a possible string-related function.

Occurrence parameters allow searching/comparing from left-to-right and right-to-left.
To find the nth or nth-to-last occurrence.

## Implementation: offset parameter pairs

I'm generally happy with the functionality.
The one possible concern might be that having offset parameters pairs, can be a bit verbose.

A possible solution, would be to remove those offset pairs in 4 of the 5 functions, and add a method to Buffer objects, that returns either a buffer, or a basic AHK object with `Ptr`/`Size` properties. Something like one of these:

```
oBuf := Buffer(100, 0)

;equivalent to {Ptr:oBuf.Ptr+20, Size:oBuf.Size-20}:
oBuf2 := oBuf.Offset(20)
oBuf2 := oBuf.Slice(20)
oBuf2 := oBuf.Sub(20) ;cf. SubStr
oBuf2 := oBuf[20]

;equivalent to {Ptr:oBuf.Ptr+oBuf.Size-20, Size:20}:
oBuf2 := oBuf.Right(20)
oBuf2 := oBuf[-20]
```

## Documentation

```
BinCompare(Buffer1, Buffer2 [, Size, Offset1, Offset2])
BinCopy(SourceBuffer, DestBuffer [, Size, Offset1, Offset2])
InBin(HaystackBuffer, NeedleBuffer [, Step:=1, StartOffset, Occurrence])
BinDiff(Buffer1, Buffer2 [, Step:=1, Size, Offset1, Offset2, Occurrence])

StrDiff(Str1, Str2 [, CaseSense:=0, Length, Offset1, Offset2, Occurrence])
```

## BinCompare
Params 1 and 2 can be addresses.
Offset1 and Offset2 are 0-based.
If Size is omitted, params 1 and 2 must be buffers, and the buffer sizes minus respective offsets must be equal.

The return value:
(Note: taking any size/offset stipulations into account:)
0, if Buffer1 is identical to Buffer2
a positive integer, if Buffer1 is greater than Buffer2
a negative integer, if Buffer1 is less than Buffer2

## BinCopy
Params 1 and 2 can be addresses.
Note: the parameter order is source then destination (like AutoHotkey's `FileXXX` and `DirXXX` functions, but unlike `memmove`).
Offset1 and Offset2 are 0-based.
[...]

PR #308, opened 2022-12-11
v2: Added BinCopy() and BinCompare().

jeeswg:
## Introduction

Replaces #280. Simplified versions of 2 of the 5 functions.
2 small functions, wrappers for `memmove` and `memcmp`.

```
BinCompare(Buffer1, Buffer2 [, Size])
BinCopy(SourceBuffer, DestBuffer [, Size])
```
`Buffer1` and `Buffer2` can be integers.
`SourceBuffer` and `DestBuffer` can be integers.
If exactly one is an integer, `Size` can be omitted.
If both are buffers of equal size, `Size` can be omitted.
The `BinCopy` parameter order differs from `memmove`.

## Rationale

It seemed pretty practical to have these built-in, because they're used a lot.
But no problem if you don't want them.

## Implementation

Anything can be changed as desired, e.g. the function names, internal variable names, all the various details for handling integers versus buffers.

## `ST_UNICODE` undefined and `bin_debug`
Visual Studio 2017 complained that `ST_UNICODE` (value 8, in error.cpp) was undefined.
It might be helpful for people editing the source code, to add a directive to define it if it's undefined.

It might also be helpful to add `bin_debug` to .gitignore.

### DIFF
--- a/source/script.cpp
+++ b/source/script.cpp
@@ -36,6 +36,8 @@ FuncEntry g_BIF[] =
 	BIFn(ACos, 1, 1, BIF_ASinACos),
 	BIFn(ASin, 1, 1, BIF_ASinACos),
 	BIF1(ATan, 1, 1),
+	BIFn(BinCompare, 2, 3, BIF_BinCopyCompare),
+	BIFn(BinCopy, 2, 3, BIF_BinCopyCompare),
 	BIF1(CaretGetPos, 0, 2, {1, 2}),
 	BIFn(Ceil, 1, 1, BIF_FloorCeil),
 	BIF1(Chr, 1, 1),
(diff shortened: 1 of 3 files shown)


Lexikos:
I am not a fan of the function names and am unsure about the usage. I think that if I decide these simple functions should be implemented, I will decide on names and usage, and implement them myself.

> Visual Studio 2017 
I will not test with any compilers other than the latest I have installed, and I will not install 2017.

> ST_UNICODE (value 8, in error.cpp) was undefined.
Your installed Windows SDK may be too old.

jeeswg:
That's great, thank you.

Btw `ST_UNICODE` (used by the new error dialog RichEdit control) is the only time AutoHotkey v2.0 has failed to compile (for me).
AutoHotkey v1.1 has never failed.

PR #309, opened 2022-12-11
v2: Added ShortcutKey parameter to FileGetShortcut.

jeeswg:
Replaces #271.
This version of the PR, adds a `ShortcutKey` parameter as the **last** parameter.
Therefore it is backwards compatible with the current function.

(It has been nice to work with the new MdFunc set-up for the first time!)

### DIFF
--- a/source/MdFunc.cpp
+++ b/source/MdFunc.cpp
@@ -12,7 +12,7 @@ struct MdFuncEntry
 	LPCTSTR name;
 	void *function;
 	MdType rettype;
-	MdType argtype[23];
+	MdType argtype[25];
 };
 
 
--- a/source/lib/functions.h
+++ b/source/lib/functions.h
@@ -106,7 +106,7 @@ md_func_v(FileExist, (In, String, Pattern), (Ret, String, RetVal))
 md_func(FileGetAttrib, (In_Opt, String, Path), (Ret, String, RetVal))
 md_func(FileGetShortcut, (In, String, LinkFile), (Out_Opt, String, Target), (Out_Opt, String, WorkingDir),
 	(Out_Opt, String, Args), (Out_Opt, String, Description), (Out_Opt, String, IconFile),
-	(Out_Opt, Variant, IconNum), (Out_Opt, Int32, RunState))
+	(Out_Opt, Variant, IconNum), (Out_Opt, Int32, RunState), (Out_Opt, String, ShortcutKey))
 md_func(FileGetSize, (In_Opt, String, Path), (In_Opt, String, Units), (Ret, Int64, RetVal))
 md_func(FileGetTime, (In_Opt, String, Path), (In_Opt, String, WhichTime), (Ret, String, RetVal))
 md_func(FileGetVersion, (In_Opt, String, Path), (Ret, String, RetVal))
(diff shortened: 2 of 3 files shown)


Lexikos:
Increasing the size of `MdFuncEntry::argtype` from 23 to 25 bytes has the effect of increasing code size by about 2KB...

Changing it to use `std::initializer_list` increases code size by 1208 bytes instead. This effectively separates the argtype arrays in the data, requiring each entry to have a pointer instead of having data directly. It appears that 23 bytes is the limit for a fixed-size array to be the most efficient option. (Increasing the array to 24 bytes increases the overall size of the struct by more than 1 byte due to alignment requirements.)

I will likely further defer any changes to FileGetShortcut/FileCreateShortcut.

> FileGetShortcut and FileCreateShortcut have too many parameters, but I haven't decided whether to change them or just replace both with a function to get a ShellLinkObject (through COM), either built-in or as a trivial script. ShellLinkObject has all the same capabilities plus a Hotkey property and maybe others.
> [v2-thoughts](https://web.archive.org/web/20220528174056/https://www.autohotkey.com/v2/v2-thoughts.htm)

jeeswg:
Thanks very much for the important details.

Since I'm a little unclear on what you're saying, I will leave this PR unchanged unless you follow up with further comments.
Of course, you're welcome to add any optimisations and/or modifications.

From my perspective, `FileCreateShortcut`/`FileGetShortcut` aren't used that much.
I don't desire any changes beyond what I've proposed.
If they were changed to use object syntax, I wouldn't oppose that, but I'm happy with what we have now.

Lexikos:
What I'm saying is that the changes needed to support a function having this many parameters presently come with a measurable cost. This cost in combination with the inherent issues of having that many parameters outweighs the value of this addition, in my opinion.

I personally think that there is little to no value in setting shortcut keys in .LNK files, due to [issues described elsewhere](https://superuser.com/questions/426947/slow-windows-desktop-keyboard-shortcuts). If someone specifically needs a way to detect LNK files that have shortcuts, such as to remove them, it can be done trivially with the ShellLinkObject COM object.

> From my perspective, FileCreateShortcut/FileGetShortcut aren't used that much.

Setting or retrieving a shortcut key would be used even less, so why are we wasting time on this?

jeeswg:
There were certain elements of 'cleaning up' AutoHotkey, and adding consistency, that I deemed sufficiently worthwhile to write C++ code for.
Since no other functions have this many output parameters, this PR needed to increase that limit, but this seems to have a surprising 'measurable cost' as you put it, that I didn't foresee. So I'll close this PR.

My main interest in lnk file shortcut keys was to identify any lnk files that had them, and to then remove those shortcut keys from the lnk files. IIRC, an AutoHotkey hotkey was failing, and it was because of a WinMerge lnk file. Being new to AutoHotkey, I had to use NirSoft ShortcutsMan to get shortcut key information.
(IIRC, there was a surprising amount of demand for *setting* lnk shortcut keys with specific modifier keys. I had considered using such functionality, for temporary key blocking, and permanent/temporary file opening situations.)

I can remember how difficult things could be as an AHK newbie, and would definitely recommend custom AHK code be added to the documentation in place of this PR.

PR #310, opened 2022-12-11
v2: Changed FileCreateShortcut's ShortcutKey parameter to be more permissive.

jeeswg:
Replaces #277.
This PR, like the original, allows any combination of modifier keys.
But for this version of the PR, if the modifier keys are omitted, it assumes `Ctrl+Alt` (instead of throwing).
Therefore it is backwards compatible with the current function.

### DIFF
--- a/source/script_autoit.cpp
+++ b/source/script_autoit.cpp
@@ -1264,13 +1264,31 @@ bif_impl FResult FileCreateShortcut(StrArg aTargetFile, StrArg aShortcutFile, op
 			psl->SetIconLocation(aIconFile.value(), icon_index - (icon_index > 0 ? 1 : 0)); // Convert 1-based index to 0-based, but leave negative resource IDs as-is.
 		if (aHotkey.has_value())
 		{
+			WORD mods = 0;
+			LPCTSTR cp = aHotkey.value();
+			for (;; ++cp)
+			{
+				if (!*cp || !cp[1]) // The last char is never a modifier.
+					break;
+				else if (*cp == '^')
+					mods |= HOTKEYF_CONTROL;
+				else if (*cp == '!')
+					mods |= HOTKEYF_ALT;
+				else if (*cp == '+')
+					mods |= HOTKEYF_SHIFT;
+				else
+					break;
+			}
+
+			// For backwards compatibility: if modifiers omitted, assume CTRL+ALT.
+			if (!mods)
+				mods = HOTKEYF_CONTROL | HOTKEYF_ALT;
+
 			// If badly formatted, it's not a critical error, just continue.
-			// Currently, only shortcuts with a CTRL+ALT are supported.
-			// AutoIt3 note: Make sure that CTRL+ALT is selected (otherwise invalid)
-			vk_type vk = TextToVK(aHotkey.value());
+			vk_type vk = TextToVK(cp);
 			if (vk)
 				// Vk in low 8 bits, mods in high 8:
-				psl->SetHotkey(   (WORD)vk | ((WORD)(HOTKEYF_CONTROL | HOTKEYF_ALT) << 8)   );
+				psl->SetHotkey( (WORD)vk | (mods << 8) );
 		}
 		if (aRunState.has_value())
 			psl->SetShowCmd(*aRunState); // No validation is done since there's a chance other numbers might be valid now or in the future.


Lexikos:
What gave you the idea to make `#` set the `HOTKEYF_EXT` flag? It has nothing to do with the Win key. As per the Hotkey control, it should be set based on the key (sc1xx vs sc0xx), not the modifiers.

>  // Shell Link (.LNK) standard says this should never be set.

Where? All I can find is:

> **SetHotkey**
> The modifier flags can be a combination of the values specified in the description of the IShellLink::GetHotkey method.

> **GetHotkey**
> The modifier flags can be a combination of the following values. ...
> HOTKEYF_EXT
> Extended key

However, it has no effect on which keys activate the shortcut, only the displayed text in the Properties dialog.

Lexikos:
`if (!cp[1])` will be indeterminate if the caller passes an empty string for `aHotkey`.

jeeswg:
## `HOTKEYF_EXT`
The standard is here:
[[MS-SHLLINK]: Shell Link (.LNK) Binary File Format | Microsoft Learn](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-shllink/16cb4ca1-9339-4d0c-a68d-bf1d6cc0f943)
'HOTKEYF_' appears 3 times in one table.

> This value MUST be one or a combination of the following:
[0x00 / HOTKEYF_SHIFT / HOTKEYF_CONTROL / HOTKEYF_ALT.]

The rationale behind using `#` (in this PR and #309) was so that there was some way to read/write the value of `HOTKEYF_EXT`.
Either we remove the ability to read/write it, or we choose a different character to represent it.
The documentation would explain the details to the user.
It's possible that someone else set this `#` precedent elsewhere, not myself.

## `if (aHotkey.has_value())`
So `if (aHotkey.has_value())` is similar to `if (!ParamIndexIsOmitted(i))`?
Is there any major difference?
(IIRC, before the MdFunc interface was introduced, this line was just an 'is string blank' check.)

I have force-pushed this update. Thanks!
```
before/after:
if (!cp[1])
if (!*cp || !cp[1])

```

## Notifications

Clipboard append bug in AHK v1, not in v2. It has existed since AHK Basic. #235.
`ControlExist` should never throw (except for invalid input) rationale. #244.

Lexikos:
We are not writing the file. The standard for the binary format of the file does not define what we can or cannot do with the API.

> Either we remove the ability to read/write it, or we choose a different character to represent it.

As I said, it should be set based on the key (sc1xx vs sc0xx), not the modifiers - if it is set at all.

> The rationale behind using #

This is one way to provide a capability, but you haven't given any rationale for providing the capability, nor for choosing this particular syntax over other possibilities. I assume this means that there wasn't any particular logic to it, which is fine because there is no way that I would have implemented it anyway.

> It's possible that someone else set this # precedent elsewhere, not myself.

What does this even mean? *Maybe* someone, *somewhere* has used `#` for this purpose, and that might make it more acceptable? It sounds like an unlikely excuse, not an answer to my (now mostly rhetorical) question.

> So `if (aHotkey.has_value())` is similar to `if (!ParamIndexIsOmitted(i))`?

It is directly equivalent. Optional string parameters (`optl<StrArg>`) also have `has_nonempty_value` and several other methods for readability.

TextToVK currently handles the empty string case.

> Notifications

FYI, I make it a point to ignore these remarks that have nothing to do with the PR in which they are noted.

jeeswg:
> We are not writing the file.

I'm unsure what you mean. (We are creating or overwriting a lnk file.)

> The standard for the binary format of the file does not define what we can or cannot do with the API.

I agree, although generally speaking we should avoid doing anything *undocumented*. So I'd recommend that source code and/or documentation comments should indicate that...

There seem to be inconsistencies in the Microsoft documentation.
`HOTKEYF_EXT` is listed as valid for `GetHotkey` and `SetHotkey` that read/write from/to lnk files, and yet the lnk standard says that `HOTKEYF_EXT` should never be set in lnk files.
This PR (and #309) would allow it to be gotten/set.

> The rationale behind using #

One reason to allow getting/setting `HOTKEYF_EXT` is to get and then set a lnk file's attributes: modifying some, but preserving the rest, leaving them unchanged. The shortcut key might contain `HOTKEYF_EXT` *for whatever reason*.

`+^!` set the Shift/Ctrl/Alt modifiers, consistent with AutoHotkey hotkeys.
IIRC, someone suggested that setting the `HOTKEYF_EXT` flag would create a WinKey hotkey, hence `#` would be the natural choice, since AutoHotkey hotkeys already use it.

I would be happy to change the character, and believe choosing *some* character is useful, for preserving hotkeys as mentioned above, and for simplicity/consistency with `+^!`.
However, a new character could confuse the Community, since `+^!` could be used in hotkeys and elsewhere, but not the new character. Whereas using `#`, an established character, but for a slightly different purpose, would cause less confusion.

(If not by a character, how else to input/output `HOTKEYF_EXT`.)

> As I said, it should be set based on the key (sc1xx vs sc0xx), not the modifiers - if it is set at all.

This PR and #309 read/write `HOTKEYF_EXT` just like any other modifier.
Are you suggesting they should also read from the `vk` value, and convert the VK (virtual key) to an SC (scan code)?
I.e. the key that isn't the modifier? E.g. `^!a`, the `a` would also determine the presence of `HOTKEYF_EXT`?

> Notifications

Please consider adding notification and code style guidelines/links to the readme.
At present, I see no standardised public way to provide important notifications for the `AutoHotkey/AutoHotkey` repo.
(The readme doesn't state this, but IIRC, GitHub notifications, and Issues, are both disabled for the `AutoHotkey/AutoHotkey` repo.)

jeeswg:
Would removing these 2 lines, give an acceptable PR? Or are there other concerns?
```
				else if (*cp == '#')
					mods |= HOTKEYF_EXT; // Shell Link (.LNK) standard says this should never be set.
```

If you *had* to have a mechanism that would allow you to get/set `HOTKEYF_EXT`, what would you propose? Would `#` or a different character suffice?

If you could, please expand on any `SC` comments.

jeeswg:
I've force-pushed all of the updates that you specified.

## Scan codes and virtual keys

I found it difficult to find good information on scan codes.

This information that you provided is very valuable and obscure, and would be useful to add to the source code and/or documentation:
- `HOTKEYF_EXT` has no effect on which keys activate the shortcut file, only the displayed text in the shortcut file's Properties dialog.
- As per the Hotkey control, `HOTKEYF_EXT` should be set based on the friendly key name (by consulting a lookup table), not the modifiers.

I know the relevant source code predates you, but ideally the code would have already handled or at least mentioned `HOTKEYF_EXT`. For example, at present, if you specify `^!Down`, the file Properties dialog displays `^!NumpadDown`. I didn't see any hint of this in the source code or documentation.

`GetKeyName("scxxx")` (where the x's are hex digits) gives very different results in AHK v1 and AHK v2.
It also gives key names for some values sc2xx/sc3xx and above in both versions.
Can sc2xx/sc3xx and above ever be valid scan codes?

I am working on a tutorial containing valuable/obscure Winapi information. Hotkeys are the one key area I haven't looked into deeply. In case anyone's interested in doing a write-up.

PR #292, opened 2022-07-16
v1: Added binary integer support to Format. (New.)

jeeswg:
Replaces #284. (See that PR for test code.)

`String := Format(FormatStr , Values...)`

`FormatStr`:
This PR can handle `{:b}` *and* variants such as `{:04b}`.
#284, could only handle `{:b}`. E.g. it treated `{:04b}` as though it were `{:b}`.
The new code is simpler and more powerful.

`Values...`:
Like #284, this PR handles strings that begin with '0b', for the `diouxX/eEfgGaA/cCp/b` options.
For AHK v2: this part of the code would be unnecessary if #285 were implemented.

In both cases, the code handles `b`, but not `B`. This can be changed if desired, to match `o` (octal) or `x/X` (hex).

The `Format` code is very fiddly, so it's possible I overlooked something.

PR #269, opened 2022-05-05
v2: Added DTPGetTime/DTPSetTime/MonthCalGetTime/MonthCalSetTime functions.

jeeswg:
## Introduction: RemoteBuffer

An acceptable alternative to these 4 functions would be to implement a `RemoteBuffer` class (e.g. #268).
To make it easy to write simple/understandable custom calendar/date picker (and other) functions.

`RemoteBuffer` would make it possible to get/set the text and get icon indexes etc from multiple control types succinctly, e.g. simplified versions of these functions: [GUIs via DllCall: text functions (get/set internal/external control text) - AutoHotkey Community](https://autohotkey.com/boards/viewtopic.php?f=6&t=40514).

Further examples:
- [Manipulating TreeView Control in AHKv2 - AutoHotkey Community](https://autohotkey.com/boards/viewtopic.php?f=82&t=110749)
- [Detect icons from a ComboBox controls - AutoHotkey Community](https://autohotkey.com/boards/viewtopic.php?style=2&f=5&t=83859) [ComboBoxEx controls]

## Introduction

[Thoughts for v2.0](https://www.autohotkey.com/v2/v2-thoughts.htm)
> We seem to lack functions for getting/setting the value of external controls of some types, such as Slider, UpDown, MonthCal (I have not tested ControlSetText on these).

A summary of GUI functionality:
- I have written code, here, for SysDateTimePick32 and SysMonthCal32 controls. Note: for some control types such as SysDateTimePick32 you may need to 'jog' the control for the change to be recognised, e.g. by sending an arrow key press. (But in NirSoft SearchMyFiles for example, you do not need to 'jog' the controls.)
- In a separate PR (#270), I have written code for SysHeader32 (listview header) controls, to complement `ListViewGetContent()`.
- Slider (msctls_trackbar32) controls can be handled by SendMessage one-liners (TBM_GETPOS/TBM_SETPOSNOTIFY).
- UpDown (msctls_updown32) controls can be handled by SendMessage one-liners (UDM_GETPOS32/UDM_SETPOS32, or UDM_GETBUDDY to get a handle to the Edit control, then ControlGetText/ControlSetText).
- In a separate PR, I have written RemoteBuffer code (#268), which can be used for all other controls, e.g. Scintilla controls.

## Implementation

The function names and any other aspects can be changed as desired.

The functions were based on existing DTP/MonthCal functionality for internal GUIs.
Perhaps if that functionality ever changes, then these functions should be changed to match.

## Documentation

```
[...]

PR #311, opened 2022-12-11
v1: Added DecToBase() and BaseToDec().

jeeswg:
## Introduction

An alternative to #292. This function is more versatile, and simpler to implement.

```
BaseToDec(Str, Base) [a wrapper for _wcstoui64/_strtoui64] [similar to JavaScript's parseInt(str, base)]
DecToBase(Num, Base, MinLen:=1) [a wrapper for _ui64tow/_ui64toa] [similar to JavaScript's num.toString(base)]
```

`MinLen` can be used to pad the output with leading zeros.

`Base` must be between 2 and 36 inclusive, else it throws.
`MinLen` must be positive or 0, else it throws.

## Implementation

The output is currently lower case, this can be changed if desired.
(I'm very 50/50 on this, I tend to lean towards upper case for bases, hex, RGB values and CRC-32 hashes, but I've often seen hashes written as lower case, and RGB values fairly often.)

## Rationale

These can be surprisingly useful in a variety of circumstances, including for newbies, hence why I leaned towards them being built-in.
E.g. for generating lists of combinations/permutations.
E.g. for selecting (including/excluding) items in an array.
E.g. for handling binary integers.
E.g. `DecToBase` is convenient for displaying an Int64 as a UInt64 or hex. (Versus `Format`.)

That said, equivalent custom functions can be relatively short.
So, no problem if you don't want them.

### DIFF
--- a/source/script.cpp
+++ b/source/script.cpp
@@ -8747,6 +8747,18 @@ Func *Script::FindFunc(LPCTSTR aFuncName, size_t aFuncNameLength, int *apInsertP
 		min_params = 2;
 		max_params = 2;
 	}
+	else if (!_tcsicmp(func_name, _T("DecToBase")))
+	{
+		bif = BIF_DecToBase;
+		min_params = 2;
+		max_params = 3;
+	}
+	else if (!_tcsicmp(func_name, _T("BaseToDec")))
+	{
+		bif = BIF_BaseToDec;
+		min_params = 2;
+		max_params = 2;
+	}
 	else
 		return NULL; // Maint: There may be other lines above that also return NULL.
 
--- a/source/script.h
+++ b/source/script.h
@@ -3377,6 +3377,8 @@ BIF_DECL(BIF_Trim); // L31: Also handles LTrim and RTrim.
 BIF_DECL(BIF_Hotstring);
 BIF_DECL(BIF_InputHook);
 BIF_DECL(BIF_VerCompare);
+BIF_DECL(BIF_DecToBase);
+BIF_DECL(BIF_BaseToDec);
 
 
 BIF_DECL(BIF_IsObject);
--- a/source/script2.cpp
+++ b/source/script2.cpp
@@ -18892,6 +18892,54 @@ BIF_DECL(BIF_VerCompare)
 
 
 
+BIF_DECL(BIF_DecToBase)
+{
+	aResultToken.symbol = SYM_STRING;
+	aResultToken.marker = _T("");
+	unsigned __int64 num = (unsigned __int64)ParamIndexToInt64(0);
+	int base = ParamIndexToInt(1);
+	int len_min = ParamIndexToOptionalInt(2, 1);
+	if (base < 2 || base > 36)
+		_f_throw(ERR_PARAM2_INVALID);
+	if (len_min < 0)
+		_f_throw(ERR_PARAM3_INVALID);
+
+	TCHAR buf[65];
+	_ui64tot(num, buf, base);
+	int len_buf = _tcslen(buf);
+	int len_out = len_min > len_buf ? len_min : len_buf;
+	LPTSTR output = tmalloc(len_out+1);
+
+	LPTSTR cp;
+	int i;
+	for (cp = output, i = 0; i < len_out-len_buf; ++i, ++cp)
+		*cp = '0'; // literal zero
+	tmemcpy(cp, buf, len_buf+1);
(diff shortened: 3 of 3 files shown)


Lexikos:
> These can be surprisingly useful in a variety of circumstances

I can at least agree it would be surprising...

> So, no problem if you don't want them.

To be quite frank, I generally don't like your ideas and don't want to spend any time reading/debating/criticizing them. For anyone *who is not banned from the forum*, I would suggest posting there first, where other users are more likely to voice their support or debate the idea.

Just as an aside, BaseToDec does not convert anything to decimal.

jeeswg:
> BaseToDec does not convert anything to decimal.

It outputs an integer, and integers have no explicit base. So `BaseToInt`?
However:
- The implicit base is arguably 'Dec' since `String(myint)` and `ListVars` give decimal output. In string concatenation/printing, integers are typically treated as 'Dec'.
- 'Dec' is easier on the ear and eye.
- 'Dec' describes the intention better.
- It's common that function names don't describe the return value type (since in typed languages, the *type* specifies the *type*).
- 'Dec' would match other programming languages:

E.g. MS Excel: `HEX2DEC`, `BIN2DEC`, `OCT2DEC`.
E.g. PHP: `hexdec`, `bindec`, `octdec.`
E.g. Rosetta Code examples ([Non-decimal radices/Convert - Rosetta Code](https://rosettacode.org/wiki/Non-decimal_radices/Convert)):
- `ToDecimal` / `FromDecimal`
- `ConvertToDecimal` / `ConvertToBase`
- `DecToBase` / `BaseToDec`
- `Dec2Base` / `Base2Dec`
- `HexDec` / `DecHex`
- `Dec` / `ToBase`

> I generally ... don't want to spend any time reading/debating/criticizing [ideas]

I generally write for the AutoHotkey Community.

They want to know what's being proposed, all the possible benefits, and if it's ultimately rejected, they want it to be clear why it's rejected.
E.g. #212, the 'extending the pattern' comment doesn't say why the pattern is bad, I (and others) currently value the pattern, and see nothing wrong with it.
E.g. #269, I felt I ought to expand my reasoning for why I rejected it, for any interested users/developers.

They also hope that even if someone doesn't make the strongest case for something, that you would still give it the strongest possible consideration.
And there are many many arguments I omit, e.g. AutoHotkey is being ported to Linux, which increases the benefit to users of having a few more functions built-in, and makes AutoHotkey 'ANSI' (i.e. UTF-8) more useful.

> your ideas

Base conversion functions aren't my idea...
Overwhelmingly I submit common functionality that exists in other languages, and/or that a lot of users have asked for.

Since I'm generally not interested in other people's PRs, your relative lack of interest in other people's PRs is not surprising. Although my success rate has been high.

You've refused `A_CRLF` for 10 years, it's not one of my priorities, but it is common in programming languages, simple and uncontroversial to implement, and can't be done via custom code.
Many wonder... Is this little string variable such a problem? Will you ever change your mind, what will it take for you to change your mind (in the next 10 years)? The state of limbo continues for them.

You've had 10 years, so generally we suppose that, for small things, if you valued it, it would already be in AutoHotkey (e.g. you added `StrReplace`/`StrSplit`/`ComCall`).
We know you're a minimalist, and that you've even proposed removing functionality added in AutoHotkey v1.0 (from before you took over).
All of this makes people, including myself, reluctant to even propose any functionality.

You share truly excellent code, but you don't regularly share small bits of code or libraries, which means you haven't experienced the pains of rewriting and reconsidering small functions every time you submit to the forum.
I on the other hand, experienced this in spades, as I was the forum user with the most (over 300) green tricks for 2013-2019.
Or when forum user *just me* proposed `local`, he strongly felt the need for this, just like I did, because he shared/shares a tonne of libraries on the forum.

What I propose is part of complex, and layered, carefully thought-out plans.
I have written and am writing the plurality of code in this programming language.
And I am considering releasing a library with thousands of functions.
It's clear to me that making the library two-way compatible, not writing it twice, would be the best approach.
`ByRef` is the only real obstacle right now, `VarRef(var)` for AHK v1/v2 would be one solution.
`A_IsUnicode`-or-a-replacement/`A_DQ`/`LoopFiles`/`LoopParse`/`Loop (count)`-for-v1/`Range` would be a massive help.

I would favour a more professional system with votes taken for functionality, so that you, and all of us, could always have a better gauge of the true demand for things, but I would still favour granting the main developer a veto.

Btw congrats on AutoHotkey v2.0.0!!!

PR #314, opened 2022-12-20
Added CueText for Edit controls

jNizM:
Working example:
```AutoHotkey
Main := Gui()
Main.AddEdit("xm ym w200")
ED2 := Main.AddEdit("xm y+5 w200")
ED2.CueText("cue banner")
ED3 := Main.AddEdit("xm y+5 w200")
ED3.CueText("cue banner with option true", true)
ED4 := Main.AddEdit("xm y+5 w200 r2")
ED4.CueText("multiline edit control is not supported")
Main.OnEvent("Close", (*) => ExitApp)
Main.Show("AutoSize")
```
Forum: https://www.autohotkey.com/boards/viewtopic.php?f=13&t=70852

### DIFF
--- a/source/script.h
+++ b/source/script.h
@@ -2282,6 +2282,7 @@ struct GuiControlType : public Object
 	static ObjectMemberMd sMembersLV[];
 	static ObjectMemberMd sMembersTV[];
 	static ObjectMemberMd sMembersSB[];
+	static ObjectMemberMd sMembersEdit[];
 
 	static Object *sPrototype, *sPrototypeList;
 	static Object *sPrototypes[GUI_CONTROL_TYPE_COUNT];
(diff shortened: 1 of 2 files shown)


Lexikos:
The convention for methods is to include a verb such as `Set`. Examples:
  - SetFont. There's no obvious way to implement this as a property with the current font handling, so it's only a setter method.
  - SetFormat (DateTime). I don't think there's any way to get the current format, so it's not a property.

The main reason I would not want to have a method named `CueText` is that it implicitly creates a `CueText` property that returns a function object. It is possible to define both a property (getter/setter) and a method named `CueText`, but I'm not sure whether that's the best way to handle it.

`aExact` should be renamed. I assume this is a copy-paste artifact.

Lexikos:
I unintentionally pushed a commit which merged this, but have undone it. There doesn't appear to be any way to re-open the PR, so feel free to merge this with #315 (ignoring what I said about that).

Regarding #313, if you need to fix or add something, you should generally push a new commit (with `push -f` if needed), not create a new PR.

jNizM:
Do you mean something like this?
```
ObjectMemberMd GuiControlType::sMembersEdit[] =
{
	md_member_x(GuiControlType, SetCue, Edit_SetCue, CALL, (In, String, CueText), (In_Opt, Bool32, Activate))
};

ObjectMemberMd GuiControlType::sMembersCB[] =
{
	md_member_x(GuiControlType, SetCue, CB_SetCue, CALL, (In, String, CueText))
};
```

I use GitHub Desktop for nearly everything here.

Lexikos:
Yes.

> I use GitHub Desktop for nearly everything here.

Then you have no excuse, because it certainly has the capability to push changes to a branch, [even force push](https://stackoverflow.com/questions/46866009/is-there-a-force-push-option-in-github-desktop). (For instance, pushing to the Edit_CueText branch would affect this PR.)

PR #315, opened 2022-12-20
Added CueText for ComboBox controls

jNizM:
Working example:
```Autohotkey
Main := Gui()
Main.AddCombobox("xm y+5 w200", ["CB1", "CB2", "CB3"])
CB2 := Main.AddComboBox("x+5 yp w200", ["CB1", "CB2", "CB3"])
CB2.CueText("cue banner")
Main.OnEvent("Close", (*) => ExitApp)
Main.Show("AutoSize")
```
Forum: https://www.autohotkey.com/boards/viewtopic.php?f=13&t=70852


iseahound:
Should use a property unless methods can be chained

jeeswg:
In the thread you linked, Lexikos suggested a property (rather than a method), as has iseahound above, and I agree also.
Unless you want to specify multiple parameters, in which case a `SetCueText` method might be best.

For #314 and #315, it might have been better to combine them. Simpler to assess. Lexikos might agree/disagree, I'm not sure.

Re. both PRs:
- Commit messages are usually: `Sentence ending with a full stop.`
- Comments (I believe) are usually: `// Sentence ending with a full stop.` (Perhaps even for one-worders?)
- Perhaps, in the comments, `Edit control` is better than `edit control`.
- Indentation is Allman style. (Although yours is fine.)

I've basically completed all my PRs, but a style guide might be helpful for all.

Excellent work jNizM, well done for making the effort.
Cheers!

Lexikos:
My comment about naming in #314 also applies to this.

The issue of property vs. method is more complicated for Edit because EM_SETCUEBANNER also controls whether the cue text is shown when the control is focused. On the other hand, I guess that might also be possible for a ComboBox if you send EM_SETCUEBANNER to its Edit control.

Do not merge the pull requests.

jNizM:
Couldn't test it this time because the current alpha gives me the error when compiling:

grafik

jNizM:
After reinstall VS I could compile it. And fixed a typo

```AutoHotkey
Main := Gui()

Main.AddEdit("xm ym w200")

ED2 := Main.AddEdit("xm y+5 w200")
ED2.SetCue("Cue banner")

ED3 := Main.AddEdit("xm y+5 w200")
ED3.SetCue("Cue banner", true)

Main.AddCombobox("x+5 ym w200")

CB2 := Main.AddComboBox("x+5 ym w200")
CB2.SetCue("Cue banner")

Main.OnEvent("Close", (*) => ExitApp)
Main.Show("AutoSize")
```
grafik

Lexikos:
When you copy-paste code, it is best to rename variables like "whole_match" so that they make sense.

Positive FResult values other than FR_ABORTED currently don't have any special meaning, but it is best to return OK on success and a failure code otherwise. Edit_SetCue was returning 0 (NULL) if the control is multi-line.

I don't see much reason to avoid making the attempt when the control is multi-line. I don't think it's appropriate to silently ignore the call, so I've changed it to make the attempt and report the failure.

PR #316, opened 2023-01-05
PCRE 8.30 -> 8.45

sswwaagg:
updated to the last ever supported PCRE1 version, which fixes some internal and other regex related bugs and introduces UTF32 string support.

the resulting binary is larger though


miketestz:
RegexMatch function for binary data does not function normally when PCRE is updated.


When I updated it,


Original


Script Code Partial

This script is part of Ahk2exe(AHKType.ahk).

The updated pcre returns PCRE_ERROR_BADUTF8,16,32 after running the pcret_exec function.

Lexikos:
We would need a patch for the issue @miketestz described before this can be merged. I don't plan to work on it myself at this stage.

Lexikos:
I'm not planning to update v1.1, but left this open in case of updating PCRE for v2. Unfortunately, the AutoHotkey changes for callouts were not preserved, and callouts like `(?C:functionname)` are not working. I do not know whether PCRE itself has made changes to the callout encoding compared to the previous PCRE version.

Also, the update introduces build warnings on x64.

I intend to look into updating the alpha branch to PCRE2 instead, but last time I looked into it, I could not find any information about changes to the API or differences in behaviour, aside from some issues caused by the transition between PCRE versions for PHP. It still isn't a priority for me.

PR #319, opened 2023-03-04
v2: Add BIF_ATan2(y, x).

sswwaagg:
https://www.autohotkey.com/boards/viewtopic.php?p=510371#p510371
give this man a atan2


Lexikos:
This won't be merged into v2.0. Adding a feature implies that the minor version number should be bumped (e.g. v2.1). Pull requests which add features should be targeted at the `alpha` branch.

I have changed the target branch of the PR. I cannot fathom why it is now showing a bunch of commits that are in the target branch (both v2.0 and alpha) and not in the source branch. It merged cleanly, and a diff of the merge only shows the expected changes, originally present in the PR, despite what GitHub shows.

Animan8000:
It should probably be `BIF_ATan2(x, y)`, so basically x being the first parameter and y the second one, making it similar to other functions like Image/Pixelsearch where first the x and then y input is needed, to avoid confusion later on.

RaptorX:
> It should probably be `BIF_ATan2(x, y)`, so basically x being the first parameter and y the second one, making it similar to other functions like Image/Pixelsearch where first the x and then y input is needed, to avoid confusion later on.

Totally agree!, I would be so confused when getting weird results after a `BIF_Atan2(x, y)` call. And then I will get mad at the dev once I figure out the problem 😂.

PR #320, opened 2023-03-06
Added CaseSense check to Map::Create

Descolada:
To turn off case-sensitivity for all Maps, one obvious solution would be `Map.Prototype.DefineProp("CaseSense",{value: "Off"})`. Unfortunately this doesn't work, because the case-sensitivity flag is stored in `ObjectBase.mFlags` and `Map::Create` doesn't update it when the base is copied over. 

This pull request adds a check whether any `Map` object bases contain a `CaseSense` property with a set value and calls `Map::CaseSense` to update it if one is found.

Test cases:
```
test := Map("test", 1, "TEST", 2)
MsgBox "CaseSense: " test.CaseSense "`n" test["test"] " " test["TEST"] ; On, 1 and 2
```
```
Map.Prototype.DefineProp("Casesense",{value: "Off"})
test := Map("test", 1, "TEST", 2)
MsgBox "CaseSense: " test.CaseSense "`n" test["test"] " " test["TEST"] ; Off, 2 and 2
```
```
__ObjDefineProp := Object.Prototype.DefineProp
__ObjDefineProp(Object.Prototype, "CaseSense", {value:"Off"})
test := Map("test", 1, "TEST", 2)
MsgBox "CaseSense: " test.CaseSense "`n" test["test"] " " test["TEST"] ; Off, 2 and 2
```

I am only getting started with C++, so likely a better solution exists and feel free to delete this request in that case.

### DIFF
--- a/source/script_object.cpp
+++ b/source/script_object.cpp
@@ -127,6 +127,17 @@ Map *Map::Create(ExprTokenType *aParam[], int aParamCount)
 
 	Map *map = new Map();
 	map->SetBase(Map::sPrototype);
+	// Check whether CaseSense is already defined in a base and set mFlags if necessary.
+	ExprTokenType casesense;
+	for (Object* that = map->Base(); that; that = that->Base())
+	{
+		if (that->GetOwnProp(casesense, _T("CaseSense")) && casesense.marker_length)
+		{
+			ExprTokenType* param[] = { &casesense };
+			map->CaseSense(ResultToken(), 0, IT_SET, param, 0);
+			break;
+		}
+	}
 	if (aParamCount && !map->SetItems(aParam, aParamCount))
 	{
 		// Out of memory.


Lexikos:
I am not in favour of permitting this scenario, especially if there is a cost to performance or code size.

If you globally change or override default behaviour, you hinder the ability to copy-paste or share code snippets with others.

If you are creating a lot of case-insensitive maps and do not wish to specify the case sensitivity at each occurrence, you can define either a simple function or a subclass which defaults to case-insensitive. This makes it obvious that the code is creating something non-standard.

If you insist on changing the default behaviour of Map, you can do it by redefining the constructor (`Map.Prototype.__New`), making sure to have your constructor call the original constructor function first.

Using DefineProp to redefine a built-in property generally only affects what happens when the script invokes the property. For instance, redefining the Base property with DefineProp does not affect inheritance. There is no precedent for built-in objects checking for properties defined on the prototype.

---

As for the implementation;

`Map.Prototype.DefineProp("CaseSense",{value: "Off"})` overwrites the property setter and getter, thus making it impossible to change the case-sensitivity mode of a map.

`marker_length` has an indeterminate value if the value is not a string. Normally CaseSense can be assigned the strings "1" and "0", which implies that they can be assigned the numbers 1 and 0, because numbers are expected to be converted to strings when necessary. Your check does not allow for pure numeric values.

Checking the bases for an optional property every time a map is created implies a performance penalty for *everyone*, not just users who want to change the default case sensitivity.

Suppose that `marker_length` was 0 when the value is not a string, `Object.Prototype.CaseSense` is redefined by the script and `Map.Prototype.CaseSense` is not. In that case, the check within `Map::Create` would override the default case sensitivity even though `Map.Prototype.CaseSense` is supposed to take precedence over `Object.Prototype.CaseSense`. There should never be a need to check `Object.Prototype` unless the script has explicitly deleted `Map.Prototype.CaseSense`, which wouldn't be very logical.

PR #325, opened 2023-07-02
Fixed memory leak caused by incorrect reference count.

thqby:


Lexikos:
I've applied your fix upstream (v2.0). Thanks.

thqby:
https://github.com/AutoHotkey/AutoHotkey/blob/v2.0/source/script_com.cpp#L1630
This should be `*pCeltFetched = pc;`

and https://github.com/AutoHotkey/AutoHotkey/blob/v2.0/source/ahklib.cpp#L542
```
aResultToken.symbol = SYM_INTEGER; // For _f_return_i() and consistency with BIFs.
var->Get(aResultToken);
if (aResultToken.symbol == SYM_OBJECT)
	aResultToken.object->AddRef();
```

Lexikos:
Yes (for the v2 branches). I've applied those changes.

PR #328, opened 2023-10-14
Update script.h

Animan8000:
Updated warning message since `#NoEnv` directive doesn't exist in v2. Can this warning message even be triggered in v2 at this point?


Ragnar-F:
No, WARNING_USE_ENV_VARIABLE is not used in the v2 code.

PR #331, opened 2024-02-03
Fixed chained dead key behavior

Geobert:
This fixes the bug describe in https://www.autohotkey.com/boards/viewtopic.php?f=14&t=125016

This will also fix https://www.autohotkey.com/boards/viewtopic.php?f=76&t=68434

and all layout that uses chained dead keys (Windows seems to ship some according to https://kbdlayout.info/features/deadkeys)

As far as my testing goes, single dead keys are still working the expected way. 

Also, nothing has been done in the UWP case as nothing can be done: UWP apps doesn’t manage chained dead keys at all from my tests (no AHK involved and it doesn’t work). I’m planning opening a ticket for MS, but I don’t know where, and UWP are supposed to be dropped by MS anyway.

Let me know if more work is needed :)

### DIFF
--- a/source/hook.cpp
+++ b/source/hook.cpp
@@ -2240,6 +2240,12 @@ bool CollectInput(KBDLLHOOKSTRUCT &aEvent, const vk_type aVK, const sc_type aSC,
 	static sc_type sPendingDeadKeySC = 0; // Need to track this separately because sometimes default VK-to-SC mapping isn't correct.
 	static bool sPendingDeadKeyUsedShift = false;
 	static bool sPendingDeadKeyUsedAltGr = false;
+
+	// tracking of chained dead keys
+	static vk_type sChainedDeadKeyVK = 0;
+	static sc_type sChainedDeadKeySC = 0;
+	static bool sChainedDeadKeyUsedShift = false;
+	static bool sChainedDeadKeyUsedAltGr = false;
 	
 	bool transcribe_key = true;
 	
@@ -2299,7 +2305,7 @@ bool CollectInput(KBDLLHOOKSTRUCT &aEvent, const vk_type aVK, const sc_type aSC,
 	TCHAR ch[3] = { 0 };
 	BYTE key_state[256];
 	memcpy(key_state, g_PhysicalKeyState, 256);
-
+	
 	if (sPendingDeadKeyVK && sUwpAppFocused && aVK != VK_PACKET)
 	{
 		AdjustKeyState(key_state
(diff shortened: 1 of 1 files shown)


Geobert:
Hi @Lexikos, ergol.org 1.0 is out now and uses chained dead keys. Any chance to have this merge?

Geobert:
Thank you for the merge! Let me know if any issue occurs by pinging me :)

Lexikos:
Refactoring the code to enable the H option (which lets InputHook take precedence over hotkeys) led me to review the meaning of the flags for ToUnicodeEx. I assume `g_MenuIsVisible ? 1 : 0` was used by the original author (Chris Mallett) because the documentation for ToAsciiEx doesn't actually say what it does, only to pass 1 if a menu is active. At some point, Microsoft documented the effect for ToUnicodeEx, which is to disable Alt-Numpad processing (it's the same for ToAsciiEx, but the documentation doesn't say).

While looking into that, I found that the right combination of parameters can give us the answer without modifying the dead key state, even on older OSes. That way, it doesn't matter how many dead keys there are in a sequence; the function will just give us a result based on the current state. Not changing the state means there is much lower risk of breaking user input just by having hotstrings enabled.

I would have liked for it to be that simple, but unfortunately in the case of non-visible InputHook or UWP apps, the active window doesn't change the dead key state, and therefore we need to do it in the hook to get the right characters. This previously worked by calling ToUnicodeEx and (erroneously) leaving the dead key buffered, but in some instances this gives the wrong result or unwanted side-effects.

  - It should be noted that having a v1.1 or v2.0.17 script active with hotstrings can adversely affect the collection of input by the "fixed" versions, as those earlier versions can alter the dead key state between when they process the dead key and when the sequence is complete.

Now we track which dead keys were pressed in sequence, and call ToUnicodeEx in that sequence prior to the call which translates each new event, if there is a UWP app or non-visible InputHook active. The state is then reset by calling ToUnicodeEx, so there should be no outwardly visible effects. (UWP apps handle dead keys internally, and non-visible InputHook suppresses dead keys, so there should be no pending dead keys under those conditions.)

So in short, chained dead keys should work:
  1. with visible InputHook collection or hotstrings, because each event is translated by a single ToUnicodeEx call which does not change the dead key state.
  2. with non-visible collection or UWP apps, because the hook "replays" the dead key sequence.

The size of the `sPendingDeadKeys` array determines how many dead keys can be recorded and replayed (not counting the live key at the end), and therefore how long a chain can be when used with non-visible InputHook or UWP apps (even though the UWP app itself might never get the right character).

The main changes are by 42f9e3451db455b51ed60ce71de7aba886f002ef.

@Geobert 
I tested the sequence for producing greek characters with Ergo-L, and didn't observe any issues. I can't say that I understand why you were getting an extra character. I did not test Ergo-L with the old method.

In the case of a third dead key in a row, I don't understand your claim that a single call to ToUnicodeEx can restore the normal behaviour of the pressed key. If the current state was reached by pressing sPendingDeadKeyVK and then sChainedDeadKeyVK, wouldn't you need one call for each, in order to reproduce that state?

Geobert:
Wow, that’s a great discovery! I need to build that to test it :D

Geobert:
I build alpha branch and can confirm it works very well! Well done!

PR #332, opened 2024-02-18
v2.1: Implements A_HotstringRecognizer

Descolada:
### Introduction
This PR gives access to the `g_HSBuf` internal variable, allowing read-write to the hotstring recognizer. It was mainly created from the desire to determine the hotstring trigger word case conformity, eg whether `::btw::` was triggered by `btw`, `Btw` or `BTW`. An alternative solution would be to implement something like `A_LastHotstringTrigger`, but this way seems more flexible, as it also provides a way to debug hotstrings more easily, and perhaps implement custom methods around hotstrings (eg restoring the hotstring recognizer after the user tabs to another window, interacts with it, and then tabs back to the old one).

### Documentation
`A_HotstringRecognizer`
Can be used to get or set the hotstring recognizer. If the assigned string is longer than 89 characters then it's automatically trimmed from the beginning to fit 89 characters.

### Example
```
F1::ToolTip '"' A_HotstringRecognizer '"'
F2::A_HotstringRecognizer := "Btw"
:B0:btw::
{
	MsgBox 'ThisHotkey: "' ThisHotkey '"`nA_HotstringRecognizer: "' A_HotstringRecognizer '"'
}
```
F1 displays the current recognizer state
F2 sets it to `Btw`, after which pressing spacebar will trigger the hotstring.

### DIFF
--- a/source/lib/vars.cpp
+++ b/source/lib/vars.cpp
@@ -489,6 +489,23 @@ BIV_DECL_W(BIV_Hotkey_Set)
 		g_HotkeyModifierTimeout = value;
 }
 
+BIV_DECL_R(BIV_HotstringRecognizer)
+{
+	wcscpy(_f_retval_buf, g_HSBuf);
+	_f_return_p(_f_retval_buf, g_HSBufLength);
+}
+
+BIV_DECL_W(BIV_HotstringRecognizer_Set)
+{
+	auto value = BivRValueToString();
+	auto size = wcslen(value);
+	if (size > (HS_BUF_SIZE - 1)) // Avoid overflow
+		_f_throw_value(ERR_INVALID_LENGTH);
+	wcscpy(g_HSBuf, value);
+	g_HSBufLength = size;
+	g_HShwnd = GetForegroundWindow(); // Also reset active window
+}
+
 BIV_DECL_R(BIV_MenuMaskKey)
 {
 	if (!g_MenuMaskKeyVK && !g_MenuMaskKeySC)
--- a/source/lib/vars.cpp
+++ b/source/lib/vars.cpp
@@ -498,10 +498,10 @@ BIV_DECL_R(BIV_HotstringRecognizer)
 BIV_DECL_W(BIV_HotstringRecognizer_Set)
 {
 	auto value = BivRValueToString();
-	auto size = wcslen(value);
-	if (size > (HS_BUF_SIZE - 1)) // Avoid overflow
-		_f_throw_value(ERR_INVALID_LENGTH);
-	wcscpy(g_HSBuf, value);
+	auto len = wcslen(value);
+	auto size = min(len, HS_BUF_SIZE - 1); // Avoid overflow
+	wcsncpy(g_HSBuf, &value[len-size], size);
+	g_HSBuf[size] = '\0';
 	g_HSBufLength = size;
 	g_HShwnd = GetForegroundWindow(); // Also reset active window
 }
--- a/source/script.cpp
+++ b/source/script.cpp
@@ -185,6 +185,7 @@ VarEntry g_BIV_A[] =
 	A_w(FileEncoding),
 	A_wx(HotkeyInterval, BIV_Hotkey, BIV_Hotkey_Set),
 	A_wx(HotkeyModifierTimeout, BIV_Hotkey, BIV_Hotkey_Set),
+	A_w(HotstringRecognizer),
 	A_x(Hour, BIV_DateTime),
 	A_(IconFile),
 	A_w(IconHidden),
--- a/source/script.h
+++ b/source/script.h
@@ -3168,6 +3168,7 @@ BIV_DECL_RW(BIV_SendMode);
 BIV_DECL_RW(BIV_SendLevel);
 BIV_DECL_RW(BIV_StoreCapsLockMode);
 BIV_DECL_RW(BIV_Hotkey);
+BIV_DECL_RW(BIV_HotstringRecognizer);
 BIV_DECL_R (BIV_KeybdHookInstalled);
 BIV_DECL_R (BIV_MouseHookInstalled);
 BIV_DECL_RW(BIV_MenuMaskKey);


Descolada:
Adding some more thoughts on this. 
1. This would probably make writing malicious code a tiny bit easier, as newbies could simply access `A_HotstringRecognizer` to keylog. However, doing the same with an InputHook takes about 4 short lines of code and there are plenty of examples of that available on the web, so I'd argue that it doesn't really matter.
2. `A_HotstringRecognizer` could also be replicated by using an InputHook. It would require knowing exactly how the hotstring recognizer works internally and replicating it, which is not an easy task (taking account the keys which reset it, window focus changes, mouse input etc). It makes more sense to just give access to what AHK is using itself.
3. The alternatives. I think the best option, ignoring backwards and forwards compatibility, would be for `ThisHotkey` to be in the same case as the trigger. Eg if `:X:btw::MsgBox ThisHotkey` was triggered by `Btw` then `ThisHotkey` should be `:X:Btw` not `:X:btw`. The user could still figure out which hotstring it is by doing a case-insensitive comparison: `if ThisHotkey = ":x:btw:" ...`. Another option would be to introduce a new variable such as `ThisHotstring` which would contain the trigger word in the correct case. It could possibly be backwards compatible for hotstrings defined with the hotstring syntax? But not for hotstrings defined with the Hotstring function, as it is expected for the callback function to require just one parameter. 
`A_LastHotstring` would be the most realistic alternative, which compared to the proposed solution would be without any other benefits around hotstrings.

Lexikos:
I suspect that this implementation may be prone to race conditions, such as:
  - Content is written into the buffer by both threads at the same time, and the content and/or length of the buffer become inconsistent.
  - Content is read from the buffer by the script while the hook is writing to or resetting it, resulting in inconsistency. For instance, the length may not match the content, or it might be possible for old and new content to be mixed.

Such issues can be difficult to understand or debug when they occur.

> to determine the hotstring trigger word case conformity, eg whether `::btw::` was triggered by `btw`, `Btw` or `BTW`. 

The limitations being that the hotstring must disable backspacing (`B0`) and must not reset the hotstring recognizer (no `Z`).

There's nothing to stop events from modifying the buffer before the hotstring has a chance to inspect it, although for most uses in practice it might not be an issue. It can be proven easily by using SendLevel and Send to trigger the hotstring but follow it with additional events.

Being able to inspect the text which triggered the hotstring would be necessary for RegEx hotstrings. The hardest part might be sending the captured text back to the script in a thread-safe way. A solution could be applied to normal hotstrings, or one could just use a RegEx hotstring if one needs to know what was typed.

> It would require knowing exactly how the hotstring recognizer works internally and replicating it

The difficult part is done for you: events are interpreted into characters once by the keyboard hook and fed into both the hotstring recognizer and all InputHooks. I suppose that you would usually just need to know what was typed after some event (e.g. hotstring trigger) occurs. If InputHook had circular buffering, you wouldn't need to reset it.

> (taking account the keys which reset it, window focus changes, mouse input etc)

That's basically just a list of keys and a focus check. If you don't want to miss any conditions, you can read the source code. (I think it isn't necessary to perfectly replicate the program's behaviour; the user's habits are more important.)

> I think the best option ... would be for ThisHotkey to be in the same case as the trigger.

The last text you typed to trigger the hotstring is not *the hotstring itself*. Hotkeys can be triggered multiple ways, yet ThisHotkey is always the original definition of the first hotkey variant, not even matching the current variant, let alone the exact combination of keys which triggered it. (Keep in mind that `!` allows both left and right Alt, while `*` allows additional modifiers, as do custom combinations.) It would be incongruous for hotstrings to behave otherwise.

It would also be incongruous for a variable to contain the *originally defined* hotstring options but the actual typed trigger text. If the script is to compare the variable to a hard-coded string to identify the hotstring, it is important that the options are predictable. A string with the current options would be less useful for that purpose. Also be aware that the hotstring recognizer uses the locale mode for comparisons, so some users would need to use StrCompare rather than `=` if they are not using the exact case.

> But not [*backwards compatible*] for hotstrings defined with the Hotstring function, as it is expected for the callback function to require just one parameter.

A function object with MinParams = 2 would currently cause an exception to be thrown when the hotstring is triggered, although a function with MinParams = 1 and MaxParams = 2 would be called with one parameter.

Descolada:
@Lexikos 
> I suspect that this implementation may be prone to race conditions

I suspect you are correct, as I've totally missed the fact that hooks are set up in another thread. Creating a thread-safe implementation would most likely have a serious impact on the general performance of hotstring processing. I think some other approach would be better, so I'm closing this pull request.

> That's basically just a list of keys and a focus check. If you don't want to miss any conditions, you can read the source code. (I think it isn't necessary to perfectly replicate the program's behaviour; the user's habits are more important.)

I can definitely replicate all the conditions, but I cannot be 100% sure that the internal hotstring recognizer is in sync with my custom recognizer. For example, a second script might send a keystroke (perhaps triggering a hotstring in the first script?) and immediately switch to another window. In that case `g_HSBuf` will be reset on the next keystroke because `g_HShwnd` has changed, but my custom recognizer might lag behind and think that the keystroke was sent to the second window, thus resetting the buffer right away. Probably a rare situation, but could happen.

> A function object with MinParams = 2 would currently cause an exception to be thrown when the hotstring is triggered, although a function with MinParams = 1 and MaxParams = 2 would be called with one parameter.

This sounds like the best solution so far, as it would be in sync with the internal hotstring recognizer and also most likely wouldn't break existing scripts. An exception might be if the user has decided to use an optional second parameter as something to bind arguments to:
```
Hotstring ":C:abc", HotProc
Hotstring ":C:ABC", HotProc.Bind(, "Upper")
Hotstring ":C:Abc", HotProc.Bind(, "Title")
HotProc(HotstringName, style?) {
    MsgBox HotstringName ", " (style ?? "case unset")
}
```
Would you consider this to be acceptably rare to ignore it? Otherwise a companion to `A_ThisHotkey` such as `A_ThisHotstring` might be the only suitable option to access the typed case of the trigger word (other than writing a custom recognizer via an InputHook)...

Lexikos:
I had forgotten I had merged this into a local test branch, so accidentally noticed that it causes a build warning/error on x64. Better watch out for that before pushing commits in future.

> I think some other approach would be better, so I'm closing this pull request.

Regex hotstrings (capable of capturing parts or all of the buffered text) would also have this difficulty, but greater utility, so I would approach it from that angle.

> An exception might be if the user has decided to use an optional second parameter as something to bind arguments to:

The hotstring will be calling the BoundFunc, not the function to which it is bound. Parameters which have values bound to them are not relevant to the caller of the BoundFunc. If the target function has three parameters `(a, b, c)` and you bind only the second `(, b)`, the BoundFunc accepts two parameters: `(a, c)`.

However, each BoundFunc has a signature like `(args*)` as per the documentation. If that isn't changed, the caller of Hotstring would need to specifically redefine MinParams, MaxParams and IsVariadic before passing the BoundFunc, or use a closure instead. Likewise, redefining those properties has an effect on the validation performed by Hotkey, OnMessage, etc. (Hotstring currently doesn't perform validation).

PR #337, opened 2024-07-05
v2.1: Allow passing object with Ptr property in NumPut

Descolada:
This PR is related to [this forum thread](https://www.autohotkey.com/boards/viewtopic.php?f=37&t=131209). 

I believe it would be useful to allow `NumPut("Ptr", obj, target)` where `obj` contains a `Ptr` property, and this PR implements that. I didn't see a good way to avoid doing the comparison `TypeOfToken(*aParam[n_param]) == SYM_OBJECT)` twice (first to validate that if the type is `ptr` then the number may also be an object, and second to extract the `ptr` property from it). I considered simply validating for the number to not be a string, but that meant that non-ptr types also accepted objects with ptr properties. However, better error validation seemed worth more than the slight performance benefit.

### DIFF
--- a/source/lib/interop.cpp
+++ b/source/lib/interop.cpp
@@ -200,7 +200,7 @@ BIF_DECL(BIF_NumPut)
 
 		// See comments in NumGet about the following section:
 		if (!op.num_size
-			|| !TokenIsNumeric(token_to_write)
+			|| !(TokenIsNumeric(token_to_write) || (op.is_integer && op.num_size == sizeof(void*) && TypeOfToken(*aParam[n_param]) == SYM_OBJECT)) // Validate for numbers OR whether the type name is Ptr and value is an object (checking for the Ptr property is done later on)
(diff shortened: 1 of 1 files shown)


Lexikos:
I am currently not inclined to accept any increase in code or documentation size to further support the use of **NumPut** to fill structs, given that a primary feature of v2.1 is more direct struct support.

vmech:
Then maybe it is possible to add this feature to v2.0 ?
The vast majority of people use v2.0.

Lexikos:
No, that would violate the versioning scheme.

Descolada:
Given that struct support *does* alleviate the need for this PR and I don't have other strong arguments for it besides "for the sake of completeness" (NumPut supporting this feature fits in with DllCall/ComCall and NumGet source argument pretty well), I'm inclined to close this PR. @vmech perhaps you have a better argument for implementing this?

vmech:
@Descolada no, I have long been accustomed to standard 2.0 methods.

PR #338, opened 2024-07-06
Experimental: added disableWindowFiltering to manifest

Descolada:
EnumWindows by default skips over some UWP / Microsoft Store apps, making them very hard to detect with AutoHotkey. Example: [To Do app](https://www.autohotkey.com/boards/viewtopic.php?p=506799). FindWindowEx can be used as a work-around, but is very inconvenient to use.

Adding [disableWindowFiltering](https://learn.microsoft.com/en-us/windows/win32/sbscs/application-manifests#disableWindowFiltering) to the package manifest disables the filtering implemented by Microsoft and allows EnumWindows to detect the previously undetectable top-level windows. 

The following example is for detecting the Notification Centre window in Windows 11 that appears when the time/date is clicked in the taskbar:
```
Loop {
	ToolTip WinExist("Notification Centre") ; if testing in Windows 10, replace with "Date and Time Information"
	Sleep 100
}
```

I have no idea what side effects this will cause, but certainly it would break some previous scripts. However, it seems like a *very* useful change.

### DIFF
--- a/source/resources/AutoHotkey.exe.manifest
+++ b/source/resources/AutoHotkey.exe.manifest
@@ -63,4 +63,11 @@
     </v3:requestedPrivileges>
   </v3:security>
 </v3:trustInfo>
+<asmv3:application xmlns:asmv3="urn:schemas-microsoft-com:asm.v3">
+	<asmv3:windowsSettings xmlns="http://schemas.microsoft.com/SMI/2005/WindowsSettings">
+		<!-- This allows enumerating immersive (UWP / Microsoft Store) windows. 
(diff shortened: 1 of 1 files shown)


Descolada:
Another window that is nearly uninteractable without this addition is the touch keyboard app. In Windows 10 it can be opened by right-clicking the taskbar and selecting "Show touch keyboard button", then a keyboard icon will appear near the clock and the window opens by clicking it. With disableWindowFiltering added, it can be targeted with `WinExist(, "Microsoft Text Input Application")`, without it the window doesn't have any distinguishing features. This problem was encountered in a [Discord thread](https://discord.com/channels/115993023636176902/1292568199434469560).

PR #340, opened 2024-09-26
Optimized DllCall with param count <= 4 for x64

DartVanya:
If there is no stack arguments we can jump over ineffective code that copying stack arguments.
Tested on:
```
DllCall(CallbackCreate(TestFunc1,, 5), "Ptr",228, "Str","Test string", "Ptr",1, "Ptr",1, "Ptr",1337)

DllCall(CallbackCreate(TestFunc2,, 2), "Ptr",228, "Str","Test string")


TestFunc1(param1, param2, param3, param4, param5) {

	MsgBox('param1`t=`t' param1 '`nparam2`t=`t"' StrGet(param2) '"`nparam5`t=`t' param5 )
}

TestFunc2(param1, param2) {

	MsgBox('param1`t=`t' param1 '`nparam2`t=`t"' StrGet(param2))
}
```

### DIFF
--- a/source/lib/DllCall.cpp
+++ b/source/lib/DllCall.cpp
@@ -78,7 +78,7 @@ struct DYNAPARM
 
 #ifdef _WIN64
 // This function was borrowed from http://dyncall.org/
-extern "C" UINT_PTR PerformDynaCall(size_t stackArgsSize, DWORD_PTR* stackArgs, DWORD_PTR* regArgs, void* aFunction);
+extern "C" UINT_PTR PerformDynaCall(size_t stackArgsCount, DWORD_PTR* stackArgs, DWORD_PTR* regArgs, void* aFunction);
 
 // Retrieve a float or double return value.  These don't actually do anything, since the value we
 // want is already in the xmm0 register which is used to return float or double values.
@@ -215,7 +215,7 @@ void DynaCall(void *aFunction, DYNAPARM aParam[], int aParamCount, DWORD &aExcep
 	int params_left = aParamCount, i = 0, r = 0;
 	DWORD_PTR regArgs[4];
 	DWORD_PTR* stackArgs = NULL;
-	size_t stackArgsSize = 0;
+	size_t stackArgsCount = 0;
 
 	if (!register_return)
 	{
@@ -230,8 +230,8 @@ void DynaCall(void *aFunction, DYNAPARM aParam[], int aParamCount, DWORD &aExcep
 	// Copy the remaining parameters
 	if(params_left)
 	{
-		stackArgsSize = params_left * 8;
-		stackArgs = (DWORD_PTR*) _alloca(stackArgsSize);
+		stackArgsCount = params_left;
+		stackArgs = (DWORD_PTR*) _alloca(stackArgsCount*sizeof(DWORD_PTR));
 
 		for(int i = 0; i < params_left; i ++)
 			stackArgs[i] = DynaParamToElement(aParam[i+4]);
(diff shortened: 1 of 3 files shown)


Lexikos:
Was any testing done to verify that has a measurable and consistent benefit to performance and does not negatively impact performance of calls with >4 parameters?

DartVanya:
I did some benchmarks on this script below. Performance isn't change much for both cases, but in my test it 3-4% faster for <4 args and almost no difference for >4 args. 

image


```
cb2 := CallbackCreate(TestFunc2,, 2)
cb7 := CallbackCreate(TestFunc7,, 7)

max_iterations := 100 * 1000

Loop 10
{
    start := QPC()
    Loop max_iterations
        res1 := DllCall(cb2, "Ptr",282, "Int",300, "Ptr")
    time := QPC() - start
    r1 := "Time for " max_iterations " calls of 2 argument func:`t" time " us = " time / 1000 " ms"

    start := QPC()
    Loop max_iterations
        res2 := DllCall(cb7, "Ptr",282, "Int",300, "Ptr",1, "Ptr",1, "Ptr",1337, "UInt",2, "UInt",3, "Ptr")
    time := QPC() - start
    r2 := "Time for " max_iterations " calls of 7 argument func:`t" time " us = " time / 1000 " ms"

    result .= r1 "`n`n" r2 "`n--------------------------------------------`n"
}


MsgBox result


TestFunc2(param1, param2) {
    return param1 + param2
}

TestFunc7(param1, param2, param3, param4, param5, param6, param7) {
    return param1 + param2
}

QPC() {
    static c := 0, f := 1, _f := DllCall("QueryPerformanceFrequency", "int64*", &f)
    DllCall("QueryPerformanceCounter", "int64*", &c)
    return c * 1000000 // f
}
```

[AHK DllCall asm bench.xlsx](https://github.com/user-attachments/files/17989791/AHK.DllCall.asm.bench.xlsx)

Lexikos:
> cb7 := CallbackCreate(TestFunc7,, 7)

The benchmark should be measuring the cost of calling the function, but the mere presence of additional parameters in TestFunc7 increases the execution time of the function. The call itself becomes a smaller fraction of the overall time, so any difference in the performance of the call is less noticable.

If you need a native function to call safely, consider MulDiv. It has a defined result for overflow and divide by zero, so any input is safe. But I suppose even just calling the same callback function in both tests would be fine.

Is there any meaning in running 100,000 iterations and then showing 10 sets of times? An average of all iterations is easier to read and gives a more consistent result.

Each set of 100,000 iterations gives very different results, and the results vary more than 3-4% within the test and on separate runs of the overall script. Based on this, you can't claim a 3-4% increase in performance. We need to establish a reliable baseline and margin of error first, before trying to measure a difference between pre- and post-optimization.

Did you build both binaries used in your benchmarks? I found that rebuilding the EXE from the same source as the official v2.1-alpha.19 release consistently gave different times than the release itself (which could be explained by my release process using different settings or build tools than when I build from the IDE). Again, the differences exceeded 3-4% and sometimes *reliably* increased one half of the test more than the other. Running at different times gives different results.

Repeatedly running your test on different builds including the v2.1-alpha.19 release, a rebuild and a build with your PR merged, I see no clear difference in the results.

Also, I figured that if avoiding that small part of the ASM function was helpful, it would be even more helpful to inline a four-register call to the target function into DynaCall itself. This can be done by just calling the function...
```C++
		if (!stackArgsSize)
			Res.UIntPtr = ((UINT_PTR(*)(UINT_PTR, UINT_PTR, UINT_PTR, UINT_PTR))aFunction)(regArgs[0], regArgs[1], regArgs[2], regArgs[3]);
		else
			Res.UIntPtr = PerformDynaCall(stackArgsSize, stackArgs, regArgs, aFunction);
```
...but again, there's no clear difference in the benchmark results.

PR #341, opened 2024-09-27
Increase MAX_HOTSTRING_LENGTH from 40 to 400

wyattscarpenter:
(I haven't tested this change, but hopefully it just works, given that this is just an arbitrary limit.)

(If the documentation, such as https://www.autohotkey.com/docs/v2/Hotstrings.htm#remarks, also needs to be changed, I also haven't done so.)

The comment for this constant (which I have kept for the new quantity ha ha) mentions "hard to imagine a need for more than this"; after years, we have found such a need to exceed 40. https://github.com/alexmick/emoji-to-ahk/blob/f7471fceda3f23cf36a83ba150c6a22fbb2796f1/custom_short_names.py#L2 . In this project that lets you type emoji based on their name (in a bbcode-style, :like_this:), some of the emoji names exceed 40 characters, and so something awkward has to be done to deal with that; why not allow the project to be perfect, instead? :)

I don't think those emoji names are necessarily the official unicode names for the characters, but I also want to be able to type such unicode names as bbcode-style hotstrings for another project of mine (which I may never get to). I don't think there's any limit on how long a unicode character name can be, but according to this random Quora answer from two years ago, the longest unicode character name is is 83 characters long https://www.quora.com/Which-are-the-longest-Unicode-characters/answer/Thomas-Phinney . 400 should be ample for this as well.

So why 400?

1. It's amusing simple to just add a 0.
2. It's still obviously an arbitrary (instead of technical) limit, unlike my favorite numbers, which are all powers of two.
3. An order of magnitude greater than our current need, so inshallah we will never have to increase it again.
4. Does not exceed INT_MAX, in case that matters.
5. A comment in https://github.com/AutoHotkey/AutoHotkey/blob/alpha/source/hotkey.cpp notes that "The length is limited for performance reasons, notably so that the hook does not have to move memory around in the buffer it uses to watch for hotstrings". 400 is still a pretty small amount of memory for a modern computer (although this comment suggests the hook might have to be updated, perhaps?)

### DIFF
--- a/source/hotkey.h
+++ b/source/hotkey.h
@@ -334,8 +334,8 @@ private:
 
 ///////////////////////////////////////////////////////////////////////////////////
 
-#define MAX_HOTSTRING_LENGTH 40  // Hard to imagine a need for more than this, and most are only a few chars long.
-#define MAX_HOTSTRING_LENGTH_STR _T("40")  // Keep in sync with the above.
+#define MAX_HOTSTRING_LENGTH 400  // Hard to imagine a need for more than this, and most are only a few chars long.
+#define MAX_HOTSTRING_LENGTH_STR _T("400")  // Keep in sync with the above.
 #define HOTSTRING_BLOCK_SIZE 1024
 typedef UINT HotstringIDType;
 


Descolada:
Your project has about 5700 hotstrings, and with the current limit of 40 unicode characters that means the memory footprint is about 0.4MB only for the hotstring triggers. With your proposal you would increase that to 4MB, and the net gain would be to include 6 extra hotstrings without modifications (out of >5700!). Note that the 6 hotstrings have extremely hard to remember names, whereas the aliases are much simpler (eg black_right_pointing_double_triangle_with_vertical_bar -> fast_forward), so it's questionable whether you'd even actually use those hotstrings. So I don't think the cost/benefit ratio justifies a 10-fold limit increase.

The longest word in the English language is 45 characters long, so perhaps a limit of 50-60 characters would technically make more sense. However, as the limit of 40 hadn't been hit for many years then maintaining the status quo would be okay in my opinion.

wyattscarpenter:
It's ok with me; I have 4 spare MB of RAM.

(Digression: I'm not sure how many hotstrings other users of AHK use, or how much RAM they're willing to lose to it. Maybe about 5700 is a good benchmark figure to consider for a user of AHK, as the distribution probably follows a power law? 🤔 So the typical user probably has like 10 but the mean user probably has like 1,000. So I think it's useful to compute quantities about this, as you have done.)

I agree that it is a small portion of the hotstring triggers. Probably a larger portion if I included all unicode names, although I can't substantiate that with statistics at the moment. I also agree that the short aliases are much nicer to use than the long names. However, since the long names are laid out somewhat systematically, they are not as hard to remember as may first appear... for instance the related characters 
```
23E9 ⏩ BLACK RIGHT-POINTING DOUBLE TRIANGLE
23EA ⏪ BLACK LEFT-POINTING DOUBLE TRIANGLE
23EB ⏫ BLACK UP-POINTING DOUBLE TRIANGLE
⋮
```
all follow the same pattern, and so if you work with unicode('s character names) often you'll probably be able to remember these.

There is also a great virtue in being able to do everything in the complete verbose way in addition to the fast aliased way. For instance, you can use any standard unicode character name resources with this script in that case, instead of having weird holes filled with aliases. (Now that I type this I realize I'm just describing the virtue of a standard.)

Anyway, thanks for the review of the idea! If some number lower than 400 makes this PR more palatable, I'm happy to go for it. Perhaps a nice multiple of 100...

Banaanae:
Would it be possible to add a directive for this?

Lexikos:
The trigger text of a hotstring never changes, and therefore only the actual length of the string is allocated for each hotstring. The memory cost of increasing MAX_HOTSTRING_LENGTH is a fixed amount not dependent on the number of hotstrings.

As far as I can tell, this constant is used only indirectly to provide a fixed size for the following buffers:
  - A stack-allocated buffer used while carrying out a replacement (`Hotstring::DoReplace`). Having a fixed size means less overhead, and generally instant allocation (just adjusting the stack pointer). Due to the nature of the function and stack allocation, a difference of 360 bytes should not have any impact on performance or memory usage.  
  - The hotstring collection buffer (`g_HSBuf`), which is statically allocated. Only one exists.
    
    Increasing MAX_HOTSTRING_LENGTH by 360 would increase HS_BUF_SIZE by 720. The current total value is only 90.  

    Each time a character is added to the hotstring buffer while at capacity, half of the buffer's content (a little more than the longest potential hotstring) is moved to make space (and the rest is discarded). Increasing MAX_HOTSTRING_LENGTH therefore has a performance cost, applied on each character input. This could be mitigated, but only at the cost of complexity and/or memory usage.

To implement a directive, both of these buffers would need to be dynamically-allocated, which implies a small increase in code size (and so does the directive itself). It would probably be acceptable.

wyattscarpenter:
Thanks for your investigation! Seems like increasing the limit is an even better deal than we thought! I appreciate the new depth you've brought to the discussion.

I do not think a directive is a promising avenue to pursue. In general the number of options and directives should be minimized, and the best behavior should simply be done by default; I do not think this difference in behavior is worth having two versions of if that can be selected by directive. I think upping the limit directly will just work out great for everyone.

g_HSBuf seems like it should be implemented as a ring buffer. At least from this description. However, leaving it the way it is is probably fine.

Lexikos:
As I said,

> This could be mitigated, but only at the cost of complexity and/or memory usage.

I suppose that using a ring buffer would slightly increase the complexity of comparing each hotstring against the buffer content. Depending on the number and nature of hotstrings used, it may be possible that the current method performs better, especially given the current size of the buffer.

I should emphasize "while at capacity" - for a buffer of 810 (400 * 2 + 10), every 400 (or so) characters typed, it would move 400 (or so) characters one time. Maybe it is negligible.

I think typing a 40+ character trigger string does not have enough practical use to warrant a change to the limit.

It would be possible to "fake" a longer hotstring by using InputHook to verify the full trigger text after the first 40 characters triggers a hotstring.

Descolada:
If changing the storage option of hotstrings is on the table, then perhaps use a trie instead and an algorithm such as Aho–Corasick to search for matches in linear time? There was a proof of concept of the idea in this [Discord post](https://discord.com/channels/115993023636176902/1253556116785463357/1253556116785463357), and there are Aho-Corasick C++ implementations available out there so it wouldn't need to be written from scratch.
In addition to offering speed increases it would also remove trigger word length limits.

PR #343, opened 2024-12-27
CallbackCreateSync

DartVanya:
CallbackCreate - add '$' option to create callback that is safe to pass to multithreaded APIs. If callback thread not equal to AHK thread execution will be synchronized via SendMessage invoke.

### DIFF
--- a/source/hook.h
+++ b/source/hook.h
@@ -41,6 +41,7 @@ enum UserMessages {AHK_HOOK_HOTKEY = WM_USER, AHK_HOTSTRING, AHK_USER_MENU, AHK_
 	, AHK_HOOK_SYNC // For WaitHookIdle().
 	, AHK_INPUT_END, AHK_INPUT_KEYDOWN, AHK_INPUT_CHAR, AHK_INPUT_KEYUP
 	, AHK_HOOK_SET_KEYHISTORY
+	, AHK_INVOKE
 };
 // NOTE: TRY NEVER TO CHANGE the specific numbers of the above messages, since some users might be
 // using the Post/SendMessage commands to automate AutoHotkey itself.  Here is the original order
--- a/source/lib/CCallback.cpp
+++ b/source/lib/CCallback.cpp
@@ -43,6 +43,7 @@ struct RCCallbackFunc // Used by BIF_CallbackCreate() and related.
 	UCHAR actual_param_count; // This is the actual (not formal) number of parameters passed from the caller to the callback. Kept adjacent to the USHORT above to conserve memory due to 4-byte struct alignment.
 #define CBF_CREATE_NEW_THREAD	1
 #define CBF_PASS_PARAMS_POINTER	2
+#define CBF_SYNCH_EXECUTION	4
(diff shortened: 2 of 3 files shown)


Lexikos:
I preferred to avoid handling any window messages in a way which allows execution of arbitrary code. I touched on this in [RegisterSyncCallback for v2](https://www.autohotkey.com/boards/viewtopic.php?t=98226).

Perhaps that concern is unfounded, as there are other methods that a process with sufficient permission could use to inject code into our process, and any number of window messages which could be sent with invalid pointers or malformed data. The message would be blocked by default if sent from a lower-integrity process (due to UIPI).

This does have the benefit of simplicity, which is good for code size, performance and maintainability.

I think there would be some benefit in having this be the default behaviour, as allowing the callback to proceed on the wrong thread is likely to only cause frustration. Crashes are the best thing that can happen; multiple threads trying to use the same internal global variables can have unpredictable results, and lines apparently executing out of sequence can be quite confusing. Some errors are subtle or might appear only under some conditions, so might not bring an author's attention to the need for synchronization.

I would consider backward-compatibility a non-issue because executing script on other threads was never supported, but some authors might have been utilizing it successfully, through some miracle.  I think that few users understand the program internals (or multi-threading) well-enough to be a good judge of when it is safe to multi-thread script functions (unless the judgement is "it's never safe").

One reason to make synchronization "opt in" would be that future changes could make it safer to execute script on other threads (such as by reducing reliance on internal global/static variables), but given that any script function intended to be used that way would have to be carefully designed, it would probably be better to require the author to explicitly opt out of synchronization.

If we add an option, I do not think I would choose to represent it with `$`. It Synchronizes using SendMessage. I suppose you might have chosen `$` over `S` because `FAST` contains `S`. The existing option parsing is very lazy, and just checks for letters in the string rather than parsing each option. CallbackCreate should perhaps be changed to validate the options as per the documentation, to detect errors and reserve other characters/words for future use.

PowerPacer:
(written by GPT)
I reproduced the failure this PR is intended to prevent using a locally built x64 Debug executable from the alpha.31 source.
My application mistakenly exposed an AHK-backed WinRT event delegate as IAgileObject. Windows then invoked two CallbackCreate callbacks concurrently on separate Windows.UI worker threads while the main thread was also executing AHK.

The full dump shows:
- worker A: RegisterCallbackCStub -> UiSettingsQueryInterface -> Line::ExpandArgs -> malloc
- worker B: RegisterCallbackCStub -> UiSettingsInvoke -> Line::ExpandArgs -> free
- the second worker asserts in the debug CRT while freeing Line::sDerefBuf
- g_nThreads was only 3 with g_MaxThreadsTotal 10, so this was not pseudo-thread-limit exhaustion

Removing IAgileObject from the delegate fixed the application-side contract error. Nevertheless, ordinary CallbackCreate currently permits foreign native threads to enter shared AHK execution state concurrently, resulting in heap corruption rather than rejection, synchronization, or a documented precondition.
This appears to be the exact case addressed by the proposed $ callback option. Would you prefer a separate bug report and minimal reproduction linked to this PR? I can provide sanitized debugger stacks and matching source/PDB details privately if useful.

PR #345, opened 2025-07-07
Add Windows on ARM support

t-satoru:
Now we can run AHKv2 on Windows on ARM!

(I admit that I used Cursor in helping me achieve this target)

image


t-satoru:
I realized that I only solved compiling problems, install scripts are not updated
Will open a new PR after everything is ready

Jy-EggRoll:
That's great! We'll be able to use native ARM64 AHK in no time!

PR #352, opened 2025-11-30
KeyHistory: Add options to filter display of mouse and/or keyboard events

ThioJoe:
This updates the KeyHistory function to allow users to hide mouse or keyboard events from the history window. Feel free to adjust it or use the logic as scaffolding or something if you'd implement it different yourself.

### **Motivation:**

It can be annoying when trying to debug or observe hotkey behavior while the history is getting inundated with mouse clicks or scrolls, for example.

### **Changes:**

- Updated `KeyHistory` signature to accept two new optional parameters. 
Full new signature is: `KeyHistory([MaxEvents, ShowMouse, ShowKeyboard])`
- Added global flags `g_ShowMouseKeyHistory` and `g_ShowKeyboardKeyHistory` (Both default `true`)
- Modified `GetHookStatus` in `hook.cpp` to skip adding entries to the text buffer if the specific input type is disabled

### Important Notes:
- For the sake of simplicity I made it so it doesn't change how key events are collected behind the scenes. This means key events will still take up "slots" in the key event history even if they aren't shown. It simply hides them from display. But on the upside it adds no additional performance penalty for general script running (the additional logic only comes into play when displaying or refreshing the list for the KeyHistory window).
- In `GetHookStatus` I just have a switch case to check the event against the 9 VK values of the mouse keys. There might be a more elegant way to do this, but on a quick search I couldn't find any pre-existing array of the mouse keys or anything.
- The switch case assumes that any mouse events will always have a `.vk` value associated with them. Any keypresses that don't have any `VK` value will be considered as keybaord key presses and fall through to the default case.

### **Other Remarks:**
- The new parameters are optional. The default values cause `KeyHistory` to behave as it currently does.
- Theoretically both mouse and keyboard events could be hidden if the user wants. The window still shows fine, just no events will show, as expected.
- I did compile and test it myself to make sure it works

### Various Example Usages:

```ahk

; Examples: Only show mouse keys in history
KeyHistory(unset, unset, False)
KeyHistory(unset, True,  False)

; Examples: Only show keyboard keys
KeyHistory(unset, False, unset)
KeyHistory(unset, False, True)

; Examples: Set max items
KeyHistory(500)
KeyHistory(500, unset)
KeyHistory(500, unset, unset)

; Examples: Set max items and don't show mouse
KeyHistory(500, False, True)
KeyHistory(500, False)
```


7pxvr:
ThioJoe spotted!

PR #355, opened 2026-02-19
Update Gui.ListView.cpp

iPhilip:
Fix to a counter-intuitive issue where a user has to create a large-icon image list and assign it explicitly as small-icon image list in order for the ListView to show large icons. See [this post](https://www.autohotkey.com/boards/viewtopic.php?f=82&t=140179) for more details on the topic.

### DIFF
--- a/source/lib/Gui.ListView.cpp
+++ b/source/lib/Gui.ListView.cpp
@@ -751,11 +751,10 @@ FResult GuiControlType::LV_SetImageList(UINT_PTR aImageListID, optl<int> aIconTy
 	int list_type;
 	if (aIconType.has_value())
 		list_type = aIconType.value();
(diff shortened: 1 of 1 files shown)


Lexikos:
This change would be contrary to documented behaviour:
> If omitted, the type of icons in the ImageList is detected automatically as large or small.

It does not say it will default to whichever ImageList the current view would use.

As I understand it, the IconType parameter is purely for selecting which views the ImageList applies to. Changing this *will* break scripts which use the following documented feature without specifying *IconType*:
> A ListView may have up to two ImageLists: small-icon and/or large-icon. This is useful when the script allows the user to switch to and from the large-icon view.

If the ListView is intended to be used this way, it is not intuitive for IconType to default to the current view.

If you think it is confusing, consider improving the documentation instead.

